Issue #25, opened 2016-10-06
Exported video file dimensions don’t match Kap settings


stursby:
It looks like the actual video file that Kap saves has slightly smaller dimensions from the settings.

For example, I set mine to 1280x720, and here's the `Get Info` panel for that file.

![screen shot 2016-10-06 at 1 23 13 pm](https://cloud.githubusercontent.com/assets/527849/19165151/5b0abdc8-8bc8-11e6-89ea-f43ffa3780a4.png)

The dimensions are *1276x716* (4px smaller in each direction).

matheuss:
Thanks fo the report, @stursby! I had that I mind but completely forgot about it. We'll make sure to include it in the next patch! 😄

CodeTheory:
@stursby On a side note, could I ask how the size of the window got that small? If it's a bug, I need to fix it.

stursby:
@CodeTheory which window are you referring to?

CodeTheory:
@stursby The Kap window in your screenshot looks off. 

Here's yours:
![image](https://cloud.githubusercontent.com/assets/1695613/19165848/b3114874-8c05-11e6-93f7-f62956b9bb4a.png)

Versus mine: 
![image](https://cloud.githubusercontent.com/assets/1695613/19165874/d113a272-8c05-11e6-8b16-95fb8b9d92b2.png)

stursby:
It looks like the window is changing widths after toggling the Controls panel. Here's three states:

1. Open Kap (320px)
2. Toggle Controls (~290px)
3. Detach window & Toggle Controls (~260px)

![kap-window-size](https://cloud.githubusercontent.com/assets/527849/19166219/58006fde-8bcc-11e6-9835-6b14802444fb.gif)

Hope this helps!

matheuss:
Whoa @stursby, that's really strange. Could you open an issue for it so we can discuss? 😄

stursby:
Done. https://github.com/wulkano/kap/issues/26

Issue #24, opened 2016-10-06
Gif support 


sirgalleto:
Can you export gif like recordit app?

doot0:
Hey, closing this as it's a dupe of https://github.com/wulkano/kap/issues/14.

It's in the works and should be landing soon! Thanks for your patience. 👌

sirgalleto:
@doot0 thanks and congrats, this app looks very nice

doot0:
@sirgalleto Thanks boss! 🖖

matheuss:
Ops, closed by accident 😩

PR #36, opened 2016-10-07
Switch cropperWindowBuffer to const

stursby:
Taking a stab at https://github.com/wulkano/kap/issues/25. I noticed in the renderer `index.js` the cropperBounds width and height are being subtracted by 4px. Seen [here]:(https://github.com/wulkano/kap/blob/master/app/src/renderer/js/index.js#L100-L101)

```javascript
cropperBounds.width -= 4;
cropperBounds.height -= 4;
```

So I added a `cropperWindowBuffer` which sets the cropperWindow preview 4px wider and taller, but correctly exports the same dimensions now.

kap-dimensions-fix

### DIFF
--- a/app/src/main/main.js
+++ b/app/src/main/main.js
@@ -26,6 +26,7 @@ require('./reporter');
 
 let appState = 'initial';
 let cropperWindow;
+const cropperWindowBuffer = 4;
 let mainWindowIsDetached = false;
 let mainWindow;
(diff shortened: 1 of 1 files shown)


matheuss:
Amazing again, @stursby! You just fixed two issues with one PR 😌 – #24 and the fact that the recorded area was not the same as the one shown in the marquee ✨ Thanks!

PR #37, opened 2016-10-08
[INCOMPLETE] Port to Linux

PaulBGD:
This PR is mostly waiting on the aperture PR to support Linux. There are a lot of inconsistencies with this code that need to be fixed before it gets merged however (see the review.)

Image
Image

Once the aperture PR is finished and pulled, the dependency in this project should be updated.


PaulBGD:
Huh. So it turns out Linux support might not even be possible, because you can't make a transparent window.. https://github.com/electron/electron/issues/2170

CodeTheory:
Really looks like some great work @PaulBGD. It'd be a shame if it really isn't possible! 😢

matheuss:
Amazing work @PaulBGD. I'll leave it as `on hold` until we manage to merge wulkano/aperture.js#3 😄

PaulBGD:
Tried that, on my computer it made random colors semi-transparent and was seemingly random.

GottZ:
--enable-transparent-visuals --disable-gpu does work on linux if you run a window manager that supports compositing.

PaulBGD:
@GottZ I'll check it out on a few window managers, but for some reason it didn't seem to work in Cinnamon (which uses compositing since 2012.)

PaulBGD:
Current status
Image

app/src/main/index.js
```diff
@@ -95,53 +115,90 @@ ipcMain.on('open-cropper-window', (event, size) => {
...
-        setCropperWindowOnBlur();
+    if (platform === 'linux') {
+      selectorWindow = createSelector(x, y, width, height);
+      selectorWindow.on('end', () => {
+        selectorWindow = undefined;
+        mainWindow.webContents.send('cropper-window-closed');
+      });
+      selectorWindow.on('move', event => {
+        console.log(event);
+        // because GTK/GDK makes it hard to get move/resize events differently, we do this
```
fearphage: Did you see this? http://stackoverflow.com/questions/1060039/gtk-detecting-window-resize-from-the-user

skllcrn:
Woah, thank you for reviewing this @fearphage!

scripts/unzip-ffmpeg.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-cd $1
+if [[ $OSTYPE == darwin* ]]; then
```
PaulBGD: Good point.

app/src/renderer/js/index.js
```diff
@@ -43,13 +46,19 @@ document.addEventListener('DOMContentLoaded', () => {
...
+  const trayTriangle = document.querySelector('.tray-arrow');
   const titleBar = document.querySelector('.title-bar');
   const trafficLights = document.querySelector('.title-bar__controls');
-  const trayTriangle = document.querySelector('.tray-arrow');
   const triangle = document.querySelector('.triangle');
   const updateNotification = document.querySelector('.update-notification');
   const windowTitle = document.querySelector('.window__title');
 
+  if (platform === 'darwin') {
+    content.classList.add('darwin');
```
PaulBGD: Great idea!

app/src/renderer/js/index.js
```diff
@@ -43,13 +46,19 @@ document.addEventListener('DOMContentLoaded', () => {
...
+    titleBar.style.display = 'block';
+    trayTriangle.style.display = 'block';
```
PaulBGD: Another good idea.

app/src/main/index.js
```diff
@@ -95,53 +115,90 @@ ipcMain.on('open-cropper-window', (event, size) => {
...
-      cropperWindow.webContents.on('devtools-opened', () => {
-        setCropperWindowOnBlur();
+    if (platform === 'linux') {
+      selectorWindow = createSelector(x, y, width, height);
+      selectorWindow.on('end', () => {
+        selectorWindow = undefined;
+        mainWindow.webContents.send('cropper-window-closed');
+      });
+      selectorWindow.on('move', event => {
+        console.log(event);
```
PaulBGD: Thanks.

app/src/main/index.js
```diff
@@ -95,53 +115,90 @@ ipcMain.on('open-cropper-window', (event, size) => {
...
+          x: event.x,
+          y: event.y
+        };
+        settings.setSync('cropperWindow.position', position);
+
+        selectorBounds = event;
+      });
+      selectorWindow.on('blur', () => {
+        if (!recording) {
+          selectorWindow.kill();
```
PaulBGD: Yeah originally I was planning on making an abstraction that would cover both of them, this was just the quickest way to check if it worked with Kap.

app/package.json
```diff
@@ -21,6 +21,7 @@
     "file-size": "^1.0.0",
     "first-run": "^1.2.0",
     "insight": "^0.8.3",
+    "linux-area-selector": "^1.2.0",
```
PaulBGD: My bad, I didn't add a repository field in the package.json.

scripts/download-ffmpeg.js
```diff
@@ -1,4 +1,11 @@
 #!/usr/bin/env node
+const os = require('os');
+
+if (os.platform() !== 'darwin') {
```
PaulBGD: I think this might be an accidental 2nd comment, but what do you mean about the message? I think it's pretty accurate as is, since the install script is being skipped since ffmpeg is already installed on Linux.

app/src/main/index.js
```diff
@@ -338,6 +403,10 @@ ipcMain.on('will-stop-recording', () => {
   if (cropperWindow) {
     cropperWindow.close();
   }
+  if (selectorWindow) {
+    selectorWindow.kill();
+    selectorWindow = undefined;
```
PaulBGD: Not sure if this is needed, since I unset the variable above in the quit event.

app/src/main/index.js
```diff
@@ -95,53 +115,90 @@ ipcMain.on('open-cropper-window', (event, size) => {
...
-  if (cropperWindow && !recording) {
+  if ((cropperWindow || selectorWindow) && !recording) {
     mainWindow.setAlwaysOnTop(false); // TODO send a PR to `menubar`
     menubar.setOption('alwaysOnTop', false);
-    cropperWindow.close(); // TODO: cropperWindow.hide()
+    if (cropperWindow) {
+      cropperWindow.close(); // TODO: cropperWindow.hide()
+    } else {
+      selectorWindow.kill();
+      selectorWindow = null;
```
PaulBGD: Also not sure this is needed. I also accidentally inconsistently set it to null, where ahead I set it to undefined.

PaulBGD:
I'm uninterested in maintaining this anymore. All code is licensed MIT.

PR #49, opened 2016-10-12
README: brew cask update to brew update

vitorgalvao:
`brew cask update` is simply a convenience alias to `brew update`. We don’t really promote `brew cask update` as that leads users to erroneously `brew update && brew cask update`, basically running the same command twice.


CodeTheory:
Thanks @vitorgalvao!

On a side note: Could you possibly submit the same thing to Hyper (https://github.com/zeit/hyper)? It'd be extremely helpful! 😄

vitorgalvao:
https://github.com/zeit/hyper/pull/866.

Issue #61, opened 2016-10-22
Launch on Startup


bnb:
Would be 💯 to see Kap be able to launch whenever I restart/power on my machine. There's a Node library that allows this pretty easily, which works with Electron: [node-auto-launch](https://github.com/Teamwork/node-auto-launch)

matheuss:
@bnb 👌 🚢 🔜 
<img width="287" alt="screen shot 2016-10-22 at 9 00 08 pm" src="https://cloud.githubusercontent.com/assets/4721750/19623085/8e50213a-989a-11e6-9c34-f692aedbcd07.png">

matheuss:
@bnb @65 🎉

PR #65, opened 2016-11-01
Implemented launch on startup

tjisousa:
Based on https://github.com/wulkano/kap/issues/61 .

I did test on macOS Sierra and it did work fine.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "aperture.js": "^0.1.0",
     "aspectratio": "^2.2.2",
+    "auto-launch": "^5.0.1",
     "electron-is-dev": "^0.1.2",
     "electron-settings": "^2.2.1",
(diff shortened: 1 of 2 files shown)


zeke:
I've had great luck with `auto-launch`, so I'm 👍  on shipping with that, but I figured I'd mention there is now an Electron API (for Windows and Mac) called `app.setLoginItemSettings(settings)` that enables this behavior:

http://electron.atom.io/docs/api/app/#appsetloginitemsettingssettings-macos-windows

tjisousa:
@zeke Nice news, thank you! 😃 
Since the app doesn't support just yet Linux, maybe I'll redo this and use the Electron API, what do you guys think? @matheuss @CodeTheory @skllcrn

matheuss:
@tjisousa I think we can stick with `auto-launch` for now 😄 Feel free to open another PR when you have some time 🙌

PR #68, opened 2016-11-03
Electron API to launch on startup

tjisousa:
Based on this https://github.com/wulkano/kap/pull/65#issuecomment-257733754

🐼


matheuss:
Thank you @tjisousa ☺️

PR #75, opened 2016-11-10
Save window position and size until Quit

stursby:
A couple of things with this one:
- Creating a `electron-settings` "Defaults" object for easy reset on Quit
- Listening for cropperWindow `move` and 'resize` and saving x,y & position
- On Quit, reset to "Defaults"


stursby:
Gonna close this, as #76 does the same thing, and more.

PR #76, opened 2016-11-10
Feature/persist config settings

stursby:
This one builds on #75 and refactors `settings` naming convention to camelCase.


doot0:
Having a look 👋

doot0:
This LGTM but prompted discussion of an `xo` linter rules change internally (thanks @stursby 😉). 

We'll merge this after we've come to a decision r.e. `xo`.

PR #77, opened 2016-11-11
Fix/xo expect longform method syntax

stursby:


matheuss:
@stursby thank you for the PR 😅 But as we discussed on Slack, we don't like the idea of disabling `XO` rules 😅

Issue #81, opened 2016-11-15
Where to find Eightpoint? 


StefMa:
Hey guys,

I've searched a long time on the big internet to find a solution for a electron renderer which is transparent and can be moved and resized like your cropper window. 
Great! 🎉 
I ❤️ it.

However. I need such a renderer in one of my applications as well. So I have looked around in your repository. My status:
<img width="383" alt="bildschirmfoto 2016-11-15 um 07 48 49" src="https://cloud.githubusercontent.com/assets/10229883/20295868/fbf71016-ab07-11e6-9f05-ef873ae715ce.png">
It works 🎉 

But I don't get it how do you implement the [eighpoint](https://github.com/wulkano/eightpoint). 
[Your css](https://github.com/wulkano/kap/blob/fe5d5eca08da061d63a87d0fe8f31a8f36a85e54/app/src/renderer/css/cropper.css) describes only
```css
@import 'eightpoint/dist/eight.min';
```
But if I run it with that (of course, I've added `"eightpoint": "0.0.1"` as dependency) the dev console say:
>  Failed to load resource: net::ERR_FILE_NOT_FOUND

If I change the import to
```css
@import 'node_modules/eightpoint/dist/eight.min.css';
```
it works.

But I guess this isn't the right way to use eightpoint, right?
So what I'm doing wrong? 🤔

CodeTheory:
Hi @StefMa, sorry for the late reply!

Eightpoint is a CSS helper library made bit wulkano (https://github.com/wulkano/eightpoint), so it isn't directly responsible for the draggable, transparent window itself. Eightpoint includes a css reset built with consideration for Electron apps. So it adds a system font stack, removes default margins on body's etc. 
In short, I'm not sure why you'd need Eightpoint for the window. 🤔 

But to answer your question:
We use a PostCSS module, [postcss-import](https://github.com/postcss/postcss-import), that has the ability to find modules in `node_modules` without directly referencing `node_modules`. So:
```css
@import 'eightpoint/dist/eight.min
```
would equate to
```css
@import '/node_modules/eightpoint/dist/eight.min
```

By also not including the filetype `.css`, the module that we use inlines the CSS instead of using default CSS import rules.

So for you to use eightpoint in the same way as us, you'll need a module like postcss-import. 

Does this answer sufficiently?

StefMa:
> Does this answer sufficiently?

Yep, thank you. That was everything I need to know 👍 

[...]

PR #84, opened 2016-11-17
Fix window without eightpoint

StefMa:
Currently eighpoint helps the window to remove the default margin.

But, for other interests looking into this project, makes hard to find out why the window looks wired without eightpoint.
So I've added the zero margin to the cropper css.

Further reading: https://github.com/wulkano/kap/issues/81

### DIFF
--- a/app/src/renderer/css/cropper.css
+++ b/app/src/renderer/css/cropper.css
@@ -3,6 +3,7 @@
 html, body {
   width: 100%;
   height: 100%;
+  margin: 0px;
   overflow: hidden;
 }
 


StefMa:
retest this please

CodeTheory:
Hi @StefMa. I'm sorry for the delay on this, I've been extremely busy. 

I don't think there's a point in adding this extra line of code since we're using eightpoint as a dependency for the entire app and we should view that dependency (since it's a reset) as immutable. So we wouldn't remove eightpoint, even if we're just using one line of code from it in this particular instance. It's better for consistency I think.

Thanks for your interest in contributing!

PR #96, opened 2016-12-08
Feature/add hotkey support

stursby:
Here's my go at adding `globalShortcut` support. Here's a recap of what's going on:

- <kbd>Cmd</kbd> + <kbd>Shift</kbd> + <kbd>5</kbd> will prepare Kap for recording (_not_ start)
- Once the `cropperWindow` is activate, press <kbd>Enter</kbd> to _begin_ recording
- Once _recording_, register <kbd>Esc</kbd>
- When `stop-recoding` is fired, remove Esc globalShortcut

Other:
- when running locally with `npm run dev` && `npm start`, dev tools will have focus, so you have to click on `cropperWindow` for <kbd>Enter</kbd> to `start-recording`


skllcrn:
I still think <kbd>Space</kbd> should be used to start the recording, as that's how I'm used to controlling media. I get that this isn't playback, but I still feel it's the natural option, maybe in addition to <kbd>Enter</kbd>.  While I like being able to stop the recording with <kbd>Esc</kbd>, I'm worried you'd want to use that key for other uses while recording.

stursby:
@skllcrn ok, maybe both <kbd>Enter</kbd> and <kdb>Space</kbd> should work for _start_?

I was trying to think of a situation where you'd need <kdb>Esc</kbd> during a recording? Could aways use <kdb>Ctrl</kbd> + <kdb>Esc</kbd>... just felt cumbersome. But if we eventually let users set custom key commands in Preferences, might be a moot point.

skllcrn:
Let's try both and see how that works until we revisit this when adding shortcut settings to Preferences!

stursby:
@skllcrn updated w/ #cf94ddd 😎

sindresorhus:
For reference, QuickTime supports <kbd>Esc</kbd> to stop recording. ScreenFlow.app and ScreenFlick.app uses Cmd+Shift+2 (User can change this) to both start and stop recording.

sindresorhus:
I think using <kbd>Esc</kbd> would be annoying as it could easily be used during capture. I would much prefer having the same shortcut to both start and stop the recording.

app/src/renderer/js/cropper.js
```diff
@@ -37,6 +39,10 @@ document.addEventListener('DOMContentLoaded', () => {
     if (event.keyCode === ESC_KEY_CODE) {
       autoDestroy();
     }
+
+    if (event.keyCode === ENTER_KEY_CODE || SPACE_KEY_CODE) {
```
sindresorhus: Use https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key instead, and you don't need to have constants for the different keys ;)

Khaledgarbaya:
I think using <kbd>Esc</kbd> to stop recording will super annoying for vim users like me.

app/src/renderer/js/cropper.js
stursby: Good to know! Refactored with `KeyboardEvent.key`

stursby:
@skllcrn @sindresorhus @Khaledgarbaya Updated. The new process is as follows:

1.  <kbd>Cmd</kbd> + <kbd>Shift</kbd> + <kbd>5</kbd> : Bring cropperWindow to front / prepare for recording
2. <kbd>Enter</kbd> or <kbd>Spacebar</kbd> : Start recording
3. <kbd>Cmd</kbd> + <kbd>Shift</kbd> + <kbd>5</kbd> : Stop recording

Issue #97, opened 2016-12-08
Dropping files into Kap results in unexpected behavior


scferg:
I'm new to using Kap, and I thought that maybe dragging an already exported GIF back into Kap would allow me to edit it (size, FPS, etc.). Looks like that's not possible, but it also does some strange things...
![item-0_and_item-0_and_item-0](https://cloud.githubusercontent.com/assets/1112334/21029343/4631d794-bd67-11e6-92e4-a951a4bc7d7f.png)

It just pops the image into the Kap window and then there's no way to quit/reset Kap without going into Activity Monitor and forcing the process to quit. Right-clicking on the menubar icon no longer displays a menu.

Just to see what else happens... 😈
- Dragging a zip file into Kap prompts you to save the zip file
- Dragging a folder into Kap results in the menubar icon not displaying _anything_ when clicked

Not many people should come across this issue, but perhaps it should be prevented from breaking the app with no way to reset it.

Great app though 😀

skllcrn:
Oops, ha. 😅

> I'm new to using Kap, and I though that maybe dragging an already exported GIF back into Kap would allow me to edit it (size, FPS, etc.). 

That's an excellent idea though!

stursby:
@skllcrn @scferg https://github.com/wulkano/kap/pull/100 fixes this. We might later revisit with the "drop to edit" idea! 👍

PR #100, opened 2016-12-09
Ignore file drops in Kap

stursby:
This fixes https://github.com/wulkano/kap/issues/97. It prevents dropping files in the *Main*, *Editor*, and *Preferences* windows. The Cropper window did not respond to file drops, so I let that alone.

kapture 2016-12-09 at 16 05 54


matheuss:
Thank you @stursby ❤️

PR #109, opened 2016-12-11
Ensure the `Control` toggle gets a hand cursor

sindresorhus:
It didn't work previously as something else was overriding its style.


matheuss:
For the record: it's the `txt` class that was overriding it 😅

Issue #111, opened 2016-12-12
Record Region dimension is not updated correctly


Khaledgarbaya:
Hi Guys, First of all I am super happy and excited for the **v1.0** release.

Bug report : 
- version: **1.0**
- Step to reproduce : 
  open kap from the menu bar, before clicking the red icon to define the region, expand _Controls_ pane and enter different dimension. than click define region button (the red icon). the record region will show up with the old dimensions (512 x 512), instead of the defined one in the text inputs.

- Current behaviour : The record region is shown with the wrong dimensions
- Expected behaviour :  The record region dimensions should be correct
- Wordaround : Re-enter the dimensions again .

PR #114, opened 2016-12-12
Fix cropper window

Khaledgarbaya:
Fixes https://github.com/wulkano/kap/issues/111

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -85,7 +85,9 @@ ipcMain.on('open-cropper-window', (event, size) => {
   if (cropperWindow) {
     cropperWindow.focus();
   } else {
-    const {width = size.width, height = size.height} = settings.get('cropperWindow.size');
+    let {width = 512, height = 512} = settings.get('cropperWindow.size');
+    width = size.width || width;
+    height = size.height || height;
     const {x, y} = settings.get('cropperWindow.position');
     cropperWindow = new BrowserWindow({
       width: width + cropperWindowBuffer,


Khaledgarbaya:
@matheuss  @sindresorhus @CodeTheory any progress on this ?

CodeTheory:
Sorry for the delay @Khaledgarbaya. It's been a busy delay for all of us. I'll get one of the team to review it this week. ❤️

doot0:
Hey @Khaledgarbaya. Taking a quick look at your PR (thank you by the way!), LGTM.

I'm wondering if perhaps the following is a bit more readable however;

```
const dimension = 512;

let {width, height} = settings.get('cropperWindow.size');
width = size.width || dimension;
height = size.height || dimension;
```

Would like your thoughts. :)

Khaledgarbaya:
hi @doot0 ,
In your snippet, getting the width and height from the settings will be broken.

```js
let {width=512, height=512} = settings.get('cropperWindow.size');
```
☝🏼 what that snippet is doing is checking for example if the settings had already some value for the width otherwise it will assign 512 to it, same for height.

and for 

```
width = size.width || width;
height = size.height || height;
```
will check  if there is a size.width, and size.height otherwise it will stick with the default values saved in the settings.

Let's take this scenario 
You open the app first time and you set the copper dimension. Kap should save that in the settings.
when you come back 

```js
let {width, height} = settings.get('cropperWindow.size');
```
☝️ should have the values from your settings. But

```js
width = size.width || dimension;
height = size.height || dimension;
```
☝️ will set the width and height to dimension if the size is not provided, which will break the feature of saving the last saved setting

doot0:
LGTM. 👍

PR #148, opened 2017-02-18
Updating ffmpeg download URL.

Meandmybadself:
_npm install_ was failing when attempting to download ffmpeg at an erroring URL.

This updates the URL to its valid equivalent.


matheuss:
Thank you 🤗

PR #175, opened 2017-05-14
Improve the look of the DMG

sindresorhus:
Makes the icons larger and the distance to drag shorter, to be more user-friendly.

## Before

screen shot 2017-05-14 at 15 07 55

## After

screen shot 2017-05-14 at 17 28 25

*The arrow is slightly blurry because of https://github.com/electron-userland/electron-builder/issues/1506*


skllcrn:
So the blur will be fixed upstream?

sindresorhus:
@skllcrn Yes, it will be fixed soon, before we'll ship a new release.

PR #180, opened 2017-05-15
Move export format buttons to the editor

sindresorhus:
This is the first step towards plugins. Would like to merge this as soon as possible.

- Move export format buttons to the editor.
- Removes discard/save buttons.
- Makes it possible to export to multiple formats.
- The user now gets the save dialog right away instead of after the conversion is complete. That is a better UX, as the user can then go do something else while it converts.
- Lots of refactoring.

screen shot 2017-05-15 at 17 39 50

A few minor TODO left:

- [x] Clean up the original MP4 file on editor window close
- [x] When the user clicks the red traffic light in the editor, show the prompt for confirmation to discard.
- [x] Disable the export buttons during export
- [x] With these changes it would make more sense if the export option buttons weren't grouped.

---

**Please help me test this.**


CodeTheory:
For anyone wanted to test, download the artifact here: https://kap-artifacts.now.sh/move-export-format-buttons

CodeTheory:
Actually @sindresorhus I've just noticed that you can't close the editor window. It just hides. As soon as you click the kap icon in the menu-bar, it re-opens the editor.

sindresorhus:
All done.

skllcrn:
5x approved, for good luck.

PR #183, opened 2017-05-18
Fix settings corruption

### LINKED ISSUE #159: App completely unuseable
**Platform:**
macOS Sierra 10.12.3
image

**Kap Version:**
1.0.1
### Steps to reproduce
Opening App. Nothing works. First attempted with a second display, but does not work with single display either.
### Current behaviour
It seems that no buttons are clickable. Nothing happens when I click buttons on the app. I cannot even quit the app without using the activity monitor.

screen recording 2017-03-23 at 12 50 pm
### Expected behaviour
I expect to be able to use the app.
### Workaround
 I have found no workaround. I even uninstalled the app and installed the latest update a month apart. Still have the same issue.

sindresorhus:
This sets the default settings no matter what. Previously it would not be set if the Aperture called failed.

Fixes #159

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -12,7 +12,7 @@
   },
   "homepage": "https://getkap.co",
   "dependencies": {
-    "aperture": "2.1.0",
+    "aperture": "2.1.1",
     "aspectratio": "^2.2.2",
     "eightpoint": "0.0.1",
     "electron-is-dev": "^0.1.2",
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -13,7 +13,8 @@ const DEFAULTS = {
   showCursor: true,
   highlightClicks: false,
   fps: 30,
-  recordAudio: false
+  recordAudio: false,
+  audioInputDeviceId: 'none'
 };
 
 const volatiles = {
@@ -37,17 +38,19 @@ function sync() {
 }
 
 function init() {
+  settings.defaults(DEFAULTS);
+  settings.applyDefaultsSync();
+  sync();
+
   // We need to fetch a input device because if the user opens the app for the first time
   // and toggle the mic in the main window to record audio, we will not record any audio
   // if we do not have a input id stored.
   // TODO: if no input device is available (could happen in an iMac, for example), we need
   // to tell the user
-  // TODO: rewrite this comment (it's 4AM and i'm fucking tired)
   aperture.getAudioSources().then(devices => {
-    DEFAULTS.audioInputDeviceId = (devices && devices[0] && devices[0].id) || 'none';
-    settings.defaults(DEFAULTS);
-    settings.applyDefaultsSync();
-    sync();
+    if (devices.length > 0) {
+      settings.setSync('audioInputDeviceId', devices[0].id);
+    }
   });
 }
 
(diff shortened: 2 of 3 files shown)


sindresorhus:
@torkiljohnsen @iamtirado @andreleon Can you try it out?

You can download it here: https://kap-artifacts.now.sh/settings-corruption
Make sure you delete `~/Library/Application Support/Kap` first.

The call to get the microphone might still fail, so let me know if it does, so I can find a way to handle it.

torkiljohnsen:
Not sure what you mean by the call to get the microphone, but I can't record with sound if that's what you're asking. Have no sound input to select either.

image

Also, when opening Kap I get this error:
skjermbilde 2017-05-18 kl 12 57 55

Other than that, it seems to be capturing just fine now 👍

iamtirado:
This app is wonderful. Glad I can record with it now.

I probably won't use audio much, but I have the same audio issue as @torkiljohnsen where I can't select audio input and I also get the same popup at beginning.

sindresorhus:
@torkiljohnsen Can you go to `Kap.app/Contents/Resources/app/node_modules/aperture/index.js` and change line 19 to the following:

```js
return execa.stdout(path.join(__dirname, 'swift/main'), ['list-audio-devices']).then(x => Promise.reject(new Error(x)));
```

And then run Kap and paste the error message here?

torkiljohnsen:
image

iamtirado:
my error: is it different? 
image

sindresorhus:
Thanks for the help debugging this.

@torkiljohnsen Looks like you have an outdated version of the Bartender app installed or some remainders of it: https://webcache.googleusercontent.com/search?q=cache:lW3xY3OCuLUJ:https://origin-discussions-us.apple.com/thread/7286340+&cd=1&hl=en&ct=clnk&gl=th&lr=lang_en%7Clang_no

@iamtirado I think it's because we're by mistake building the binary in debug mode.

Can you both download and unzip [this](https://www.dropbox.com/s/rgled65edanlpgv/aperture%20main.zip?dl=0) and replace `Kap.app/Contents/Resources/app/node_modules/aperture/swift/main` and then start Kap?

iamtirado:
Something changed in the debug log, but still can't select audio.
image

torkiljohnsen:
@sindresorhus Indeed I had some old Bartender2 stuff around :) 

Replaced the `main` file and got this error when starting Kap:
image

Still no input devices.

image

sindresorhus:
Alright. Can you try out https://kap-artifacts.now.sh/settings-corruption It should work now. Let me know.

sindresorhus:
@iamtirado @torkiljohnsen Would you be able to confirm ⬆️ it's fixed?

iamtirado:
I get the following error upon opening.



I can record my screen. Audio input is working, and all ui seems to be functional. When I go to save to gif, mp4, or webM the window for choosing save location closes, and the file never saves. Preview window stays open. so I can keep watching what I just recorded, but it doesn’t save. Nothing happens when I hit save.

—
Eric Tirado
Brand Design | Web Development
erictirado1@gmail.com
678-577-8064

CodeTheory:
The error you're getting is normal for a build artifact. 

Are you sure it's not saving? The new behaviour of Kap is to let the editor remain open after saving so that you can perform multiple export actions. Did you check that the file is not in the location that you requested it to be saved in?

iamtirado:
Yes i saved to desktop and other places and food not see the file.

sindresorhus:
Merging as it fixes what it was intended too. We'll look into the file not saving thing.

Issue #164, opened 2017-04-13
Export to APNG


FezVrasta:
The browser support for APNG seems pretty good nowadays, do you have plans to support it?

https://en.wikipedia.org/wiki/APNG#Support

CodeTheory:
Yes, we'll definitely look into supporting it. All for progress of better formats!

fokusferit:
I did a initial search through the code and ffmpeg and it seems straight forward -> adding a new convert function to convert.js + some addition to the editor + main.js . 

And since v3.3 ffmpeg supports apng. If no one is currently working on it, I'd like to do it.

CodeTheory:
Added in #185 😄

FezVrasta:
Cool!

PR #185, opened 2017-05-24
Add APNG support

fokusferit:
I've added a first apng support as commented in #164 . 

I'm not experienced enough to be 100% sure if my converter function to apng is really the best one. Sadly the FFMpeg website has no clear information regarding APNG Muxer. According to https://github.com/FFmpeg/FFmpeg/blob/master/Changelog#L253 ffmpeg has basic APNG support since v2.7.

- I added a convertToApng function
- tested in on dev mode + on the dist dmg file after running `npm run pack` and it is working
- Tested .apng functionality on FF v53 and Chrome Canary v60.X
-  it is working 🎉 .

The only "concern" is that .apng's with the current convert function tend to be quite 'fat', e.g. the .gif below is X mb whereas the same .apng is 21,4 MB 😨

dogfood_convertapng

Maybe someone here has more experience and can help me out with better flags / filters.

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -443,6 +443,8 @@ ipcMain.on('open-editor-window', (event, opts) => {
     frame: false
   });
 
+  app.kap.editorWindow = editorWindow;
+
   editorWindow.loadURL(`file://${__dirname}/../renderer/views/editor.html`);
 
   editorWindow.webContents.on('did-finish-load', () => editorWindow.webContents.send('video-src', opts.filePath));
@@ -467,7 +469,6 @@ ipcMain.on('open-editor-window', (event, opts) => {
     }
   });
 
-  app.kap.editorWindow = editorWindow;
   menubar.setOption('hidden', true);
   mainWindow.hide();
   tray.setHighlightMode('never');
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -436,10 +436,10 @@ ipcMain.on('open-editor-window', (event, opts) => {
   }
 
   editorWindow = new BrowserWindow({
-    width: 768,
-    minWidth: 768,
-    height: 428,
-    minHeight: 428,
+    width: 800,
+    minWidth: 800,
+    height: 430,
+    minHeight: 430,
     frame: false
   });
 
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -512,8 +512,9 @@ document.addEventListener('DOMContentLoaded', () => {
       if (!outputPath) {
         return;
       }
-
-      app.kap.editorWindow.send('toggle-format-buttons', {enabled: false});
+      // Initially an exception is triggered that editorWindow is undefined
+      // Using getter here should make it 'safer'
+      app.kap.getEditorWindow().send('toggle-format-buttons', {enabled: false});
       app.kap.mainWindow.show();
 
       const input = Object.assign({}, data, {outputPath});
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -512,9 +512,8 @@ document.addEventListener('DOMContentLoaded', () => {
       if (!outputPath) {
         return;
       }
-      // Initially an exception is triggered that editorWindow is undefined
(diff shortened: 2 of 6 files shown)


skllcrn:
Perfect timing, thank you!

app/src/scripts/convert.js
```diff
@@ -82,6 +82,19 @@ function convertToWebm(opts) {
   });
 }
 
+// Should be similiar to the Gif generation
+function convertToApng(opts) {
+  return Promise.resolve().then(() => {
+    return convert(opts.outputPath, opts, [
+      '-i', opts.filePath,
```
sindresorhus: You need to set the framerate and width/height too, like done with the GIF.

https://www.ffmpeg.org/ffmpeg-formats.html#apng

Or maybe try using:

```
'-filter_complex', `fps=${opts.fps},scale=${opts.width}:${opts.height}:flags=lanczos[x];`
```

app/src/scripts/convert.js
```diff
@@ -82,6 +82,19 @@ function convertToWebm(opts) {
...
+      // Strange for apng instaed of -loop it uses -plays see: https://stackoverflow.com/questions/43795518/using-ffmpeg-to-create-looping-apng
```
sindresorhus: `apn` => `APNG`

`instaed` => `instead`

app/src/renderer/js/main.js
```diff
@@ -510,8 +512,9 @@ document.addEventListener('DOMContentLoaded', () => {
       if (!outputPath) {
         return;
       }
-
-      app.kap.editorWindow.send('toggle-format-buttons', {enabled: false});
+      // Initially an exception is triggered that editorWindow is undefined
+      // Using getter here should make it 'safer'
```
sindresorhus: I think that is because of a race issue. Can you try moving this https://github.com/wulkano/kap/blob/46c63f6d9a95e30fbab301ccbc99c6fb821ddbb0/app/src/main/index.js#L470 up to line 445 in the same file?

app/src/main/index.js
```diff
@@ -436,10 +436,10 @@ ipcMain.on('open-editor-window', (event, opts) => {
...
 
   editorWindow = new BrowserWindow({
-    width: 768,
-    minWidth: 768,
-    height: 428,
-    minHeight: 428,
+    width: 800,
+    minWidth: 800,
+    height: 430,
+    minHeight: 430,
```
sindresorhus: Can you explain these changes?

fokusferit: The text and size block are stacked as I've added a new button 'APNG'
Old values: 
screen shot 2017-05-24 at 13 43 01

New values (800x430) 
screen shot 2017-05-24 at 13 41 52

app/src/scripts/convert.js
fokusferit: The filter_complex throws errors in the console, not sure why tbh. The convert function just executes `converter.catch(reject);`without any real error message

sindresorhus: Try it directly on the command-line.

fokusferit: This command (testing)
`ffmpeg -i test.mp4 -filter_complex fps=14,scale=1024:1024:flags=lanczos[x] tests.apng`
returns the following error
`Filter scale has an unconnected output`. According to ffmpeg Filters [docs](https://ffmpeg.org/ffmpeg-all.html#Filtering) filter_complex should be used for multiple outputs. Is it not enough to use `-vf` instead of -filter_complex ?
I've tried your settings just with `-vf`and it works in the command-line and in the electron app.
`ffmpeg -i test.mp4 -vf fps=14,scale=1024:1024:flags=lanczos[x] -plays 0 test_convert.apng`

sindresorhus: > I've tried your settings just with -vfand it works in the command-line and in the electron app.

👍

app/src/renderer/js/main.js
fokusferit: Looks like your tip solved it, thanks!

sindresorhus:
Looks great. Thank you @fokusferit 🙌🎉

Issue #73, opened 2016-11-09
Cancel processing


vadimdemedes:
Would be very convenient to be able to cancel processing, if I see it's going too slow and obviously will result in a very large gif (162mb, in my case). My macbook's cooling fan was going like crazy :D

doot0:
Hey @vdemedes, that sounds like a great idea. Seems silly not to add this. 👍

delcore92:
Hi all, I wouldn't mind giving this issue a shot.
My plan is to store the save / convert action as deferred promise. Then adding a cancel button to the "progress"  view which will resolve the promise and perform all necessary clean up. 

Here's a screen grab of my proposed ui change:
<img width="309" alt="screen shot 2017-05-26 at 12 31 10 pm" src="https://cloud.githubusercontent.com/assets/3030010/26478442/a40cc536-420f-11e7-9da4-db42cd6a7a07.png">

sindresorhus:
I think it would be better to use a cancelable promise. You could use https://github.com/sindresorhus/p-cancelable

sindresorhus:
And awesome that you want to work on this 🙌

delcore92:
Great suggestion, thank you! I'll try my best!

sindresorhus:
@delcore92 I merged my big PR (https://github.com/wulkano/kap/pull/186), so you feel free to work on this whenever you see fit :)

delcore92:
@sindresorhus great, i'll have a look now :)

delcore92:
I've just raised a PR that includes the small UI change. I would love to get some conversation around how we can approach this issue. 

To be honest, i'm a bit lost with the latest refactor. I'm struggling to find a good place to wrap the export process in a cancelable promise.

skllcrn:
Related to #263

PR #189, opened 2017-05-29
Cancel image processing

delcore92:
**This is a work in progress**

This pr attempts to address #73 :

> Would be very convenient to be able to cancel processing, if I see it's going too slow and obviously will result in a very large gif (162mb, in my case). My macbook's cooling fan was going like crazy :D

Proposed UI change: 

I would really appreciate some guidance regarding the cancelation logic.

### DIFF
--- a/app/src/renderer/css/main.css
+++ b/app/src/renderer/css/main.css
@@ -237,6 +237,7 @@ section.progress {
 
 .progress-bar {
   width: 206px;
+  margin-bottom: 16px;
 }
 
 .progress-bar-label {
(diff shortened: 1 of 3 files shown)


fokusferit:
I think the cancelation needs a way to cancel the convert function here https://github.com/wulkano/kap/blob/master/app/src/scripts/convert.js#L12 .
Whenever the user clicks on the cancel button you would need to cancel the conversion process. 

I don't know yet how to cancel [execa](https://github.com/sindresorhus/execa) there are some options https://github.com/sindresorhus/execa#api but one solution could be to save the execa process globally and cancel / overwrite it when cancel Button is clicked. 

Otherwise I think this might need a change in how the convert function is currently written as Promises can't be cancelled.

sindresorhus:
@fokusferit You can use `p-cancelable` and call `converter.kill` in the `onCancel` function provided by `p-cancelable`.

PR #191, opened 2017-05-29
Use the native traffic lights for the Preferences window

sindresorhus:
To make it look consistent with other macOS apps.

This also made it easy to disable minimizing the window. Preference windows should not be minimizable on macOS.

## Before

screen shot 2017-05-29 at 19 05 54

## After

screen shot 2017-05-29 at 19 14 03

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -193,8 +193,10 @@ function openPrefsWindow() {
   prefsWindow = new BrowserWindow({
     width: 480,
     height: 480,
-    frame: false,
     resizable: false,
+    minimizable: false,
+    maximizable: false,
+    titleBarStyle: 'hidden',
     show: false
   });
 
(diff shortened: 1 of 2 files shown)


CodeTheory:
We made them ourselves because they were easier to position. Need @skllcrn to review this one because of the positioning, I think.

sindresorhus:
@CodeTheory I agree that it's useful for the editor window, since you wanted to animate the traffic lights. But there's no benefit in using a custom implementation in the Preferences window.

skllcrn:
@CodeTheory @sindresorhus I believe the reason for not using the native traffic lights in the main window was due to them being hidden in the attached state and it not being possible to display them again when detached from the menubar. Basically you have window controls or you don't. Great that we can use the native ones in other windows, the custom ones are designed and positioned to be identical to the native ones.

PR #192, opened 2017-05-29
Fix share selection index bug

SamVerschueren:
When having 3 export selections for GIF like this

1. Save to disk
2. Share to GIPHY
3. Share to X

And 2 for MP4

1. Save to disk
2. Share to X

When selecting `Share to X` in the MP4 dropdown, it will still share it to giphy. This PR solves that.


sindresorhus:
Great. Thanks for catching that issue Sam.

PR #193, opened 2017-05-29
Percentage in `setProgress()` should be set as 0-1, not 0-100

sindresorhus:
This is a better way to do percentage in JS.


kevva:
:+1:, this is how I expect percentages to work.

Issue #53, opened 2016-10-17
Rewrite the integration with the menubar


matheuss:
The current code to deal with the tray icon/menubar is very confusing and complex: we're relying on (a fork of) the `menubar` module and we're doing a lot of things on our on. There's a lot of confusing and duplicated code.

skllcrn:
Upcoming as part of the 2.0.0 milestone! Hopefully this can close or progress #1, #3, #4, #7, #44 and #112.

skllcrn:
Suggest dealing with #162 in conjunction with this issue.

skllcrn:
#472

skllcrn:
Kap no longer attaches to the menubar (#494 and #223)

PR #196, opened 2017-05-29
Improve dragging of the main window

vadimdemedes:
This PR doesn't fix all issues, but it improves the overall dragging experience and removes often-occuring glitches when dragging main window close to the menubar icon. I confirmed if `resetMainWindowShadow()` isn't called during dragging, there are no more glitches left with this PR. Perhaps this is something we could just improve/remove later when it's fixed in Electron.

Before: https://www.dropbox.com/s/fp5ph28co96s8vi/before.mov?dl=0
After: https://www.dropbox.com/s/icpgz03p3wujur8/after.mov?dl=0

This PR stops relying on the native `move` event from Electron's `MainWindow`, because it's triggered only when the dragging is finished on the OS X. That's why Kap can't react to the dragging changes immediately. As a result, it's hard to predict in which place the window will get stick/unstick  to/from the menubar. Instead, I'm relying the combination of detecting dragging operation in the renderer process and using `screen.getCursorScreenPoint()`, which provides accurate coordinates of the mouse position. `start-move-window` event is used to record initial mouse position inside the main window, so that we can later detect correct coordinates of the main window position.

It also stops showing/hiding dock icon during the drag operation, because that may result to lags. Instead, it happens when the dragging is finished (mouse is released).

Let me know what you think!


sindresorhus:
@CodeTheory @skllcrn Can you try this one out?

matheuss:
It feels way more buggy than the current implementation 😔 https://file-hqerarpbfp.now.sh/kap-improve-dragging.mov

I remember trying something like this over the IPC back in the early days, and I wasn't very successful – the outcome was pretty much similar 😕 Not sure if there's a trivial way to improve it. 

It would be **neat** if `move` wasn't an *alias* for `moved` 😕

vadimdemedes:
Hmm, that's not the behavior that I was getting. I'll get back to this PR later and see what can be done.

skllcrn:
For me it's a little better in certain situations. There's a few issues related to the menubar implementation, if you're interested see #53

sindresorhus:
Closing this as we're going a different way (#409) with Kap 3. Sorry, Vadim.

PR #198, opened 2017-05-30
Implement default share service

albinekb:
in 99% of cases i just want the gif as fast as possible, this is a first stab at doing so:

image

This adds an option to skip the editor and use a default share plugin directly, thoughts?

Needs some more work like if you uninstall a plugin etc


albinekb:
Maybe this is even better: 
image

CodeTheory:
@albinekb We have a redesign coming up that would make this different.
@skllcrn thoughts?

skllcrn:
This wouldn't fit with our goal of making the main window extremely compact. It also a setting you'd use to control the app behaviour for some time, for changing often it's better to keep the default of going to the editor as it has all the options. Based on that I'd say that this belongs in preferences.

sindresorhus:
I would also put it in Preferences.

app/src/renderer/js/preferences.js
```diff
@@ -139,6 +183,8 @@ document.addEventListener('DOMContentLoaded', () => {
   });
 
   loadInstalledPlugins();
+  loadInstalledPlugins();
```
sindresorhus: This is duplicated.

app/src/renderer/js/main.js
```diff
@@ -185,6 +186,8 @@ document.addEventListener('DOMContentLoaded', () => {
       displayId: display.id
     };
 
+    _apertureOpts = apertureOpts;
```
albinekb: Ideally I don't want to have this mutated variable here, any ideas?

app/src/renderer/js/preferences.js
```diff
@@ -94,6 +94,50 @@ document.addEventListener('DOMContentLoaded', () => {
...
+              <%- option.title %> [<%- option.format %>]
+            </option>
+        <% }); %>
+      </select>
+    `;
+
+    const compiled = _.template(template);
+    const current = app.kap.settings.get('defaultShare');
+
+    const options = shareServices.reduce((all, service) => {
```
vadimdemedes: Could you rewrite this using `for..of`? `reduce()` usage isn't that readable and easy to follow, so I'd just populate a results array.

app/src/renderer/js/preferences.js
```diff
@@ -94,6 +94,50 @@ document.addEventListener('DOMContentLoaded', () => {
...
+    }, []);
+
+    const html = compiled({
+      options
+    });
+
+    $j('#default-share-service').html(html);
+    $j('#default-share-service select').on('change', event => {
+      const {value} = event.target;
+      if (value) {
```
vadimdemedes: This could be rewritten as:

```js
let shareService = {};

if (value) {
  const [pluginName, format] = value.split('|');
  shareService = {pluginName, format};
}

app.kap.settings.set('defaultShare', shareService);
```

app/src/renderer/views/preferences.pug
```diff
@@ -36,6 +36,12 @@ block body
             .preference__title Allow analytics
             .preference__description Help us improve Kap by sending anonymous usage stats
           +toggle('allow-analytics')
+      .preference.container
+        .preference-part
+          .preference-content
+            .preference__title After recording
+            .preference__description Select action
+          .select-element(id='default-share-service', style="width: 17.8rem")
```
vadimdemedes: Could you extract this into stylesheet?

app/src/renderer/js/preferences.js
albinekb: rewriting it with `for..of` will make the code mutate, I prefer to write immutable code, but sure

app/src/renderer/js/preferences.js
albinekb: Wouldn't that reset the shareService? 🤔

albinekb: I prefer to keep the code immutable (no let), but if you want that, sure 👍

app/src/main/index.js
```diff
@@ -437,6 +437,23 @@ ipcMain.on('move-cropper-window', (event, data) => {
...
+        filePath: opts.filePath,
+        width: opts.apertureOpts.cropArea.width,
+        height: opts.apertureOpts.cropArea.height,
+        fps: opts.apertureOpts.fps,
+        loop: true
+      });
+
+      return;
+    }
+  }
```
sindresorhus: Can you move this to a separate function and just use it like this:

```js
if (runDefaultService(opts)) {
  return;
}
```

app/src/renderer/js/main.js
sindresorhus: I agree it should be improved, but can't think of anything right now.

app/src/renderer/js/preferences.js
sindresorhus: @albinekb I usually agree, but reduce used like this is not that readable.

albinekb:
Do we still want this? I got time to fix the comments and rebase  🙌

albinekb:
Putting an issue up in a sec 👍

PR #205, opened 2017-06-09
Use the new electron.Notification API

sindresorhus:
https://github.com/electron/electron/blob/a6ea316a5de1964761a82d9376c23e99eff16d25/docs/api/notification.md


app/src/main/share-service-context.js
```diff
@@ -29,14 +29,10 @@ export default class ShareServiceContext {
   }
 
   notify(text) {
-    // TODO: Add Kap icon as `icon` option
-    // TODO: Waiting for: https://github.com/electron/electron/pull/9269
-    // (new electron.Notification({title, body})).show();
-
-    electron.app.kap.mainWindow.send('show-notification', {
+    (new electron.Notification({
```
albinekb: Time to add the `icon`? https://github.com/electron/electron/blob/master/docs/api/notification.md#new-notificationoptions-experimental

package.json
```diff
@@ -19,9 +19,9 @@
     "chalk": "1.1.3",
     "cssnano": "3.10.0",
     "del": "2.2.2",
-    "electron": "1.7.2",
-    "execa": "0.6.3",
+    "electron": "1.7.3",
```
albinekb: Do you need to upgrade [`electron-builder`](https://github.com/electron-userland/electron-builder/releases) at the same time? 🤔

sindresorhus: No, that's unrelated.

app/src/main/share-service-context.js
sindresorhus: I considered it, but the app icon is used by default, so wasn't needed.

albinekb: aha nice! 👌

albinekb:
Maybe use the new API here too? https://github.com/wulkano/kap/blob/master/app/src/main/share-service.js#L82

sindresorhus:
> Maybe use the new API here too? https://github.com/wulkano/kap/blob/master/app/src/main/share-service.js#L82

Done

PR #209, opened 2017-07-17
Don't `.show()` the main window if Kap was opened on system startup

### LINKED ISSUE #208: Don't open on system startup
The app should open, but not the main window 😊

matheuss:
Fixes #208 👌 

Also bumped the `ffmpeg` version since the current one is gone

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -297,8 +297,8 @@ menubar.on('after-create-window', () => {
       // because of that, we need to move the window to it's expected position, since the
       // user will never release the mouse in the *right* position (diff.[x, y] === 0)
       if (!wasOpenedAtLogin(app)) {
-        // Seems like electorn fires the `move` event right aftet the app boots,
-        // and we don't highlight the tray icon if the compiuter just
+        // Seems like Electron fires the `move` event right after the app boots,
+        // and we don't highlight the tray icon if the computer just
         // booted given that we don't show the window in this case
         tray.setHighlightMode('always');
       }
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -39,6 +39,10 @@ let recording = false;
 
 settings.init();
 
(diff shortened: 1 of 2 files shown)


app/src/main/index.js
```diff
@@ -39,6 +39,10 @@ let recording = false;
 
 settings.init();
 
+const wasOpenedAtLogin = app => {
+  return app.getLoginItemSettings().wasOpenedAtLogin;
+};
```
sindresorhus:
```js
const wasOpenedAtLogin = app => app.getLoginItemSettings().wasOpenedAtLogin;
```

app/src/main/index.js
```diff
@@ -294,7 +298,12 @@ menubar.on('after-create-window', () => {
       // The `move` event is called when the user reselases the mouse button
       // because of that, we need to move the window to it's expected position, since the
       // user will never release the mouse in the *right* position (diff.[x, y] === 0)
-      tray.setHighlightMode('always');
+      if (!wasOpenedAtLogin(app)) {
+        // Seems like electorn fires the `move` event right aftet the app boots,
```
sindresorhus: electorn => Electron

aftet => after

compiuter => computer

matheuss: Seems like my PhD in typos is still valid 😂

app/src/main/index.js
```diff
@@ -294,7 +296,12 @@ menubar.on('after-create-window', () => {
       // The `move` event is called when the user reselases the mouse button
       // because of that, we need to move the window to it's expected position, since the
       // user will never release the mouse in the *right* position (diff.[x, y] === 0)
-      tray.setHighlightMode('always');
+      if (!wasOpenedAtLogin(app)) {
```
matheuss: This introduces a bug – tray will never be highlighted if this check is in place and the app was opened on system startup 🐛

PR #218, opened 2017-08-26
2.0 Design updates

CodeTheory:
🎉 

Mostly styling updates to match @skllcrn's new designs but includes some JS changes that might be worth reviewing.

Artefact: https://kap-artifacts.now.sh/design/2.0


sindresorhus:
Wow. This looks super good 👌

- Clicking "Install"/"Uninstall" in the Preferences->plugins should disable the button while it's installing/uninstalling. It did that before this PR. Bonus points for a spinner inside the button indicating that it's working (I like the `zoom-in` effect here: http://lab.hakim.se/ladda/), but that can also be added later. We should try to get this PR landed asap.
- Clicking "Open plugins folder" in the Preferences doesn't seem to work.
- In the editor window, "✔️ Choose an option" should rather be "Export Destinations".
- In the aspect ratio window, when the first "width" textfield is focused and I press `Tab`, I would have expected it to focus the "height" textfield, but it takes two more `Tab` clicks to get there.
- In the aspect ratio window, when inputting something in the "width" or "height" textfields, the ratio box should switch to "Custom".
- In the aspect ratio window, selecting the "aspect width/height lock" button does nothing.
- In the aspect ratio window, selecting the "switch width/height" button does not correct switch the places.

CodeTheory:
Thanks for the comment @sindresorhus. These are all definitely something I want to address but this PR has been waiting long enough and I didn't want to make it even busier. I plan to tackle them all in their own PR apart from the `Open plugins folder` functionality which will be fixed today and changing the text to `Export destinations` ✅ 

As far as I see it with the aspect ratio feature, it can be improved a lot by changing the way it works quite a bit.
For example, instead of having preset pixel values for each aspect ratio, we'd use the current width or height and calculate the aspect ratio based on that value and change the other value to match.
For example, the width could be `1024` and I want 16:9 aspect ratio, I should be able to select `16:9` and have Kap work out what the height should be. Should just change the height to 768. From there, if I edit a value of width or height, the other value should change accordingly.

^ This is something I'm working on since yesterday.

sindresorhus:
@CodeTheory Agreed. Get this merged asap and we can work in perfecting it later.

albinekb:
Woho @CodeTheory great job ❤️

skllcrn:
It's a stunner 💅

Issue #229, opened 2017-09-12
Double tapping record button throws error


OmgImAlexis:
**Platform:** Darwin Bella.tardis 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64
**Kap Version:** Version 1.0.1 (1.0.1.170)
### Steps to reproduce
Double tap record button
### Current behaviour
Throws error
### Expected behaviour
Recording to be cancelled before it starts
### Workaround
N/A

![screen shot 2017-09-12 at 9 35 23 pm](https://user-images.githubusercontent.com/6525926/30325132-b229d462-9802-11e7-9a16-6aa302fa849c.png)

CodeTheory:
Thanks, I forgot to fix this!

albinekb:
Fixed in #239 🎉

PR #239, opened 2017-10-23
Fix spam recording

albinekb:
this includes https://github.com/wulkano/kap/pull/237

Rather fragile fix, but this is the easiest/fastest way. This makes it so if you spam the button, it will still show the correct state, and only allow 1 instance to record at once (since aperture has an internal lock, we can rely on it). The state management is really hard to debug, so I think this minor "hack" is OK to merge 👍  It can't have any bad side effects.

### DIFF
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -179,7 +179,7 @@ document.addEventListener('DOMContentLoaded', () => {
         // This delay is due to internal framework delays in aperture native code
         if (err.message.includes('stopRecording')) {
           log(`Recording not yet started, can't stop recording before it actually started`);
-          return
+          return;
         }
         ipcRenderer.send('will-stop-recording');
         reportError(err);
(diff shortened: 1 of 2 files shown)


app/src/renderer/js/main.js
```diff
@@ -175,6 +175,10 @@ document.addEventListener('DOMContentLoaded', () => {
         log(`Started recording after ${(Date.now() - past) / 1000}s`);
       })
       .catch(err => {
+        // This prevents the button from being reset, since the recording has not yet started
+        if (err.message.includes('stopRecording')) {
+          return
```
doot0: Can we add a `log` to describe what's happening at this point?

albinekb:
This fixes https://github.com/wulkano/kap/issues/229 👍

CodeTheory:
Sorry for closing and re-opening the PR. It was the only way to trigger a build with CircleCI by the looks of it (without committing)

PR #241, opened 2017-10-24
Add ability to cancel the export

albinekb:
- Currently only works for mp4, need to update all converters in scripts/converter since the others have multi step-conversions, which needs to be cancelled correctly
- Not designed yet, just added a simple button

Opening this now to start a discussion, is this an acceptable way to do it? Any ideas on how this could be done better? 🤔


sindresorhus:
Some quick thoughts on how to do the multi-step process: https://gist.github.com/sindresorhus/b78ff84403b6dfd4a3098c3f9ed81437 I wish we had some utility method on `p-cancelable` to combine multiple cancelables.

app/src/main/share-service-context.js
```diff
@@ -12,7 +12,13 @@ export default class ShareServiceContext {
 
     this.filePath = () => convert(Object.assign({}, exportOptions, {
       defaultFileName: this.defaultFileName
-    }));
+    })).catch(err => {
+      if (err.canceled) {
```
sindresorhus: I think this should be consistent and check the error name like done elsewhere here. Or maybe we should add a `canceled` property to `p-cancelable` errors? Could be nice for convenience.

app/src/main/convert.js
```diff
@@ -37,6 +37,23 @@ export default async function (exportOptions) {
     }
   });
 
+  ipcMain.once('cancel-processing', () => {
+    converter.cancel();
+    app.kap.mainWindow.send('export-progress', {text: ''});
```
sindresorhus: The above line is repeated 3 times in this file. Would be nice to dry it up.

app/src/main/convert.js
```diff
@@ -37,6 +37,23 @@ export default async function (exportOptions) {
...
+  });
+
+  try {
+    await converter;
+  } catch (err) {
+    if (converter.canceled) {
+      app.kap.mainWindow.send('export-progress', {text: ''});
+      err.canceled = true;
+      throw err;
```
sindresorhus: This line is moot as it could just continue to the below `throw`.

app/src/main/convert.js
sindresorhus: Actually, do we really need to call this here as it's already called on line 49?

app/src/renderer/js/main.js
```diff
@@ -331,6 +332,10 @@ document.addEventListener('DOMContentLoaded', () => {
     this.classList.toggle('is-active');
   };
 
+  progressBarCancelBtn.onclick = function () {
```
sindresorhus: Arrow-function, since `this` is not used.

sindresorhus: Actually, let's start using `.addEventListener()`. It's better.

albinekb: Will do 👌

app/src/main/share-service-context.js
albinekb: I can make it check for the name, and yes, that'd be nice 👍

albinekb:
Closing in favour of #335 🙏

PR #242, opened 2017-10-24
Prevent main window from being shown when launched at login

### LINKED ISSUE #208: Don't open on system startup

albinekb:
All creds to @leo for his helpful hint: https://github.com/wulkano/kap/issues/208#issuecomment-339049228

Fix #208 
Closes #209 ping @matheuss (I saw it after I PRd this one, but it's outdated by the looks of it 👍 )

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -13,6 +13,8 @@ import analytics from './analytics';
 import {applicationMenu, cogMenu} from './menus';
 import plugins from './plugins';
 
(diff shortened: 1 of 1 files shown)


albinekb:
Thanks for moving the comment 👍  I will do it this way in the future. I fixed the extra space so tests should pass now 👍

sindresorhus:
Looks good :)

PR #248, opened 2017-10-27
Upgrade electron-builder and electron

sindresorhus:
We were on a very old version of `electron-builder`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "start": "npm run app",
     "dist": "npm run build && build",
     "pack": "npm run build && build --dir",
-    "postinstall": "install-app-deps && npm run ffmpeg",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -55,6 +55,7 @@
     "appId": "com.wulkano.kap",
     "files": [
       "**/*",
+      "dist",
```
sindresorhus: Wasted a lot of time on this error message:

> Error: Application entry file "dist/main/index.js" does not exist. Seems like a wrong configuration.

Turns out the `dist` directory was not copied over for a reason. I'll dig more into it later, but for now, it's easy enough to manually add it here. Seems like an `electron-builder` issue.

PR #250, opened 2017-10-27
Make the editor window size inputs be consistent

sindresorhus:
The input text is now the same size as other inputs and centered.

## Before

screen shot 2017-10-27 at 21 15 42

## After

screen shot 2017-10-27 at 21 26 19


albinekb:
nice!

PR #251, opened 2017-10-27
Improve aspect ratio and size controls

### LINKED ISSUE #226: Rework aspect ratio selection
Aspect ratio selection shouldn't take priority of the width and height inputs. 
Currently if you select an aspect ratio, it will overwrite the width and height with a preset pixel value. 1600x900 for 16:9 for example. It will continue to help the width and height remain at a 16:9 ratio afterwards but it shouldn't have a preset value to begin with.

How I'd like to rework aspect ratios:
- Kap would open fresh with a 512x512 width and height (1:1 ratio but listed as custom)
- Changing the width or height wouldn't lock to the 1:1 ratio
- Selecting an aspect ratio would change the opposite value to the users active input to match the selected aspect ratio. For example, I've written `1280` as my width, I select `16:9` in the ratio selector and the height would change to `720`. Not overwriting the users width/height values to start with.
- As long as a ratio is selected in the ratio selector, the inputs will lock to that and the height/width values won't deviate from each other based on the ratio. 
- Selecting `Custom` ratio again will allow the user to type any value into both inputs without them locking to each other (however we should display what the ratio is for the user anyway, e.g. `Custom 512:200`)

Image example:
I've typed a custom width and height whilst selecting `Custom ratio` in the ratio selector. This shows me the ratio that the width/height pertain to. I can then lock this ratio for my width and height to change with.
image
_note: ignore stream stuff. Just an experiment in design that I can't be bothered to remove_

Would love to hear thoughts to expand on this.

CodeTheory:
Closes #226 

This PR reworks the aspect ratio/size selector to be more maneable and less predicting for users.
A written up explanation is available in the accompanying issue that this PR closes.

### DIFF
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -200,7 +200,7 @@ document.addEventListener('DOMContentLoaded', () => {
     }
 
     if (!hadMatch) {
-      const customRatio = document.getElementById('custom-ratio-option');
+      const customRatio = document.querySelector('#custom-ratio-option');
       customRatio.value = selectedRatio;
       customRatio.innerHTML = `Custom (${selectedRatio})`;
       customRatio.selected = true;
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -50,8 +50,11 @@ document.addEventListener('DOMContentLoaded', () => {
   // Initial variables
   let lastValidInputWidth = 512;
   let lastValidInputHeight = 512;
-  let aspectRatioBaseValues = [lastValidInputWidth, lastValidInputHeight];
   let hasUpdateNotification = false;
+  let height = 512;
+  let width = 512;
+  let selectedRatio = '1:1';
+  let lockedRatio = false;
 
   // Init dynamic elements
   if (app.kap.settings.get('recordAudio') === true) {
@@ -173,6 +176,39 @@ document.addEventListener('DOMContentLoaded', () => {
     }
   }
 
+  // Helper function for retrieving the simplest ratio, via the largest common divisor of two numbers (thanks @doot0)
+  function getLargestCommonDivisor(first, second) {
+    return (second === 0) ? first : getLargestCommonDivisor(second, first % second);
+  }
+
+  function getSimplestRatio(width, height) {
+    const lcd = getLargestCommonDivisor(width, height);
+    const denominator = width / lcd;
+    const numerator = height / lcd;
+    return `${denominator}:${numerator}`;
+  }
+
+  function setSelectedRatio(width, height) {
+    selectedRatio = getSimplestRatio(width, height);
+
+    const ratios = document.querySelectorAll('.aspect-ratio-selector option');
+    let hadMatch = false;
+    for (const ratio of ratios) {
+      if (ratio.value === selectedRatio) {
+        aspectRatioSelector.value = selectedRatio;
+        hadMatch = true;
+        break;
+      }
+    }
+
+    if (!hadMatch) {
+      const customRatio = document.getElementById('custom-ratio-option');
+      customRatio.value = selectedRatio;
+      customRatio.innerHTML = `Custom (${selectedRatio})`;
+      customRatio.selected = true;
+    }
+  }
+
   const handleRecord = function () {
     if (app.kap.editorWindow) {
       // We need to keep the window visible to show the shake animation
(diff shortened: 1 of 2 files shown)


app/src/renderer/js/main.js
```diff
@@ -55,8 +53,11 @@ document.addEventListener('DOMContentLoaded', () => {
   let monitoringIntervalId;
   let lastValidInputWidth = 512;
   let lastValidInputHeight = 512;
-  let aspectRatioBaseValues = [lastValidInputWidth, lastValidInputHeight];
   let hasUpdateNotification = false;
+  let height = 512;
+  let width = 512;
+  let selectedRatio = '1:1';
+  let lockedRatio = false;
```
sindresorhus: Do these really need to be global variables? And if yes, I think they at least should be scoped in an object so it's  clearer where they belong.

app/src/renderer/js/main.js
```diff
@@ -238,6 +239,39 @@ document.addEventListener('DOMContentLoaded', () => {
...
+        aspectRatioSelector.value = selectedRatio;
+        hadMatch = true;
+        break;
+      }
+    }
+
+    if (!hadMatch) {
+      const customRatio = document.getElementById('custom-ratio-option');
+      customRatio.value = selectedRatio;
+      customRatio.innerHTML = `Custom (${selectedRatio})`;
```
sindresorhus: `.innerHTML` => `.textContent`

app/src/renderer/js/main.js
```diff
@@ -297,14 +333,16 @@ document.addEventListener('DOMContentLoaded', () => {
...
-        lastValidInputWidth = tmp[0];
-        inputWidth.value = tmp[0];
-      }
+    height = this.value;
+
+    if (lockedRatio) {
+      width = (selectedRatio.split(':')[0] / selectedRatio.split(':')[1]) * this.value;
+      inputWidth.value = Math.round(width);
+      return;
     }
```
sindresorhus: This logic is almost the same as on line 309 and 379. Could be DRY'ed up.

app/src/renderer/js/main.js
```diff
@@ -238,6 +239,39 @@ document.addEventListener('DOMContentLoaded', () => {
...
+    for (const ratio of ratios) {
+      if (ratio.value === selectedRatio) {
+        aspectRatioSelector.value = selectedRatio;
+        hadMatch = true;
+        break;
+      }
+    }
+
+    if (!hadMatch) {
+      const customRatio = document.getElementById('custom-ratio-option');
```
sindresorhus: Just use `querySelector`. `getElementById` has no useful speed benefit.

app/src/renderer/js/main.js
CodeTheory: They're used in multiple places so I think they have to be. Will assign them to an object for cleanliness 👌

app/src/renderer/js/main.js
CodeTheory: I'm having a slight bit of trouble with this one. 
It would be easy to break this logic into a function but there's another part of the code that has _very_ similar logic to the parts we want to DRY up.
(for reference, here's the logic https://github.com/wulkano/kap/pull/251/files#diff-a9a03e39450b8c7cd627d10635075959R338)
I'm wondering, because we want to create a function for working out the `width` via the selected ratio, do we want another function for the height (because it wouldn't make sense to keep that logic isolated whilst we break out the former), or is there a way we could create one function for the two.

What would be best here?

CodeTheory:
Github hides reviews for outdated files, but there's one review that's still relevant: https://github.com/wulkano/kap/pull/251#discussion_r147531078

app/src/renderer/js/main.js
sindresorhus: Yeah, you're right. Not that easy. Really, all this would have been much nicer if the DOM was automatically synced from the model instead of us having to manually update it in all the places, like with React or Vue.

Would at least be a little bit nicer if all the ratio logic was encapsulated in a class. That would reduce the need for global state and make it easier to reuse some of the logic. Then we would store the ratio as an object instead of string, and just have a property that renders our internal `{width: 16, height: 9}` to `"16:9"`.

`dimensions.height = (dimensions.ratio.split(':')[1] / dimensions.ratio.split(':')[0]) * this.value;` internally could simply be `this.height = (this.ratio.height / this.ratio.width) * this.value;`

sindresorhus:
- I get a `RangeError: Maximum call stack size exceeded` when removing the contents of one of the inputs.
- When `16:9` is selected it should be possible to have `1` width and `1` height.
- Select `5:3` in the dropdown, then `20` width and `16` height in the inputs, then click the switch button, and notice that each time you switch the number in the inputs decreases.

CodeTheory:
As per the last two commits:
- Switching values also switches the aspect ratio (as discussed with @skllcrn) and doesn't decrease any values
- stopped `Maximum call stack size exceeded` from occurring on removing a value from the height/width inputs when the ratio isn't locked

Things that need to be done now:
- Apply a minimum value to `width`/`height` when the aspect ratio is locked and make the mimum the corresponding numerator/denominator
- When `oninput` is called with no value in the width/height inputs we need to update the aspect ratio selector with the last valid input value (having a bit of trouble with this at the moment)

sindresorhus:
Another issue. Select `3:2`, then clip the flip button and you'll end up with this:

screen shot 2017-10-31 at 23 29 31

CodeTheory:
Also, changing the width/height with the cropper open no longer changes the cropper size. But I don't think that's from this PR 🤔

sindresorhus:
Merging this. We can do follow-up PRs with improvements.

kevva:
@CodeTheory, I think it is.

sindresorhus:
@kevva No, it worked before https://github.com/wulkano/kap/commit/0464f23fef059201c47aea5dc54a10e3330509d7, so the issue was introduced in this PR.

PR #254, opened 2017-10-29
Add video codec preference

### LINKED ISSUE #220: Include codec option in settings
See wulkano/aperture#45

sindresorhus:
Fixes #220

screen shot 2017-10-29 at 16 25 59

### DIFF
--- a/app/src/renderer/js/preferences.js
+++ b/app/src/renderer/js/preferences.js
@@ -16,6 +16,22 @@ const settingsValues = app.kap.settings.getAll();
 // Observers that should be disposed when the window unloads
 const observersToDispose = [];
 
+const initVideoCodecSelector = () => {
+  const videoCodecSelector = $('.js-video-codec-selector');
+
+  for (const [id, name] of aperture.videoCodecs) {
+    const option = document.createElement('option');
+    option.value = id;
+    option.text = name;
+    option.selected = id === app.kap.settings.get('videoCodec');
+    videoCodecSelector.add(option);
+  }
+
+  videoCodecSelector.addEventListener('change', function () {
+    app.kap.settings.set('videoCodec', this.value);
+  });
+};
+
 document.addEventListener('DOMContentLoaded', async () => {
   // Element definitions
   const allowAnalyticsCheckbox = $('#allow-analytics');
@@ -59,6 +75,8 @@ document.addEventListener('DOMContentLoaded', async () => {
   if (settingsValues.recordAudio === true) {
     audioInputDeviceSelector.value = settingsValues.audioInputDeviceId;
   }
+
+  initVideoCodecSelector();
 
   const tabs = $j('.prefs-nav > a');
   tabs.on('click', function (event) {
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -14,7 +14,8 @@ const DEFAULTS = {
   highlightClicks: false,
   fps: 30,
   recordAudio: false,
-  audioInputDeviceId: 'none'
+  audioInputDeviceId: 'none',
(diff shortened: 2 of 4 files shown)


sindresorhus:
At the end of implementing this, I discovered that Chrome doesn't actually support playing HEVC yet... Not even in Chrome Canary https://caniuse.com/#feat=hevc So we can't preview the clip in the editor.

Not sure where to go from here. We could wait on adding HEVC support, but that would be too bad as HEVC video files have much better quality, and HEVC support is good for marketing. Another solution could be to generate a quick video preview using FFMPEG and show that instead. We already use FFMPEG for everything else, so it's just the preview we have to fix. That way we could support other codecs like ProRes too (almost uncompressed video).

Thoughts?

app/src/renderer/js/preferences.js
```diff
@@ -208,7 +226,7 @@ document.addEventListener('DOMContentLoaded', async () => {
   };
 
   fpsSlider.onchange = function () {
-    app.kap.settings.set('fps', this.value);
+    app.kap.settings.set('fps', Number(this.value));
```
sindresorhus: For some good news, the Aperture upgrade* helped catch this subtle bug.

*The Codable stuff: https://github.com/wulkano/aperture/commit/0193e72e2a78ceaeae5178cde852686113c6da59

vadimdemedes:
> Another solution could be to generate a quick video preview using FFMPEG and show that instead.

Does it take long for ffmpeg to generate such preview?

sindresorhus:
@vadimdemedes I haven't tried yet, but FFMPEG has a fast mode, so shouldn't be too bad. I'll try later.

app/src/renderer/js/preferences.js
```diff
@@ -16,6 +16,22 @@ const settingsValues = app.kap.settings.getAll();
 // Observers that should be disposed when the window unloads
 const observersToDispose = [];
 
+function initVideoCodecSelector() {
```
vadimdemedes: Can be an arrow function, `this` isn't used.

vadimdemedes:
In case you change your mind about merging this, could you extract the changes to support Aperture v3 into a separate PR that we can merge asap?

sindresorhus:
@vadimdemedes Yup. I should have done that in the first place. I've upgraded Aperture on `master` now.

app/src/renderer/js/preferences.js
sindresorhus: Done

albinekb:
>generate a quick video preview using FFMPEG and show that instead

I like this, HEVC is good for marketing, we could add a red `*` mark and link that to a text about that the video playing in the editor is not full quality because of chrome not supporting it 👍

sindresorhus:
This will not be part of the first 2.0 beta.

skllcrn:
I would also assume that HEVC would have great performance benefits while capturing and processing, as the file format is far nimbler.

sindresorhus:
> I would also assume that HEVC would have great performance benefits while capturing and processing, as the file format is far nimbler.

No, actually the opposite. It requires more resources while capturing and processing as it requires more advanced compression algorithms to achieve higher quality in the same file size.

albinekb:
>No, actually the opposite.

Is that really true for the 2017 hardware? Isn't it all hardware accelerated? 🤔

sindresorhus:
@albinekb It's hardware accelerated, yes, but so is H264. HEVC requires more resources from the dedicated encoder, though. I'm not saying it will slow down your machine more, just that I really doubt it will be faster.

sindresorhus:
Since we can't easily preview HEVC, we've decided to drop the codec option for simplicity and rather move to HEVC sometime in the future when it's ready: https://github.com/wulkano/kap/issues/341

Issue #143, opened 2017-02-12
Touch Bar support


sindresorhus:
Depends on https://github.com/electron/electron/pull/8095.

Would be cool to display the elapsed recording time and a stop button in the Touch Bar.

If you don't have a native Touch Bar, you can use: https://red-sweater.com/touche/

skllcrn:
https://github.com/electron/electron/issues/7781#issuecomment-278121145

Khaledgarbaya:
it is here finally https://electron.atom.io/blog/2017/03/08/touch-bar-support but the version of electron still in beta, I would love to work on this once you guys decide to move to Electron **v1.6.3+** and of course when the version is out of the beta

matheuss:
@Khaledgarbaya feel free to bump the dependency to 1.6.3 🙌

Khaledgarbaya:
@matheuss @skllcrn any preferences on what you want to show in the TouchBar. 
I was thinking about start/stop recording mainly

sindresorhus:
Note the limitation that we can only be in the Touch Bar when the Kap window is focused, so not during recording.

PR #255, opened 2017-10-29
Touch Bar integration

vadimdemedes:
Fixes https://github.com/wulkano/kap/issues/143.

The initial or main touch bar:

touch bar shot 2017-10-29 at 12 49 40 pm

Customization of aspect ratio or size:

touch bar shot 2017-10-29 at 12 49 47 pm
touch bar shot 2017-10-29 at 12 49 51 pm

Touch bar when cropping is active:

touch bar shot 2017-10-29 at 12 49 56 pm

Editor's touch bar:

touch bar shot 2017-10-29 at 12 50 10 pm

Plugins reveal upon clicking a format:

touch bar shot 2017-10-29 at 12 51 51 pm

### DIFF
--- a/app/src/main/touchbar.js
+++ b/app/src/main/touchbar.js
@@ -1,0 +1,125 @@
+import {TouchBar} from 'electron';
+import plugins from './plugins';
+
+const {TouchBarButton, TouchBarPopover} = TouchBar;
+
+const aspectRatioToSize = {
+  '16:9': '1600x900',
+  '5:4': '1280x1024',
+  '5:3': '1280x768',
+  '4:3': '1024x768',
+  '3:2': '480x320',
+  '1:1': '512x512',
+  Custom: 'Custom'
+};
+
+const createAspectRatioPopover = ({onChange}) => {
+  const buttons = [
+    '16:9',
+    '5:4',
+    '5:3',
+    '4:3',
+    '3:2',
+    '1:1',
+    'Custom'
+  ].map(aspectRatio => {
+    return new TouchBarButton({
+      label: aspectRatio,
+      click: () => onChange(aspectRatioToSize(aspectRatio))
+    });
+  });
+
+  return new TouchBarPopover({
+    label: 'Aspect Ratio',
+    items: new TouchBar(buttons)
+  });
+};
+
+const createSizePopover = ({onChange}) => {
+  const sizeButtons = [
+    '1600x900',
+    '1280x1024',
+    '1280x768',
+    '1024x768',
+    '480x320',
+    '512x512',
+    'Custom'
+  ].map(size => {
+    return new TouchBarButton({
+      label: size,
+      click: () => onChange(size)
+    });
+  });
+
+  return new TouchBarPopover({
+    label: 'Size',
+    items: new TouchBar(sizeButtons)
+  });
+};
+
+export const createMainTouchbar = ({onSizeChange, onCrop}) => {
+  const aspectRatioPopover = createAspectRatioPopover({onChange: onSizeChange});
+  const sizePopover = createSizePopover({onChange: onSizeChange});
+  const cropButton = new TouchBarButton({
+    label: 'Crop',
+    backgroundColor: '#7e51e5',
+    click: onCrop
+  });
+
+  return new TouchBar([
+    aspectRatioPopover,
+    sizePopover,
+    cropButton
+  ]);
+};
+
+export const createCropTouchbar = ({onSizeChange, onRecord}) => {
+  const aspectRatioPopover = createAspectRatioPopover({onChange: onSizeChange});
+  const sizePopover = createSizePopover({onChange: onSizeChange});
+  const recordButton = new TouchBarButton({
+    label: 'Record',
+    backgroundColor: '#ff5050',
+    click: onRecord
+  });
+
+  return new TouchBar([
+    aspectRatioPopover,
+    sizePopover,
+    recordButton
+  ]);
+};
+
+const shareServices = {};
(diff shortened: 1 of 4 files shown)


sindresorhus:
During export, it shows the following in the Touch Bar (because the main window is open), but it shouldn't as it's exporting, not showing the main controls (We really shouldn't have reused the main windows for exporting, I think this is changing with the new design though, @skllcrn ?):

screen shot 2017-10-30 at 13 02 56

app/src/main/touchbar.js
```diff
@@ -0,0 +1,125 @@
...
+    '5:4',
+    '5:3',
+    '4:3',
+    '3:2',
+    '1:1',
+    'Custom'
+  ].map(aspectRatio => {
+    return new TouchBarButton({
+      label: aspectRatio,
+      click: () => onChange(aspectRatioToSize(aspectRatio))
```
sindresorhus: `aspectRatioToSize` is not a function.

app/src/main/touchbar.js
```diff
@@ -0,0 +1,125 @@
...
+
+const aspectRatioToSize = {
+  '16:9': '1600x900',
+  '5:4': '1280x1024',
+  '5:3': '1280x768',
+  '4:3': '1024x768',
+  '3:2': '480x320',
+  '1:1': '512x512',
+  Custom: 'Custom'
+};
```
sindresorhus: This should be a `Map`. You could then just use `[aspectRatioToSize.keys()...]` on line 17 and `[aspectRatioToSize.values()...]` on line 40.

app/src/renderer/js/editor.js
```diff
@@ -232,6 +233,21 @@ document.addEventListener('DOMContentLoaded', () => {
 
   registerExportOptions();
 
+  ipcRenderer.on('run-plugin', (e, pluginName, format) => {
+    const service = shareServices.find(service => service.pluginName === pluginName);
+
+    service.run({
```
sindresorhus: Can you extract this into a function and reuse it on line 218?

app/src/renderer/js/main.js
```diff
@@ -342,6 +346,13 @@ document.addEventListener('DOMContentLoaded', () => {
     }
   };
 
+  aspectRatioSelector.onchange = handleSizeChange;
```
sindresorhus: Prefer `addEventListener` for all the event handlers.

https://stackoverflow.com/a/35093997/64949

app/src/main/touchbar.js
```diff
@@ -0,0 +1,125 @@
...
+
+const shareServices = {};
+
+plugins.getShareServices().forEach(service => {
+  service.formats.forEach(format => {
+    shareServices[format] = shareServices[format] || [];
+    shareServices[format].push(service);
+  });
+});
```
sindresorhus: Can you move this into `plugins.js` as a utility method? Might be useful for other things too and better to keep that logic where it belongs.

sindresorhus:
Filename: `touchbar.js` => `touch-bar.js`

app/src/main/touchbar.js
```diff
@@ -0,0 +1,125 @@
...
+const createFormatPopover = ({format, services, onSelectPlugin}) => {
+  const serviceButtons = services.map(service => {
+    return new TouchBarButton({
+      label: service.title,
+      click: () => onSelectPlugin(service.pluginName, format)
+    });
+  });
+
+  return new TouchBarPopover({
+    label: format.toUpperCase(),
```
sindresorhus:
```js
format.toUpperCase().replace('WEBM', 'WebM')
```

Because it's `WebM` not `WEBM`.

Better yet, move this to `plugins.js` as a `titleFromFormat()` method (Or a better name?) which does this.

sindresorhus:
If anyone wanna help test this out you can download a build from here: https://kap-artifacts.now.sh/touch-bar

And if you don't have a Mac with a Touch Bar, you can use my [Touch Bar Simulator](https://sindresorhus.com/touch-bar-simulator/).

sindresorhus:
When I hit the "Record" button in the Touch Bar, the Touch Bar continues to show "Aspect Ratio", "Size", "Record", even though I'm currently recording and can't do those actions anymore.

kevva:
> When I hit the "Record" button in the Touch Bar, the Touch Bar continues to show "Aspect Ratio", "Size", "Record", even though I'm currently recording and can't do those actions anymore.

Yes, I also found that confusing. Also, when saving/using a plugin, the initial touch bar is shown. I'd expect the editor touch bar or maybe some state button.

app/src/renderer/js/main.js
kevva: We're using `.onchange`, `.onclick` everywhere else. But I agree we should them too later on.

sindresorhus: Yes, but better to do the right thing for new code.

sindresorhus:
I'm gonna merge this. We can continue improvements as new PRs to ease co-operation.

sindresorhus:
Thanks for working on this Vadim. It's really great for marketing :)

app/src/renderer/js/main.js
sindresorhus: This should really be a linting rule: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/117

skllcrn:
Regarding reusing the main window for exporting I completely agree @sindresorhus, especially now that you might want to do multiple exports at once (right now they queue up in the background but theres no UI to support it) or record while having an editor window open.

PR #259, opened 2017-10-30
Extract plugin functions into utility methods and code style tweaks

kevva:
Just a continuation of https://github.com/wulkano/kap/pull/255 with some code style fixes.

### DIFF
--- a/app/src/main/plugins.js
+++ b/app/src/main/plugins.js
@@ -125,6 +125,28 @@ class Plugins {
 
     return ret;
   }
+
+  getShareServicesPerFormat() {
+    const ret = {};
+
+    for (const service of this.getShareServices()) {
+      for (const format of service.formats) {
+        ret[format] = ret[format] || [];
+        ret[format].push(service);
+      }
+    }
+
+    return ret;
+  }
+
+  prettifyFormat(format) {
+    const ret = format.toUpperCase();
+    const formats = new Map([
+      ['WEBM', 'WebM']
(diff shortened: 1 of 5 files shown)


app/src/main/plugins.js
```diff
@@ -125,6 +125,28 @@ class Plugins {
...
+    }
+
+    return ret;
+  }
+
+  prettifyFormat(format) {
+    const ret = format.toUpperCase();
+    const formats = new Map([
+      ['WEBM', 'WebM']
+    ]);
```
sindresorhus: I know it's moot, but better for readability to just make a manual mapping of all of them, even if it works fine to just uppercase. Most consistent. Less differing code-paths.

app/src/main/touch-bar.js
```diff
@@ -3,29 +3,21 @@ import plugins from './plugins';
...
-  const buttons = [
-    '16:9',
-    '5:4',
-    '5:3',
-    '4:3',
-    '3:2',
-    '1:1',
-    'Custom'
-  ].map(aspectRatio => {
+  const buttons = Array.from(aspectRatioToSize.keys()).map(aspectRatio => {
```
sindresorhus: Why not `[aspectRatioToSize.keys()...]`?

kevva: No reason. Forgot that it's supported now :).

sindresorhus:
Looks good :)

PR #261, opened 2017-10-30
Fix dragging timeline marker moves window

### LINKED ISSUE #260: Improve trim handles
Sometimes when I drag the trim handles I end up dragging the window instead, this should not happen :+1: 

Maybe we should do like how we handle these kinds of issues in mobile: 
- Add a transparent "box" that acts as the handle, and have the thin white one just for "looks"

sindresorhus:
Fixes #260

Just disables dragging for the whole timeline, which makes sense as we don't want any window dragging there ever.

### DIFF
--- a/app/src/renderer/css/editor.css
+++ b/app/src/renderer/css/editor.css
@@ -158,6 +158,7 @@ video {
   height: 100%;
   transform: translateX(-50%);
   top: 6px;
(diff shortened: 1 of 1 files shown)


albinekb:
Try it out @ https://kap-artifacts.now.sh/timeline-marker-grab

PR #262, opened 2017-10-30
Animate the menubar icon when recording starts

### LINKED ISSUE #162: Animated menu bar icon
Indicate that Kap is preparing to record with an animated menu bar icon

sindresorhus:
Fixes #162

@onordbo @skllcrn Let me know if the speed is ok. It can easily be adjusted.


app/src/main/index.js
```diff
@@ -197,9 +197,35 @@ function resetTrayIcon() {
   mainWindow.setAlwaysOnTop(false);
 }
 
+const animateIcon = () => new Promise(resolve => {
+  const interval = 40;
+  let i = 0;
+
+  const next = () => {
+    setTimeout(() => {
+      const number = String(i++).padStart(5, '0');
```
albinekb: I think something like `frameNumber` would be more descriptive 👍

albinekb: Maybe better to have a function that generates the filename from the frame number + animation name?

albinekb:
The left icon looks kinda blurry compared to the non-animated one 🤔  Are the image sizes correct?

app/src/main/index.js
sindresorhus: I considered it, but decided not to since it's not going to be reused anyway and it would be hardcoded to our use-case here. What if we in the future need an animation without `Template` or using a different naming scheme. Probably better to wait until we have a need. And it's only a few characters. Also, in the future, I want rather refactor this into a module that accepts an APNG file and then animates it. That way we wouldn't need all these individual images.

skllcrn:
Agree with @albinekb that the speed needs some tweaking, from the looks of it we could bump it straight up to double the speed. Here's an animation for reference: 

kapture 2017-04-12 at 1 18 32

sindresorhus:
Done.

The animation I got from you only bounces once though. Not like in the GIF above.

skllcrn:
That's okay for now, was only meant as a reference for speed. @onordbo might want to help tweak this later.

skllcrn:
Right now the animations starts quite a bit after hitting the recording button, I'm guessing it waits on the same event that the main window waits for to minimise. Would it be possible to make the animation trigger on clicking the record button, and hide the main window simultaneously @sindresorhus?

skllcrn:
If you stop a recording before the loading animation concludes the icon is stuck on the last shown frame of the animation @sindresorhus.

sindresorhus:
@skllcrn Ok, will fix after #265 is merged.

PR #266, opened 2017-10-30
Install and uninstall plugins in the main process

### LINKED ISSUE #257: Error when Preferences window is closed/blurred during a plugin install
Install a plugin, then close the Preferences window, and when the plugin finishes installing, an error is thrown.

screen shot 2017-10-30 at 20 49 31

kevva:
This avoids calling functions in the main process from the renderer if the renderer is closed.

Fixes #257.


sindresorhus:
Working great! Thanks for fixing this difficult issue @kevva :)

PR #267, opened 2017-10-30
Add `Discard` button to editor Touch Bar

### LINKED ISSUE #246: Discard Kapture improvements
When closing the editor window, you get this prompt:
image

Which is good, but when using Kap a lot, it feels like an unnecessary step.

I can think of some ways to improve this:
- Only ask the user if nothing was exported (will not help in all cases)
- Have a button visible somewhere in the UI that says `discard` and when pressed, discards the recording & closes the editor without further questions
- Have a secret key (like cmd) that if you hold it and press the X it will not ask (bad UX)

This would make it easy to quickly discard a kapture, for example if you messed up and want to do a re-take, or when you're done exporting

Personally I like the idea to have a button to discard directly.

Thoughts?

kevva:
This partly fixes #246 by adding a "Discard" button to the touch bar. Not 100% sure about the color since it's the same as "Record" which might cause confusion.

touch bar shot 2017-10-30 at 19 48 40

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -51,6 +51,7 @@ const cropTouchbar = createCropTouchbar({
 });
 
 const editorTouchbar = createEditorTouchbar({
+  onDiscard: () => discardVideo(),
   onSelectPlugin: (pluginName, format) => editorWindow.webContents.send('run-plugin', pluginName, format)
 });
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add some space between `Discard` and the other buttons? We don't want people tapping it by mistake.

kevva:
touch bar shot 2017-10-30 at 21 00 40

sindresorhus:
@kevva Can you fix the merge conflict?

kevva:
Done.

sindresorhus:
Great! I decided to move it all the way to the right:

screen shot 2017-10-31 at 04 08 04

kevva:
Like `flexible` too! I thought `large` looked somewhat out of place.

albinekb:
This is awesome.

PR #269, opened 2017-10-30
Fix round corners in editor on macOS 10.13

### LINKED ISSUE #264: Missing rounded corners in the Editor
The Editor is being square and doesn't really get with my circle. 

On a serious note though, the top left and right corners of the Editor are missing some sweet rounded corners on the latest master.

sindresorhus:
Seems like some kind of Electron bug, as usual. If a background is set on `body`, it overflows the native window round corners...

Fixes #264

Test: https://kap-artifacts.now.sh/fix-editor-border-radius


skllcrn:
Ha, interesting. The fix works perfectly though!

PR #273, opened 2017-10-31
Add play/pause Touch Bar button

kevva:
Would probably look better with an icon instead of a text label.

touch bar shot 2017-10-31 at 11 05 24


sindresorhus:
Looks great, as always :D

sindresorhus:
> Would probably look better with an icon instead of a text label.

Yeah, we can do this in the future: https://github.com/wulkano/kap/commit/7c514a919090a1febfad1b7e6c1216f6246daf49

PR #281, opened 2017-10-31
Improve draggable slider thumbs

CodeTheory:
Bit of a large commit, but essentially this is some polish upon the trimmer to make it work as per the design.

### DIFF
--- a/app/src/renderer/css/editor.css
+++ b/app/src/renderer/css/editor.css
@@ -60,10 +60,17 @@ main.content {
     .c-progress.editor-progress {
       width: 60%;
       left: 50%;
-      padding: 0 7px;
       transform: translateX(-50%);
       bottom: 26px;
       border-radius: 4px;
+    }
+
+    .timeline-markers {
+      left: 50%;
+      transform: translateX(-50%);
+      width: 60%;
+      bottom: 20px;
+      height: 16px;
     }
   }
 }
@@ -146,19 +153,23 @@ video {
   transition: 200ms ease-out transform;
 }
 
-.timeline-container {
-  cursor: pointer;
+.timeline-container { cursor: pointer; }
+
+.timeline-container:hover .timeline-markers .slider::-webkit-slider-thumb {
+  width: 4px;
+  height: 24px;
 }
 
 .timeline-markers {
   opacity: 0;
   transition: opacity .12s ease-in-out;
-  position: absolute;
-  width: 60%;
+  width: 100%;
   height: 100%;
-  transform: translateX(-50%);
-  top: 6px;
+  position: absolute;
+  left: 0;
+  bottom: 0;
   -webkit-app-region: no-drag;
+  transition: .12s ease-in-out;
 }
 
 section.output {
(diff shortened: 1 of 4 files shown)


app/src/renderer/js/editor.js
```diff
@@ -92,7 +94,7 @@ document.addEventListener('DOMContentLoaded', () => {
     if (preview.duration) {
       const timeAtEvent = getTimestampAtEvent(event, preview.duration);
       previewTimeTip.style.left = `${event.pageX}px`;
-      previewTimeTip.textContent = `${moment().startOf('day').milliseconds(timeAtEvent * 1000).format('m:ss.SS')}`;
+      previewTimeTip.textContent = `${moment().startOf('day').milliseconds(timeAtEvent * 1000).format('m:ss.SS')} (${moment().startOf('day').milliseconds(currentPreviewDuration * 1000).format('m:ss.SS')})`;
```
albinekb: I think we should have a common function to format these, I also think it's quicker to reuse the same moment objects rather than creating new ones, but this could be for another PR if we want to get this in quick 👍

CodeTheory: Agreed, but for the beta, this should be fine 😄

sindresorhus:
The initial trim handlers are a bit small before you hover the timeline. Not really clear that they are for trimming.

app/src/renderer/js/editor.js
```diff
@@ -105,6 +107,13 @@ document.addEventListener('DOMContentLoaded', () => {
     }
   }
 
+  function getTrimmedVideoDuration() {
+    const inValue = getTrimmerValue(trimmerIn);
+    const outValue = getTrimmerValue(trimmerOut);
+    currentPreviewDuration = outValue - inValue;
```
albinekb: this function mutates the external var `currentPreviewDuration`, better to make this one just 
```
return outValue - inValue
```
and mutate it where you call it further down
```
currentPreviewDuration = getTrimmedVideoDuration()
```

skllcrn:
Should be clearer when we get the helper text implemented. I'll also keep tweaking the design based on beta feedback. 

editor

sindresorhus:
The initial handles are much smaller here than in the above screenshot though.
screen shot 2017-10-31 at 23 59 55

app/src/renderer/css/editor.css
```diff
@@ -146,19 +153,23 @@ video {
   transition: 200ms ease-out transform;
 }
 
-.timeline-container {
-  cursor: pointer;
+.timeline-container { cursor: pointer; }
```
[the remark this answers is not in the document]
CodeTheory: I _only_ single line CSS that has one property because it's easier to scan through and helps with merge conflicts around it 😅

skllcrn:
If we can merge this in for now, and make the design tweaks in the next beta release that'd be great.  Then I've got time to get it right.

app/src/renderer/js/editor.js
sindresorhus: @albinekb Can you do a follow-up PR with some improvements maybe?

CodeTheory: Sorry, I missed this entirely.

Issue #3, opened 2016-10-04
App window shows for a short period of time when starting a record


CodeTheory:

matheuss:
I just looked into this and here's what I concluded:

I put `console.log(Date.now())`s in the following locations:

https://github.com/wulkano/aperture.js/blob/99110e848368af117ee573ffe693433d0eff7ec4/index.js#L55

https://github.com/wulkano/kap/blob/fb1e5b058277bb55f6c156d1a679150ea8b5a878/app/src/renderer/js/index.js#L120

https://github.com/wulkano/kap/blob/fb1e5b058277bb55f6c156d1a679150ea8b5a878/app/src/main/index.js#L281

Result:

<img width="686" alt="screen shot 2016-10-13 at 5 05 00 pm" src="https://cloud.githubusercontent.com/assets/4721750/19365247/405e95ee-9168-11e6-9474-32c7a7d89b0c.png">

![kapture 2016-10-13 at 16 58 15](https://cloud.githubusercontent.com/assets/4721750/19365250/442daac0-9168-11e6-9757-61375f6c6b0f.gif)

Kap receives the `recording started` signal from `aperture` 7 ms after it starts and then the `mainWindow.hidden()` is called 1 ms later.

So: it's probably something on the `aperture` or Swift side and it's going to need some time to debug. Any help is more than welcome 😄

Khaledgarbaya:
is the waiting for recording started really needed? the window could be hidden and still wait for the signal. I might take a look into this tonight

sindresorhus:
@matheuss I'm not surprised. The API is async, so it's not going to resolve the promise instantly. The native API reports exactly when the screen recording starts, so there's going to be some lag.

sindresorhus:
@Khaledgarbaya It usually takes 1-2 seconds for the screen recording to start (Not us, but the system), so it's an indicator you're not being recorded yet. I can't really think of a better solution right now, so I guess we can at least do what you suggest for now.

matheuss:
@Khaledgarbaya I'm not sure if I like your idea 😕 If you want the window to be hidden, you can just leave it docked to the menubar. 

I will try and hack something during [2.0.0](https://github.com/wulkano/kap/milestone/3) 🕵️

matheuss:
Ok, just discussed with @skllcrn and we have an idea: we'll just hide the window when you click to record, and then the menubar icon will be more useful: we can show the time, a pause button etc. Something like CloudApp:

[...]

PR #283, opened 2017-10-31
Remove Kap window from recording

vadimdemedes:
Fixes https://github.com/wulkano/kap/issues/3. Depends on https://github.com/wulkano/aperture/pull/52.

Regarding https://github.com/wulkano/kap/compare/cut-start?expand=1#diff-3ff6aef2849fdd54608f21a77278768aR21, I tried stripping time without re-encoding (`-c copy`), but the produced video wouldn't play in the editor, saying that it's unsupported. If anyone could help with the right ffmpeg command, that'd be awesome!

Although, if we're going to merge https://github.com/wulkano/kap/pull/254, we would need to generate a preview of HEVC anyway, so maybe we can leave this command as-is. Thoughts?

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -15,7 +15,7 @@
   "main": "dist/main/index.js",
   "dependencies": {
     "ajv": "^5.1.3",
-    "aperture": "^3.0.0",
+    "aperture": "github:vadimdemedes/aperture#output-start-time",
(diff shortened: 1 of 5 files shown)


vadimdemedes:
Looks like it's cutting too much from the beginning. Don't know the reason yet.

PR #284, opened 2017-10-31
Add CSS linter

sindresorhus:
There are many people contributing to the CSS parts now. This will make it easier to enforce a clean and maintainable codebase.

This touches a lot of CSS, so let's have code style discussions later.

### DIFF
--- a/app/src/renderer/css/_base.css
+++ b/app/src/renderer/css/_base.css
@@ -3,11 +3,10 @@
 /* Set base colour and font stack */
 body {
   color: $color-primary;
-  font-family: -apple-system, BlinkMacSystemFont, "Helvetica Neue";
+  font-family: -apple-system, BlinkMacSystemFont, 'Helvetica Neue';
   overflow: hidden;
-  letter-spacing: -.010rem;
+  letter-spacing: -0.01rem;
   border-radius: 5px;
-  overflow: hidden;
   font-size: 1.4rem;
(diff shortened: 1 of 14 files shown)


sindresorhus:
Happy to change anything. As I said on Slack, this is just a good start.

sindresorhus:
@CodeTheory What do you think of this rule: https://github.com/stylelint/stylelint/blob/c85372c07239e2d0688abcda1eb1b390e649bfba/lib/rules/no-descending-specificity/README.md ? I had to disable it here as it had too many violations.

Same with https://github.com/stylelint/stylelint/blob/c85372c07239e2d0688abcda1eb1b390e649bfba/lib/rules/declaration-no-important/README.md

CodeTheory:
Both good rules @sindresorhus. First avoids a lot of un-necessary hassle and the second is just a good rule of thumb (same with the hassle).

sindresorhus:
Alright. I'll try to enable them at some point. When we can afford some churn.

CodeTheory:
If you don't around to it, I'll give it a go @sindresorhus. Should we open a separate issue for them or just try to remember.

sindresorhus:
Feel free to give it a go. I'm just a CSS noob.

sindresorhus:
I opened an issue: https://github.com/wulkano/kap/issues/285

PR #287, opened 2017-11-01
Prevent updating the Touch Bar on each trim handler drag

kevva:
Fixes an issue where the drag handlers felt sluggish because `toggle-play` was called on every drag.


sindresorhus:
Thanks for fixing :)

PR #288, opened 2017-11-02
Update trimming handles to match design

onordbo:
Testing artifact: https://kap-artifacts.now.sh/fix-handles-again

### DIFF
--- a/app/src/renderer/css/components/_slider.css
+++ b/app/src/renderer/css/components/_slider.css
@@ -5,36 +5,38 @@
 */
 
 .slider {
-  -webkit-appearance: none;
-  background: transparent;
-  position: absolute;
-  pointer-events: none;
-  border: none;
-  height: 1.6rem;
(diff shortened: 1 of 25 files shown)


app/src/renderer/css/components/_slider.css
```diff
@@ -30,10 +30,10 @@
   &::-webkit-slider-thumb {
     -webkit-appearance: none;
     pointer-events: auto;
-    width: 2px;
-    height: 8px;
+    width: 4px;
+    height: 16px;
     border-radius: 8px;
-    transition: all .12s ease;
+    transition: all .2s ease-in-out;
```
CodeTheory: `.2s` seems too slow for me

onordbo: Did you test it @CodeTheory? It might seem slow, but it doesn't feel like it when you use it, tbh.

PR #289, opened 2017-11-03
Fix saving to an external disk

### LINKED ISSUE #55: Can't save Kaptures
via Twitter: https://twitter.com/RauschenImWalde/status/788488883637985292

> [...] I like your KAP app very much, but I cannot save anything (v2.0/3.0) on macOS Sierra

sindresorhus:
Fixes #55

`fs.renameSync` doesn't work across disks.

I've tested this with an external drive and it works.

@radum @gugahoi Can you confirm? Download: https://kap-artifacts.now.sh/save-across-device

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -33,6 +33,7 @@
     "make-dir": "^1.0.0",
     "menubar": "github:matheuss/menubar",
     "moment": "^2.14.1",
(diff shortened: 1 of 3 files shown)


radum:
@sindresorhus Just tested and works fine now. Thank you.

PR #290, opened 2017-11-06
Notify on plugin install/uninstall success and failure

sindresorhus:
screen shot 2017-11-07 at 01 48 05

I deliberately didn't add a notification on uninstall, as I don't see that being useful, and I also plan on making uninstalling a plugin instant.

---

You can try it out here: https://kap-artifacts.now.sh/plugin-notifications

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -1,7 +1,7 @@
 import path from 'path';
 import fs from 'fs';
 
-import {app, BrowserWindow, ipcMain, Menu, screen, globalShortcut, dialog} from 'electron';
+import {app, BrowserWindow, ipcMain, Menu, screen, globalShortcut, dialog, Notification} from 'electron';
 import isDev from 'electron-is-dev';
 
 import {init as initErrorReporter} from '../common/reporter';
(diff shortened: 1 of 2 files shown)


app/src/main/plugins.js
```diff
@@ -50,13 +50,7 @@ class Plugins {
...
     fs.writeFileSync(pkgPath, JSON.stringify(pkg));
 
-    try {
-      await this.upgrade();
-    } catch (err) {
-      console.error(err);
-      console.error(err.stderr);
-      console.error(err.stdout);
-    }
+    await this.upgrade();
```
albinekb: Will this never throw: Don't we want the silent logging here? 🤔

sindresorhus: @albinekb It can throw if `npm install` fails. Because it's in an async function it will be propagated and reported here: https://github.com/wulkano/kap/pull/290/files/5b6a4309b5e5ef2f3e85b345bf004c13a1454e37#diff-6f1bf0a59b91e2427edb5a476a341460R625

albinekb: I see, thanks! :+1:

app/src/main/index.js
```diff
@@ -601,14 +601,29 @@ const loadPlugins = async () => {
...
+
 ipcMain.on('install-plugin', async (event, name) => {
-  await plugins.install(name);
-  loadPlugins();
+  try {
+    await plugins.install(name);
+    notify(`Successfully installed plugin ${name}`);
+    loadPlugins();
+  } catch (err) {
+    dialog.showErrorBox(`Failed to install plugin ${name}`, err.stderr || err.stdout || err);
```
kevva: `err.message` instead of `err`?

sindresorhus: Good catch. I went with `err.stack`.

PR #292, opened 2017-11-07
Make plugin uninstall happen instantly

sindresorhus:
Previously it would take up to 10 seconds to uninstall a plugin as we just ran `npm uninstall plugin-name`. Now we're a little bit smarter.

Can be tested at: https://kap-artifacts.now.sh/instant-uninstall

### DIFF
--- a/app/src/main/plugins.js
+++ b/app/src/main/plugins.js
@@ -34,37 +34,51 @@ class Plugins {
     return Object.keys(JSON.parse(pkg).dependencies);
   }
 
-  async upgrade() {
-    await execa(process.execPath, [this.npmBin, 'install'], {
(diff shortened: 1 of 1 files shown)


app/src/main/plugins.js
```diff
@@ -34,37 +34,51 @@ class Plugins {
...
-      cwd: this.cwd,
-      env: {
-        ELECTRON_RUN_AS_NODE: 1
-      }
+    this._modifyMainPackageJson(pkg => {
+      delete pkg.dependencies[name];
     });
+
+    // Intentionally not awaited as it can just finish in the background
+    this._runNpm('prune');
```
kevva: Smart :+1:

Issue #219, opened 2017-08-30
Persist settings across relaunch: Aspect ratio options and Export format


jsejcksn:
**Platform:** macOS 10.12.6
**Kap Version:** 1.0.1.170
### Steps to reproduce
1. Open Kap.app
2. Expand Controls
3. Set Aspect ratio options and Export format
4. Quit Kap
5. Open Kap again

### Current behaviour
Settings are lost

### Expected behaviour
Settings persist

jsejcksn:
@sindresorhus If you want, I can help with UX testing for Kap in general. I have some other suggestions if you'd like to hear them.

sindresorhus:
@jsejcksn Definitely. Try out the latest master version: https://kap-artifacts.now.sh/master And let us know :)

CodeTheory:
For what it's worth, the latest `master` build has somewhat broken aspect ratio selection. I'll be fixing this along with #226 😄

jsejcksn:
@sindresorhus Wilco.

@CodeTheory Ah, great! Some of my suggestions are mirrored in your linked comment about aspect ratio behavior. Also, that's the version I've been using, but I downloaded it from the repo [release page](https://github.com/wulkano/kap/releases/tag/v1.0.1).

PR #303, opened 2017-12-08
Persist settings across relaunch: Aspect ratio options

karaggeorge:
This PR:
1. Adds this feature: https://github.com/wulkano/kap/issues/219
2. Solves the ratio issue (I don't know if it's documented or not, but ratios would show up as crazy long numbers)

Worked on each doing the following:
1. Use the `settings-manager` to save and load the values of the `dimensions` object which includes:
    - width
    - height
    - ratio
    - ratioLocked
2. Added a `Math.round` call before calling the `getSimplestRatio` function since `LargestCommonDivisor` as a mathematical concept only works with integers. The display values were rounded but the actual values in the `dimensions` object were not, which was causing the long decimals

### DIFF
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -18,7 +18,7 @@ const DEFAULTS = {
   dimensions: {
     height: 512,
     width: 512,
-    ratio: '1:1',
+    ratio: [1, 1],
     ratioLocked: false
   }
 };
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -14,7 +14,13 @@ const DEFAULTS = {
   highlightClicks: false,
   fps: 30,
   recordAudio: false,
-  audioInputDeviceId: 'none'
+  audioInputDeviceId: 'none',
+  dimensions: {
+    height: 512,
+    width: 512,
+    ratio: '1:1',
+    ratioLocked: false
+  }
 };
 
 const volatiles = {
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -192,29 +192,34 @@ document.addEventListener('DOMContentLoaded', () => {
     const lcd = getLargestCommonDivisor(width, height);
     const denominator = width / lcd;
     const numerator = height / lcd;
-    return `${denominator}:${numerator}`;
+    return [denominator, numerator];
   }
 
   function setSelectedRatio(width, height) {
-    dimensions.ratio = getSimplestRatio(Math.round(width), Math.round(height));
-    app.kap.settings.set('dimensions', dimensions);
-
     const ratios = document.querySelectorAll('.aspect-ratio-selector option');
     let hadMatch = false;
+
     for (const ratio of ratios) {
-      if (ratio.value === dimensions.ratio) {
-        aspectRatioSelector.value = dimensions.ratio;
+      const [first, second] = ratio.value.split(':');
+
+      if (width / first === height / second) {
+        aspectRatioSelector.value = ratio.value;
+        dimensions.ratio = [first, second];
         hadMatch = true;
         break;
       }
     }
 
     if (!hadMatch) {
+      dimensions.ratio = getSimplestRatio(Math.round(width), Math.round(height));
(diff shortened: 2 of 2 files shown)


app/src/renderer/js/main.js
```diff
@@ -40,23 +40,27 @@ document.addEventListener('DOMContentLoaded', () => {
...
 
   // Init dynamic elements
   if (app.kap.settings.get('recordAudio') === true) {
     toggleAudioRecordBtn.classList.add('is-active');
     micOnIcon.classList.remove('hidden');
     micOffIcon.classList.add('hidden');
   }
+    
+  // Set initial values
+  [inputWidth.value, inputHeight.value] = [width, height].map(Math.round);
```
albinekb: Haha clever 😍

sindresorhus: I think it's too clever. This is much more readable and only 4 characters more:

```js
inputWidth.value = Math.round(width);
inputHeight.value = Math.round(height);
```

app/src/common/settings-manager.js
```diff
@@ -14,7 +14,13 @@ const DEFAULTS = {
   highlightClicks: false,
   fps: 30,
   recordAudio: false,
-  audioInputDeviceId: 'none'
+  audioInputDeviceId: 'none',
+  dimensions: {
+    height: 512,
+    width: 512,
+    ratio: '1:1',
```
sindresorhus: I think we should persist it as plain numbers. I know it's already represented as a string in the code, but the code we can change later. It's much harder to change the settings.

So it should be saved as `ratio: [1, 1]`

karaggeorge: I can do that. Shouldn't be too big of a change. Do you want it as part of this PR?

app/src/renderer/js/main.js
karaggeorge: Do you want me to change it?

app/src/common/settings-manager.js
sindresorhus: Sure

app/src/renderer/js/main.js
sindresorhus: Yes, please :)

sindresorhus:
And some linter issues that need to be fixed: https://circleci.com/gh/wulkano/kap/454?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

app/src/renderer/js/main.js
karaggeorge: 0fbd884

app/src/common/settings-manager.js
karaggeorge: 600a667

karaggeorge:
- Fixed the linting issues: 0fbd884
- Save ratio as an array: 600a667

In order to save the ratio as an array, I had to modify the way setSelectedRatio works a bit, otherwise, you would see a different (re-calculated) ratio when you opened the app again

sindresorhus:
There's a lot of `app.kap.settings.set('dimensions', dimensions);` calls. I made [this module](https://github.com/sindresorhus/on-change) today that you could use to remove all the duplicate calls and put it in one place. Would you be willing to use it in your PR? Otherwise I'll do a follow-up commit, no worries.

```js
const dimensions = onChange(app.kap.settings.get('dimensions'), () => {
  app.kap.settings.set('dimensions', dimensions);
});

karaggeorge:
@sindresorhus That's a quite interesting package. I like it. However when adding it to this project, I got the following issue:

When this gets called `app.kap.settings.set('dimensions', dimensions);` I get the following error:
screen shot 2017-12-11 at 12 43 45

I think it has something to do with what this:`settings.setSync(key, value)` calls on the value object internally

Any thoughts?

sindresorhus:
@karaggeorge I need to dig into it later. Shouldn't block this change though.

karaggeorge:
@sindresorhus then it should be good to go once it's reviewed. I don't think there's anything else to add

matheuss:
Tested, very nice! Changed the tittle since I see nothing related to export format in the diff 🤔

PR #304, opened 2017-12-11
Debounce cropper view size changes

### LINKED ISSUE #137: Enhancement: Add delay after new user input on canvas dimensions
Currently when updating the canvas dimensions through the control panel, there's a minor delay occurring while the canvas is animating, after a new value has been added to an input field. Consider adding a minor delay _(maybe 100ms?)_ to the animation trigger after the `keyup` event or something in a  similar fashion, to prevent this from happening.

_An added benefit of implementing this would also be that the canvas doesn't look/feel completely spastic while updating the dimensions numerically. **See attached GIF for demo.**_

***

spastic-canvas mov

karaggeorge:
Fixes #137

The first approach was to create a debounce function to use in both cases, but it became complicated when `blur` needed to clear the timeout. The code was only a few lines and only repeated twice, so I figured it wasn't worth making it more complicated than it had to be.

`debounceTimeout` is set to 400ms and the change is triggered instantly on blur

## Gif:
kapture 2017-12-11 at 17 15 40


app/src/renderer/js/main.js
```diff
@@ -303,18 +306,28 @@ document.addEventListener('DOMContentLoaded', () => {
...
     this.value = this.value || (shake(this) && lastValidInputWidth); // Prevent the input from staying empty
+    handleWidthInput.apply(this, null, true);
   };
 
-  inputHeight.oninput = handleHeightInput;
-  inputHeight.onchange = handleHeightInput.bind(inputHeight, null, true);
+  let heightTimeout;
+  inputHeight.oninput = function () {
+    clearTimeout(heightTimeout);
+    heightTimeout = setTimeout(() => handleHeightInput.apply(this, arguments), debounceTimeout);
```
sindresorhus: Use spread operator instead of `arguments`.

sindresorhus:
The cropper no longer updates when I change the number in the input box.

sallar:
This PR will conflict when my PR #312 is merged.

sindresorhus:
I fixed the merge conflicts and the inline comment.

sindresorhus:
The behavior is a little bit quirky though. It does debounce, but when you start changing the width or height, it jumps 1 pixel before debouncing. Reproduce: Focus one of the input fields and watch the cropper while you hold the up arrow key down. I think it would be better to just use `lodash.debounce` here.

karaggeorge:
@sindresorhus I'll look into that weird behavior. Sorry, I was on vacation.

karaggeorge:
@sindresorhus the bug was caused by the small timeout that exists in the system when keeping a key pressed. The first character triggers, then there is a small timeout and then it starts registering the key presses in a row. So, the `400` timeout was too small for that. Changing it to `500` solved it for me, however, I believe that is a user setting, so depending on what users have it set to, it might show up again.

I did switch to using `_.debounce`. I didn't know lodash had that method 👍

sindresorhus:
Looks good :)

PR #307, opened 2017-12-18
Upgrade Sentry implementation

sallar:
This PR does the following:

+ Upgrades Raven dependency to the latest
+ Changes Sentry config to point to the new account


app/src/common/reporter.js
```diff
@@ -1,22 +1,31 @@
...
 import isDev from 'electron-is-dev';
 import unhandled from 'electron-unhandled';
 
-let ravenClient;
+let Raven;
 
 function init() {
   if (!isDev) {
-    const raven = require('raven');
+    Raven = require('raven');
```
albinekb: Hmm, should we use `raven-js` in the renderer, and `raven` in the main process? 🤔

app/package.json
```diff
@@ -39,7 +39,8 @@
     "ms": "^2.0.0",
     "npm": "^4.6.1",
     "object-path": "^0.11.3",
-    "raven": "^0.12.1",
+    "raven": "^2.3.0",
```
albinekb: now we depend on both these, do we need both `raven` and `raven-js`? Afaik one is for the renderer (browser side) and the other for the main (node) side?

sallar: Oh.. I was wondering about that. Let me double check (both worked). Checking now

albinekb: I **think*

albinekb: I **think** it's important to use both of them as they are tailored for their respective environments, like tracking clicked links in the DOM for the browser one while the node one is tracking more OS specific stuff.

sallar: @albinekb I agree. Done :)

sindresorhus:
Tested it and it didn't work. Apparently, the DSN needs to be different in the main process version.

sindresorhus:
With https://github.com/wulkano/kap/pull/307/commits/879c6aa1bc3af586e3955f7129d8d89220daef4b, it works perfectly for both main and renderer.

PR #318, opened 2018-01-09
Fix: Keep cropper window open when app is activated

sallar:
Added an option to prevent cropper window from being closed when an app comes into focus and cropper window's `blur` event fires.


sindresorhus:
Looks good :)

PR #319, opened 2018-01-09
Open cropper window on any change

sallar:
This PR changes the cropper opening behavior. When any kind of change to width, height or ratio is done, the cropper window is opened if not already there.

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -135,6 +135,7 @@ const openCropperWindow = (size = {}, position = {}, options = {}) => {
       cropperWindow.openDevTools({mode: 'detach'});
       cropperWindow.webContents.on('devtools-opened', () => {
         setCropperWindowOnBlur(options.closeOnBlur);
+        mainWindow.focus();
       });
     } else {
       setCropperWindowOnBlur(options.closeOnBlur);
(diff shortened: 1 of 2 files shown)


sindresorhus:
Found one minor issue: When I focus on of the input fields in the main window and press arrow up key to increase the size, the main window is blurred, which means I lose the ability to change the size with the arrow keys. Same thing if you try to change the number in the input box by deleting and writing a new number.

skllcrn:
Additional minor issue: Ideally when selecting a window to record in the the dropdown and then blurring that window or Kap, the dropdown should no longer display the selected app but "Custom". This is because we can't know that the app is still open and that the crop area is set correctly on it.

sallar:
@skllcrn fixed!
@sindresorhus I couldn't reproduce that. here's me doing all that:

sindresorhus:
@sallar Your GIF shows the cropper only opening after you click the record button, so you're on the wrong branch, I think.

sallar:
@sindresorhus Ah yeah now I got what you mean.

sallar:
@sindresorhus I think I cracked it. can you check it out?

sindresorhus:
Looks good :)

PR #320, opened 2018-01-10
Fix fullscreen coordinates

### LINKED ISSUE #310: Resolve feedback in #306
It was merged to unblock @sallar, but there were some unresolved things that need to be fixed: https://github.com/wulkano/kap/pull/306#issuecomment-355160210

sallar:
Fixes #310 

Output:

fullscreen

### DIFF
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -355,7 +355,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
     if (app.isFullscreen) {
       // Fullscreen
-      ipcRenderer.send('open-cropper-window', {width: app.width, height: app.height}, {x: 1, y: 1});
+      ipcRenderer.send('open-cropper-window', {width: app.width, height: app.height}, {x: 0, y: 0});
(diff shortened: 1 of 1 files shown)


app/src/renderer/js/main.js
```diff
@@ -100,6 +100,12 @@ document.addEventListener('DOMContentLoaded', () => {
     cropperBounds.width -= 2;
     cropperBounds.height -= 2;
 
+    // If we're recording fullscreen, set x, y to zero
+    if (dimensions.app && dimensions.app.isFullscreen) {
+      cropperBounds.x = 0;
+      cropperBounds.y = 0;
```
sindresorhus: The cropper buffer is on all sides, not just the bottom-left. So you need to adjust width and height too. Add 2 px.

sallar: @sindresorhus that's already done. I tested it the width and height are correct

sallar: ^ the above screen capture shows it

PR #321, opened 2018-01-10
Safely fail if there is no app icon

sallar:
If there is no app icon (and the command fails), just don't show any icon in app list.

### DIFF
--- a/app/src/renderer/js/size-selector.js
+++ b/app/src/renderer/js/size-selector.js
@@ -14,8 +14,7 @@ const RATIOS = [
 ];
 
 async function getWindowList() {
-  let windows = await getWindows();
-  windows = [...windows, {pid: 222, ownerName: 'Test'}];
+  const windows = await getWindows();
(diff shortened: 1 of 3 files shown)


app/src/renderer/js/size-selector.js
```diff
@@ -17,7 +17,8 @@ async function getWindowList() {
   const windows = await getWindows();
   const images = await getAppIconListByPid(windows.map(win => win.pid), {
     size: 16,
-    encoding: 'buffer'
+    encoding: 'buffer',
```
sindresorhus: The `encoding` option is gone, so we can remove this, I think.

app/src/renderer/js/size-selector.js
```diff
@@ -17,7 +17,8 @@ async function getWindowList() {
   const windows = await getWindows();
   const images = await getAppIconListByPid(windows.map(win => win.pid), {
     size: 16,
```
sindresorhus: Just noticed this, why is it 16 and not 32?

Issue #326, opened 2018-01-11
TypeError on saving .mp4 file


steveharman:
**Platform:** macOS 10.12.6
**Kap Version:** 2.x beta
### Steps to reproduce
Do a custom sized recording
Attempt to save to Desktop as .mp4
### Current behaviour

Unhandled Promise Rejection
TypeError: Error processing argument at index 1, conversion failure from undefined
    at ShareService.showError (/Applications/Kap Beta.app/Contents/Resources/app.asar/dist/main/share-service.js:92:31)
    at ShareService.run (/Applications/Kap Beta.app/Contents/Resources/app.asar/dist/main/share-service.js:134:12)
    at <anonymous>

### Expected behaviour
Save file without error
### Workaround
Return to the 1.x release

skllcrn:
We haven't been able to reproduce this issue @steveharman, could you please try installing the latest [`master`](https://kap-artifacts.now.sh/master) and repeating the actions that triggered it?

steveharman:
Recording seems fine now with today's update @skllcrn (2.0.0-beta 5)

Thanks

sonaro:
I was just able to reproduce this error on 2.0.0.681.  OSX 10.13.3 High Sierra.

**Error in plugin kap**
NonError: 1
    at module.exports.input (/Applications/Kap.app/Contents/Resources/app.asar/node_modules/ensure-error/index.js:14:10)
    at ShareService.showError (/Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:102:108)
    at /Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:151:17
    at Generator.throw (<anonymous>)
    at step (/Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:37:191)
    at /Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:37:402
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:135:7)

skllcrn:
@sonaro Looks like your issue is related to #377, does it happen consistently and only when exporting an MP4?

sonaro:
@skllcrn Ah yes, #377 is more fitting than this one.  Should I add my comment to #377 and close this one?

albinekb:
Yes please @sonaro 🤘

sonaro:
Please close.

sonaro:
@sklllcrn @albinekb Please close.

PR #328, opened 2018-01-11
Handle non-errors

sindresorhus:
Something throws a non-error and `.stack` doesn't exist.

Should at least show us what's wrong in #326.


sindresorhus:
@steveharman Can you try https://kap-artifacts.now.sh/handle-non-errors and report what error you get now? This time the actual error should at least be reported.

PR #332, opened 2018-01-14
Upgrade to CircleCI 2

sindresorhus:
Should result in faster builds.

https://circleci.com/docs/2.0/


sindresorhus:
Downloading artifacts still works: https://kap-artifacts.now.sh/circleci2

PR #335, opened 2018-01-17
Add cancel export functionality

### LINKED ISSUE #73: Cancel processing

sallar:
Fixes #73

cancel


app/src/main/convert.js
```diff
@@ -37,7 +37,16 @@ export default async function (exportOptions) {
...
   });
 
-  app.kap.mainWindow.send('export-progress', {text: ''});
+  ipcMain.on('cancel-export', () => {
+    convertProcess.cancel();
+  });
 
-  return outputPath;
+  return convertProcess
+    .then(() => {
```
sindresorhus: Use `await`

app/src/main/convert.js
```diff
@@ -37,7 +37,16 @@ export default async function (exportOptions) {
...
+      app.kap.mainWindow.send('export-progress', {text: ''});
+      return outputPath;
+    })
+    .catch(err => {
```
sallar: @sindresorhus how do you think I should handle the error? if this convert function errors (for example when process gets cancelled) the Kap plugin throws a giant error...

sindresorhus: You need to handle the `CancelError` here and ignore it, but rethrow if it's not a `CancelErro`: https://github.com/sindresorhus/p-cancelable#pcancelablecancelerror

sallar: @sindresorhus the problem is, the user of this `convert` function needs a `outputPath` and if there is none, it will throw

sindresorhus: Then you need to handle the cancelation higher up where they expect the `outputPath`, or handle the cancelation here, and handle not getting a `outputPath` where something expects it.

app/src/main/convert.js
sallar: @sindresorhus how to use await here with the cancel thing?

sallar: Ah I got it

app/src/main/share-service.js
```diff
@@ -109,6 +109,7 @@ export default class ShareService {
 
       if (context.canceled) {
         kap.mainWindow.send('hide-export-window');
+        kap.editorWindow.focus();
```
albinekb: 🏆

app/src/main/convert.js
```diff
@@ -37,7 +38,21 @@ export default async function (exportOptions) {
     }
   });
 
-  app.kap.mainWindow.send('export-progress', {text: ''});
+  ipcMain.on('cancel-export', () => {
+    convertProcess.cancel();
+  });
+
+  await convertProcess.catch(err => {
```
sindresorhus: Better to just use try/catch here instead of `.catch()`. Then you don't need line 52 either.

sindresorhus:
If I press "Cancel" during the "Analysing" process, I get this error:

screen shot 2018-01-19 at 19 44 13

sindresorhus:
The main window content shows for some milliseconds before showing the export content:

sindresorhus:
If I start exporting a GIF, press "Cancel", and then start exporting again, it will go back to the old progress for a couple of seconds.

sindresorhus:
Woot! \o/

PR #342, opened 2018-01-19
Blacklist redundant apps

### LINKED ISSUE #327: Blacklist redundant apps from the window selector
**Platform:** MacOS
**Kap Version:** 2.0.0-beta.4 (2.0.0-beta.4.556)
### Steps to reproduce
Currently, a vast amount of redundant apps show up in the window selector that are currently not recordable.

### Suggested fix
Create a blacklist of known redundant apps that shouldn't show up in the menu.

sallar:
Temporary workaround for #327

⚠️ The real issue is still in https://github.com/karaggeorge/mac-windows as it should not be returning the small windows or off screen ones like the menubar... but this is a temporary patch.

### DIFF
--- a/app/src/renderer/js/size-selector.js
+++ b/app/src/renderer/js/size-selector.js
@@ -13,6 +13,28 @@ const RATIOS = [
   '1:1'
 ];
 
+const BLACKLIST = [
+  'Kap',
+  'Kap Beta',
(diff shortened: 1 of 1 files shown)


sindresorhus:
This should not close #327. It's just a temporary workaround.

PR #343, opened 2018-01-19
Use the first audio source if not set by the user and remove `None` from the preferences

### LINKED ISSUE #276: Preferences for recording audio displaying "None" device
If you have not overwritten the system settings for audio recording in Kap, the selector should display "Default".

sallar:
Fixes #276

image

### DIFF
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -126,7 +126,7 @@ document.addEventListener('DOMContentLoaded', () => {
       displayId: String(display.id)
     };
 
-    if (recordAudio === true) {
+    if (recordAudio === true && audioInputDeviceId !== 'default') {
       apertureOpts.audioDeviceId = audioInputDeviceId;
     }
 
@@ -375,18 +375,6 @@ document.addEventListener('DOMContentLoaded', () => {
 
     app.kap.settings.set('recordAudio', isVisible(micOnIcon));
   };
-
-  observersToDispose.push(app.kap.settings.observe('recordAudio', event => {
-    const method = event.newValue ? 'add' : 'remove';
-    toggleAudioRecordBtn.classList[method]('is-active');
-    if (event.newValue === true) {
-      micOnIcon.classList.remove('hidden');
-      micOffIcon.classList.add('hidden');
-    } else {
-      micOnIcon.classList.add('hidden');
-      micOffIcon.classList.remove('hidden');
-    }
-  }));
 
   ipcRenderer.on('start-recording', () => startRecording());
 
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -129,7 +129,6 @@ document.addEventListener('DOMContentLoaded', () => {
     if (recordAudio === true) {
       if (audioInputDeviceId === 'default') {
         const [defaultAudioDevice] = await audioDevices();
-        console.log(defaultAudioDevice);
         apertureOpts.audioDeviceId = defaultAudioDevice.id;
       } else {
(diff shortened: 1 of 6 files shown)


sindresorhus:
There still needs to be a `None` option for people that don't want to record audio.

sindresorhus:
And maybe `Default` could show the first part of the default device name (as much as fits including `…`). `Default` => `Default (Built-in Microphone)`

sindresorhus:
@skllcrn ⏫ Thoughts?

sallar:
@sindresorhus this is how it looks like now:

image

can you please review again? :)

sallar:
@sindresorhus we've gotta think about using some other solution for dropdowns in the rewrite. native html dropdowns are not flexible. Can't show ... in the preview and full one when opened.

sindresorhus:
> we've gotta think about using some other solution for dropdowns in the rewrite. native html dropdowns are not flexible. Can't show ... in the preview and full one when opened.

Yeah, that's not a good UX.

sindresorhus:
screen shot 2018-01-20 at 22 04 43

There should only be one entry here. In short, if the user has never chosen an audio device, it should just be the first one.

I think we should drop the `Default` prefix and the shortening, seeing how it's impossible to show different content in the preview from the actual item.

sindresorhus:
So in my case, the only item in the list should be "Built-in Microphone" and it should be the default when I haven't selected one yet since it's the first in the list.

sallar:
@sindresorhus with the "default" option removed, there is a potential issue.

* If we set the `defaultAudioInputId` on app init to the first input, and user changes their default later, our selector will still show the old value. 
* If we don't set that and set it to `default` (in settings), the dropdown will not be the representation of the actual value stored, because it will change every time the user changes their setting from their system.

What do you think?

sindresorhus:
Let's just drop the concept of default in the UI. On startup, we'll take the first input if not set.


First launch:

```
| Built-in microphone |
| External input      |
```

User selects "External input.

Second launch:

```
| External input      |
| Built-in microphone |
```

We don't have to show the word "Default" anywhere. We'll just default to the first one internally if the user has not specifically set any device.

The user changing the default system input while Kap is launched sounds like an edge-case, but we could always just use the live value directly when recording starts if no preference is set.

sallar:
@sindresorhus Done. Now it works like you described.

app/src/renderer/js/main.js
```diff
@@ -9,8 +10,7 @@ import {handleKeyDown, validateNumericInput} from '../js/input-utils';
 import {handleTrafficLightsClicks, isVisible, disposeObservers} from '../js/utils';
 import buildSizeMenu, {findRatioForSize} from '../js/size-selector';
 
-const aperture = require('aperture')();
-
+const aperture = createAperture();
```
sindresorhus:
```js
const aperture = new Aperture();
```

app/src/renderer/js/main.js
```diff
@@ -1,5 +1,6 @@
 import {ipcRenderer, remote} from 'electron';
 import EventEmitter from 'events';
+import {default as createAperture, audioDevices} from 'aperture';
```
sindresorhus:
```js
import Aperture from 'aperture';
```

And then just use `Aperture.audioDevices`.

sallar: @sindresorhus default export is `createAperture` and `Aperture` is not exported. should I send a PR to that package?

sindresorhus: Oh, that's true. Duuh. I can fix the docs there. Never mind then.

sindresorhus:
Looks good :)

PR #345, opened 2018-01-23
Add default export buttons

timothyis:
This PR proposes and implements buttons for share services which would only show if there is only the `Save to file` option for the particular format the button is for. 

In theory, this decreases the effort for users to get to their desired output when there is no need for a select element with just one option.

image

In the image above, I have `kap-imgur` installed and so the GIF option is a select element due to multiple options for that format.

### DIFF
--- a/app/src/renderer/css/components/_buttons.css
+++ b/app/src/renderer/css/components/_buttons.css
@@ -167,6 +167,17 @@ button {
   }
 }
 
+.button--dark {
+  border-color: transparent;
+  color: #fff;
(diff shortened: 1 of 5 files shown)


sindresorhus:
The text in the new buttons are not correctly vertically centered. You can see that it doesn't line up with the GIF one.

timothyis:
You're right, good catch @sindresorhus. I'll fix it.

timothyis:
Fixed 👍 
Including a fix for buttons/selects not being disabled properly on export

sindresorhus:
👌😃✨

PR #348, opened 2018-01-23
Remove `cursor:pointer`

sindresorhus:
Native apps do not use the pointer cursor except for links. We should make sure Kap feels as native as possible.

I would also like to remove the grabbing hands from the editor trimmer, but I didn't include that here as I guessed it would be more controversial. Thoughts? Quicktime, for example, does not have a grabbing cursor on the trimming UI.

### DIFF
--- a/app/src/renderer/css/_helpers.css
+++ b/app/src/renderer/css/_helpers.css
@@ -8,7 +8,6 @@
 
 /* Allow element to be held down to drag the current window */
 .webkit-drag {
-  cursor: -webkit-grab;
   -webkit-app-region: drag;
 }
 
--- a/app/src/renderer/css/components/_buttons.css
+++ b/app/src/renderer/css/components/_buttons.css
@@ -7,11 +7,6 @@
   - Any extensions to input styling should go in this file
 */
 
-/* Ensure all buttons have the pointer cursor */
-.btn {
-  cursor: pointer;
-}
-
 /* Button */
 *[class^='button'] {
   height: 2.4rem;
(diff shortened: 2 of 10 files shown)


albinekb:
I don't agree with this. Why don't native apps use the correct cursor? Personally I don't like to click things that don't have the cursor pointer :'(

sallar:
I agree. cursor doesn't feel "native" on macOS. But that's just my feeling.

sindresorhus:
> Why don't native apps use the correct cursor?

I guess because native generally has consistent and obvious UI, so there's no need, while the web is a mess and there anything could be a button. The pointer was originally meant for links, but at some point, we started using it for everything...

timothyis:
Yeah, I'm not so sure how I feel about this. I think I've gotten used to the fact that kap has `pointer` set for a web like UX but I think removing them wouldn't be an issue and would maybe help with the integration into the desktop environment. 

👍  from me

sindresorhus:
> I would also like to remove the grabbing hands from the editor trimmer, but I didn't include that here as I guessed it would be more controversial. Thoughts?

@timothyis ⬆️

timothyis:
@sindresorhus Sorry, I meant a general thumbs up to this PR and that question. Should have been more specific.

timothyis:
@sindresorhus Just expanding on this:
We have interaction animations that we can do instead of cursors, which isn't necessarily unheard of in native apps.
For example, a button can change colour on hover and you'll know how you should interact with it (let alone the look of it)

sindresorhus:
@timothyis Yup

PR #350, opened 2018-01-23
Fixed Spinner svg

siamak:
– Fixed Spinner svg for `loading` toggle button.
– Fixed Animation rules for `loading`.


sindresorhus:
@siamak Look at the GIF here how it's supposed to look: https://github.com/wulkano/kap/pull/347

PR #353, opened 2018-01-27
Export window improvements

### LINKED ISSUE #340: Improve export window
app exporting

Now that #73 in we can move ahead on the design improvements for the export window. That includes displaying how many exports are cued in the title and an option to easily hide the window.

sallar:
Fixes #340

image

### DIFF
--- a/app/src/renderer/css/export.css
+++ b/app/src/renderer/css/export.css
@@ -1,0 +1,44 @@
+/*
+  Kap.app main styling
+  ----
+  Uses eightpoint (github.com/wulkano/eightpoint as a reset)
+  ( ͡° ͜ʖ ͡°)
+*/
+@import 'eightpoint/dist/eight.min';
+@import '_variables';
+@import '_helpers';
+@import '_base';
+@import 'components/_window-header';
+@import 'components/_buttons';
+@import 'components/_progress';
+
+body {
+  background-color: $background-primary;
+}
+
+section.progress-wrapper {
+  padding: 2rem;
+}
+
+section.progress {
+  /* @extend .webkit-drag; */
+  display: flex;
+  padding: 2rem;
+  margin-bottom: 15px;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  background-color: $gray-lightest;
+  border-radius: $window-border-radius;
+}
+
+.progress-bar-cancel-btn {
+  width: 100%;
+  color: $red;
+
+  &:hover {
+    background-color: $red;
+    border-color: $red;
+    color: $color-light;
+  }
+}
--- a/app/src/renderer/css/export.css
+++ b/app/src/renderer/css/export.css
@@ -24,12 +24,12 @@ section.progress {
   /* @extend .webkit-drag; */
   display: flex;
   padding: 2rem;
+  border-radius: $window-border-radius;
   margin-bottom: 15px;
   flex-direction: column;
   align-items: center;
   justify-content: center;
   background-color: $gray-lightest;
-  border-radius: $window-border-radius;
 }
 
 .progress-bar-cancel-btn {
@@ -37,8 +37,8 @@ section.progress {
   color: $red;
 
   &:hover {
-    background-color: $red;
     border-color: $red;
     color: $color-light;
+    background-color: $red;
   }
 }
--- a/app/src/renderer/css/export.css
+++ b/app/src/renderer/css/export.css
@@ -21,7 +21,6 @@ section.progress-wrapper {
 }
 
 section.progress {
-  /* @extend .webkit-drag; */
   display: flex;
   padding: 2rem;
   border-radius: $window-border-radius;
--- a/app/src/renderer/js/export.js
+++ b/app/src/renderer/js/export.js
@@ -50,6 +50,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
   ipcRenderer.on('end-export', () => {
     progressBar.value = 100;
+    progressCancelBtn.disabled = true;
     isExportInProgress = false;
   });
 });
--- a/app/src/renderer/js/export.js
+++ b/app/src/renderer/js/export.js
@@ -32,7 +32,6 @@ document.addEventListener('DOMContentLoaded', () => {
   ipcRenderer.on('end-export', () => {
     progressBar.value = 100;
     progressCancelBtn.disabled = true;
-    updateTitle('Success 🎉');
   });
 });
 
--- a/app/src/renderer/js/export.js
+++ b/app/src/renderer/js/export.js
@@ -1,55 +1,30 @@
 import {ipcRenderer, remote} from 'electron';
-import {handleTrafficLightsClicks} from '../js/utils';
 
 const {dialog} = remote;
-
-let isExportInProgress = false;
-
-window.onbeforeunload = e => {
-  if (isExportInProgress) {
-    const buttonIndex = dialog.showMessageBox(remote.getCurrentWindow(), {
-      type: 'question',
-      buttons: ['Cancel Export', 'Continue'],
-      defaultId: 1,
-      message: 'Are you sure you want to cancel exporting?',
-      detail: 'It will not be saved'
-    });
-    if (buttonIndex === 0) {
-      ipcRenderer.send('cancel-export');
-    } else {
-      // Prevent closing the window
-      e.returnValue = true;
-      e.preventDefault();
-    }
-  }
-};
 
 document.addEventListener('DOMContentLoaded', () => {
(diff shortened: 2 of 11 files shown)


app/src/renderer/js/export.js
```diff
@@ -0,0 +1,52 @@
...
+      type: 'question',
+      buttons: ['Cancel Export', 'Continue'],
+      defaultId: 1,
+      message: 'Are you sure you want to cancel exporting?',
+      detail: 'It will not be saved'
+    });
+    if (buttonIndex === 0) {
+      ipcRenderer.send('cancel-export');
+    } else {
+      e.returnValue = true; // Prevents closing
```
sallar: @sindresorhus HELP! This doesn't work. If this `e.returnValue = true` is outside of this if-else statement, it will prevent closing. But now, the window closes in both cases... do you have any idea whats happening here?

sallar: never mind. fixed.

sindresorhus:
Pressing "Continue" here hides the export window.

screen shot 2018-01-29 at 11 35 29

sindresorhus:
I don't think we should have a confirm dialog when clicking "Cancel Export". It's already clear to the user what happens when they press the button. We should still show the confirm dialog when the user tries to close the window, as it's not clear then what the user intended.

sindresorhus:
The export buttons in the editor window need to be disabled when an export is active, to prevent multiple exports.

sindresorhus:
Seems like the "Cancel Export" button is not working correctly. I tried starting an export, canceling, starting, canceling, and starting, and this is the result:



You can see that the progress bar goes back and forth as if it's multiple exports updating it.

sindresorhus:
It no longer shows the status text in the export window. Previously, it showed "Analyzing...", "Converting...", etc. This is important as an export can have multiple steps, especially with plugins, which can set their own progress state with [`.setProgress(text, percentage)`](https://github.com/wulkano/kap/blob/master/docs/plugins.md#action).

@skllcrn Where do you want to show this text?

app/src/renderer/views/export.pug
```diff
@@ -0,0 +1,15 @@
+extends _extend
+
+block head
+  title Exporting
+  +stylesheet('export')
+  +script('export')
+
+block body
+  +window-header(title='Exporting' updates).is-light.is-draggable
```
sindresorhus: Why are you not using the native traffic lights? We should use those whenever possible. The reason we don't use them in the editor windows is so that we can show them only on hover.

app/src/renderer/js/export.js
```diff
@@ -0,0 +1,55 @@
...
+
+  progressCancelBtn.onclick = () => {
+    window.close();
+  };
+
+  ipcRenderer.on('start-export', () => {
+    isExportInProgress = true;
+  });
+
+  ipcRenderer.on('export-progress', (e, data) => {
```
sindresorhus: `e` => `event`

app/src/renderer/js/export.js
```diff
@@ -0,0 +1,55 @@
...
+    if (data.percentage) {
+      progressBar.value = data.percentage * 100;
+    } else {
+      progressBar.value = 0;
+    }
```
sindresorhus:
```js
progressBar.value = percentage ? percentage * 100 : 0;
```

app/src/renderer/js/export.js
sindresorhus: `data` => `{percentage}`

app/src/renderer/js/export.js
```diff
@@ -0,0 +1,55 @@
+import {ipcRenderer, remote} from 'electron';
+import {handleTrafficLightsClicks} from '../js/utils';
+
+const {dialog} = remote;
+
+let isExportInProgress = false;
+
+window.onbeforeunload = e => {
```
sindresorhus: `e` => `event`

app/src/renderer/js/export.js
```diff
@@ -0,0 +1,55 @@
...
+  if (isExportInProgress) {
+    const buttonIndex = dialog.showMessageBox(remote.getCurrentWindow(), {
+      type: 'question',
+      buttons: ['Cancel Export', 'Continue'],
+      defaultId: 1,
```
sindresorhus: You should also specify `cancelId`.

app/src/renderer/js/export.js
```diff
@@ -0,0 +1,55 @@
...
+      message: 'Are you sure you want to cancel exporting?',
```
sindresorhus: I would write:

> Are you sure you want to cancel the export?

app/src/renderer/js/export.js
```diff
@@ -0,0 +1,55 @@
...
+      detail: 'It will not be saved'
```
sindresorhus: Needs a `.` at the end.

Exporting is not just about saving, a plugin might upload it instead of saving, so this needs a better wording. Or is it needed at all? I think it's pretty clear already in the `message` what's going to happen.

sallar: @sindresorhus I don't think this is needed. I copied this dialog from the editor window.

app/src/renderer/js/export.js
sallar: @sindresorhus Again, copied over the old code without thinking about it 🙄

app/src/renderer/views/export.pug
sallar: @sindresorhus My preference is using the default traffic lights and even the default title bar... But I thought we're doing it custom since the pug mixin seemed general purpose.

sallar:
@sindresorhus I agree with the native traffic lights and native toolbar, instead of css window header and css traffic lights.

skllcrn:
I don't see the white background and large button as a problem, and changing this one window to have more native elements makes it really inconsistent with the rest of the app. Would love some more reasoning here! It could definitively be nice to go back to a native titlebar, I'm sure that would remove a lot of complexity, but I believe one of the reasons we went with a custom one was for the behaviour when hiding it for detached and attached states. Regarding the different statuses (analysing, exporting etc.); it would be great to display those in the titlebar as @sallar suggested.

sallar:
@sindresorhus please review again. I could not find a way to have a taller native title bar, is there such a thing?

sindresorhus:
I'm still seeing https://github.com/wulkano/kap/pull/353#issuecomment-361139156

sindresorhus:
> please review again. I could not find a way to have a taller native title bar, is there such a thing?

No, there's no way to do that with Electron, currently.

sallar:
how do you reproduce this? which export format?

sindresorhus:
Record fullscreen for 10 seconds (To make a large video), the GIF export, wait 5 seconds, cancel, GIF export, then wait another 5 seconds, and GIF export. And you'll see the progress bar goes back and forth.

sallar:
@sindresorhus that sounds like an issue in "cancel" functionality but not this new window... should it be another bug or it happen with this PR?

sallar:
@sindresorhus actually I think I know what it is... give me a moment

sallar:
@sindresorhus can you try again?

sindresorhus:
The `.setProgress(text, percentage).` call from plugins are still not handled. It works it `master`. For example, test with `kap-giphy` and you'll see in master that it changes half way through to `Uploading...` with no progress. With your PR it just stops half way through and doesn't update until `Success..` at the end.

sallar:
@sindresorhus that's what I just fixed in the last commit

sindresorhus:
> that sounds like an issue in "cancel" functionality but not this new window... should it be another bug or it happen with this PR?

Can be fixed separately. Open a new issue.

app/src/renderer/css/export.css
```diff
@@ -0,0 +1,44 @@
...
+body {
+  background-color: $background-primary;
+}
+
+section.progress-wrapper {
+  padding: 2rem;
+}
+
+section.progress {
+  /* @extend .webkit-drag; */
```
sindresorhus: Leftover?

app/src/renderer/js/export.js
```diff
@@ -0,0 +1,38 @@
...
+    progressBar.value = percentage ? percentage * 100 : 0;
+    if (text) {
+      updateTitle(text);
+    }
+  });
+
+  ipcRenderer.on('end-export', () => {
+    progressBar.value = 100;
+    progressCancelBtn.disabled = true;
+    updateTitle('Success 🎉');
```
sindresorhus: I don't think we should show this anymore. We already show a real notification. That's enough. We can just close the progress window right away.

sallar:
@sindresorhus done

sindresorhus:
Great! Nice work @sallar

PR #355, opened 2018-01-29
Sort app list based on usage time and count

### LINKED ISSUE #352: Sort app recording list by last and most used
A note from working on #298: In the future we might want to sort apps you can record by last and most used.

sallar:
Fixes #352

### DIFF
--- a/app/src/renderer/js/size-selector.js
+++ b/app/src/renderer/js/size-selector.js
@@ -62,11 +62,12 @@ async function getWindowList() {
       .map(win => {
         const iconImage = images.find(img => img.pid === win.pid);
         const icon = iconImage.icon ? nativeImage.createFromBuffer(iconImage.icon) : null;
-        return Object.assign({}, win, {
+        return {
+          ...win,
           isFullscreen: false,
           icon2x: icon || null,
           icon: icon ? icon.resize({width: 16, height: 16}) : null
-        });
+        };
       })
   ];
 }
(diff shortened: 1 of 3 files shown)


app/src/renderer/js/size-selector.js
```diff
@@ -56,15 +62,49 @@ async function getWindowList() {
...
-        });
+        };
       })
   ];
 }
 
+function setAppLastUsed(app) {
+  const {count} = usageHistory[app.pid] || {};
+  usageHistory[app.pid] = {
+    count: (isNaN(count) ? 0 : count) + 1,
```
sindresorhus: `isNaN` => `Number.isNaN`

app/src/renderer/js/size-selector.js
```diff
@@ -56,15 +62,49 @@ async function getWindowList() {
...
+    lastUsed: new Date().getTime()
```
sindresorhus: `new Date().getTime()` => `Date.now()`

sallar: Im stupid

sallar:
@sindresorhus done

app/src/renderer/js/size-selector.js
```diff
@@ -56,15 +62,49 @@ async function getWindowList() {
...
-        });
+        };
       })
   ];
 }
 
+function setAppLastUsed(app) {
+  const {count} = usageHistory[app.pid] || {};
+  usageHistory[app.pid] = {
+    count: (Number.isNaN(count) ? 0 : count) + 1,
```
sindresorhus: When I asked you to change this from `isNaN` I didn't realize that the check was relying on coercion (Also the exact reason `Number.isNaN` exists). This check is now faulty as it will never be `true`. Instead use: `typeof count === 'number' ? ++count : 1`

PR #360, opened 2018-01-31
Handle plugin request cancelation

### LINKED ISSUE #357: Canceling export when exporting using a plugin causes error
I missed this when reviewing #353

```
TypeError: path must be a string or Buffer
    at Object.fs.open (fs.js:574:11)
    at Object.module.(anonymous function) [as open] (ELECTRON_ASAR.js:200:20)
    at ReadStream.open (fs.js:1883:6)
    at new ReadStream (fs.js:1870:10)
    at Object.fs.createReadStream (fs.js:1824:10)
    at action (/Users/Christopher/Library/Application Support/Kap Beta/plugins/node_modules/kap-giphy/index.js:13:25)
    at <anonymous>
```

```
TypeError: path must be a string or Buffer
    at fs.readFile (fs.js:297:11)
    at fs.readFile (ELECTRON_ASAR.js:456:25)
    at P (/Users/Christopher/Library/Application Support/Kap Beta/plugins/node_modules/pify/index.js:49:6)
    at Promise (<anonymous>)
    at /Users/Christopher/Library/Application Support/Kap Beta/plugins/node_modules/pify/index.js:11:9
    at ret (/Users/Christopher/Library/Application Support/Kap Beta/plugins/node_modules/pify/index.js:72:32)
    at action (/Users/Christopher/Library/Application Support/Kap Beta/plugins/node_modules/kap-clipboard/index.js:8:72)
    at <anonymous>

```

```
Unhandled Error
Error: Must specify either 'file' or 'files' option
    at EventEmitter.ipc.once (/Users/Christopher/Library/Application Support/Kap Beta/plugins/node_modules/kap-draggable/index.js:30:18)
```

sallar:
Fixes #357

Convert process should now throw an error instead of silently failing, because the plugins are also using that, so if it silently fails, they should handle the cancellation because they are calling `context.filePath()`. But if it throws an error, we can catch and check it's type.

### DIFF
--- a/app/src/main/convert.js
+++ b/app/src/main/convert.js
@@ -1,7 +1,6 @@
 import path from 'path';
 import {ipcMain} from 'electron';
 import tempy from 'tempy';
-import {CancelError} from 'p-cancelable';
 import {convertToGif, convertToMp4, convertToWebm, convertToApng} from '../scripts/convert';
 import {exportProgress} from './export';
 
@@ -43,14 +42,7 @@ export default async function (exportOptions) {
     convertProcess.cancel();
   });
 
-  try {
-    await convertProcess;
-  } catch (err) {
-    if (err instanceof CancelError) {
-      return;
-    }
-    throw err;
-  }
+  await convertProcess;
 
   return outputPath;
 }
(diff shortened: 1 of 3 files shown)


sindresorhus:
Good solution :)

PR #361, opened 2018-02-01
Allow only one plugin to be installing/uninstalling at the time

sindresorhus:
If you tried toggling multiple plugins at once, it would throw an error.

**DON'T MERGE**


sallar:
Oh shit im sorry. Im blind and didnt read

PR #366, opened 2018-02-01
Disable windows sub menu if there are no selectable windows

### LINKED ISSUE #363: Window menu item should be disabled when there are no windows
Currently, it just doesn't show a menu:

screen shot 2018-02-01 at 16 46 27

albinekb:
Fix #363

If I send the empty submenu array, the menu isn't greyed out 🙁 

image


sindresorhus:
Can you open an issue on Electron too? Would be a nice convenience if it was just handled there.

albinekb:
@sindresorhus https://github.com/electron/electron/issues/11813

PR #383, opened 2018-02-05
Check user environment for custom shortcut

### LINKED ISSUE #150: Add ability to
**Platform:**

OSX

**Kap Version:**
1

### Steps to reproduce

Command + Shift + 5

### Current behaviour

Shortcut is clashing with another shortcut with no way to change it.

### Expected behaviour

Preferences should have an option to set new shortcut

YoavGivati:
This will let users update their `~/.bash_profile` or whatever and add an environment variable `KAP_SHORTCUT` to customize the keyboard shortcut kap uses.

touches #150

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -304,7 +304,9 @@ app.on('ready', () => {
   // Ensure all plugins are up to date
   plugins.upgrade().catch(() => {});
 
-  globalShortcut.register('Cmd+Shift+5', () => {
+  const shortcut = process.env.KAP_SHORTCUT || 'Cmd+Shift+5';
+  
+  globalShortcut.register(shortcut, () => {
     const recording = (appState === 'recording');
     mainWindow.webContents.send((recording) ? 'stop-recording' : 'prepare-recording');
   });


sindresorhus:
It should be an option in the Preferences window, not like this.

skllcrn:
Until we ship custom shortcuts the idea was to add an option to disable them completely, additionally I've suggested creating a workaround to make use of this: https://support.apple.com/kb/PH25377

YoavGivati:
I only had 5 minutes to share a solution. I'm totally ok if this doesn't get merged in. 

But it seems to me that if no one has time to add a UI solution 
- making this work https://support.apple.com/kb/PH25377 would require some UI menu stuff
- adding an option to disable it is non ideal and a bigger fix, I can't imagine the person who has time to do this, but not just add the full keyboard shortcut UI.

... then letting users who really want this feature now get full customization via their local env is a solution that doesn't take anything away from anybody, and solves the problem for those that really want it right now.

albinekb:
@YoavGivati yes, if we merge this and release it now, we are stuck with this forever... Because if we remove it we will get 10000 issues about it not working. It's not fun.

YoavGivati:
> Because if we remove it we will get 10000 issues about it not working

Why would you ever remove it?

doot0:
@YoavGivati I would imagine because the intended solution integrates with the UI, not the CLI. If Kap ever gets a CLI interface then I'd be totally behind you but it's not the ideal solution, imho.

matheuss:
`~/.bash_profile` and friends are not run when applications are run on MacOS – they're only read by `bash` (and its respective files for its respective friends). Therefore, its very tricky to set such `KAP_SHORTCUT` to be a system-wide env var. As far as I remember, you'll need to edit a `.plist` somewhere.

With that in mind and the fact that we want to keep Kap's settings and interfaces, for now, strictly inside its own GUI, I'm closing the PR. Thank you for your time,  @YoavGivati 🙌

Issue #377, opened 2018-02-02
Getting error: `NonError: 1`


Overload119:
```
NonError: 1
     at module.exports.input (/Applications/Kap.app/Contents/Resources/app.asar/node_modules/ensure-error/index.js:14:10)
     at ShareService.showError (/Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:102:108)
     at /Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:151:17
     at Generator.throw (<anonymous>)
     at step (/Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:37:191)
     at /Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:37:402
     at <anonymous>
     at process._tickDomainCallback (internal/process/next_tick.js:135:7)
```

skllcrn:
We've been looking into this issue @Overload119, but have been struggling to reproduce it. We'll need some more details about your system and workflow; your OS version, the steps leading up to the error and anything else you think could be particular to your system. Does this by any chance happen when attempting to export a MP4 file? Does it occur consistently?

randoum:
I experience this issue too, fresh install first time user on macOs Sierra 10.12.4

Here is what I've done:
- Disable mouse cursor recording, disable analytics
- record a part of the screen by dragging the dashed square
- After recording is done, it indicates the size 798*1278 
- Select 30 fps
- click MP4

Boom:
```
Error in plugin kap

NonError: 1
    at module.exports.input (/Applications/Kap.app/Contents/Resources/app.asar/node_modules/ensure-error/index.js:14:10)
    at ShareService.showError (/Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:102:108)
    at /Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:151:17
    at Generator.throw (<anonymous>)
    at step (/Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:37:191)
    at /Applications/Kap.app/Contents/Resources/app.asar/dist/main/share-service.js:37:402
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:135:7)
```

**Note**: saving other formats (Gif, WebM, APNG) works without complaining

skllcrn:
Have you used Kap before @randoum? If you have, did you use any plugins?

randoum:
As stated:
[...]

PR #390, opened 2018-02-08
Show a better error if the error is not an error

matheuss:
Will help on https://github.com/wulkano/kap/issues/377


matheuss:
nvm, not helpful

PR #402, opened 2018-02-27
Fix `NonError: 1` and make MP4 obey `fps` and `dimensions`

### LINKED ISSUE #377: Getting error: `NonError: 1`

### LINKED ISSUE #400: MP4 export does not use `fps` nor `width` or `height`
When exporting MP4 we should use the set `fps`, `width` and `height` that we use when exporting other formats: https://github.com/wulkano/kap/blob/master/app/src/scripts/convert.js#L73

I think this was forgotten from how the old export used to work.

Steps to reproduce:
1. Record a kap in 1000x1000, 30 FPS
2. In the export window, set the size to 100x100 and FPS to 15
3. Export as MP4
5. The exported MP4 will be 1000x1000, 30 FPS (no changes)

karaggeorge:
Fixes #377
Fixes #400

The `NonError: 1` was caused by ffmpeg when trying to convert to mp4, when width or height were odd. Apparently, because of libx264, height and width have to be even (https://trac.ffmpeg.org/ticket/309). So getting the error was pretty random. Also, it was even worse, since when exporting mp4 wouldn't listen to the chosen dimensions as described in the other issue. So it was basically chances of your original recording having even dimensions. 

This PR passes the size and fps options to the mp4 converter and also makes sure the width and height are even to avoid getting that error.

Some Acceptance Testing would be nice! (I tested it a bunch already, but second pair of eyes doesn't hurt)

Also, the dimension being ignored problem was there for WebM format too, so I added the options for that converter as well

### DIFF
--- a/app/src/scripts/convert.js
+++ b/app/src/scripts/convert.js
@@ -71,10 +71,11 @@ const convertToGif = PCancelable.fn(async (onCancel, opts) => {
 });
 
 function convertToMp4(opts) {
-  // TODO: Instead of fixing the `file://` prefix here, just store it in a better place in the editor
   opts.progressCallback(0);
   return convert(opts.outputPath, opts, [
     '-i', opts.filePath,
+    '-r', opts.fps,
+    '-s', `${opts.width}x${opts.height}`,
     '-ss', opts.startTime,
     '-to', opts.endTime,
     opts.outputPath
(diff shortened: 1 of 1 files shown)


sindresorhus:
Yay! Thanks for fixing this @karaggeorge :)

PR #405, opened 2018-02-28
Add user preference to hide desktop icons during recording

### LINKED ISSUE #372: Add option to Preferences to hide desktop icons during recording
Many people have a mess of icons on their desktop, I included. Would be nice to have an option to hide them during recording.

There are two possible approaches:

1. Create a fullscreen window on the desktop-level and fill it with the user's wallpaper. See https://stackoverflow.com/a/11318100/64949 and the `BrowserWindow` `type` option.

2. https://www.indabaa.com/hide-desktop-icons-mac/

The first approach pretends to hide the icons by just putting an image above it, while the second approach is more invasive and actually hides them.

karaggeorge:
Fixes #372

Uses [hide-desktop-icons](https://github.com/karaggeorge/hide-desktop-icons) to create a window at the icon level with a picture that is the user's wallpaper, therefore hiding the icons.

Electron's BrowserWindow with the `desktop` type uses:
> The desktop type places the window at the desktop background window level (kCGDesktopWindowLevel - 1). Note that desktop window will not receive focus, keyboard or mouse events, but you can use globalShortcut to receive input sparingly.

But, we need `kCGDesktopIconWindowLevel`. Using the `desktop` option produces something like this (behind the icons):
screen shot 2018-02-27 at 15 31 32

I submitted a PR with Electron, to ask for a different type as a new feature, but until then I wrote a swift script to do it

Gifs
---
Excuse my mess of a desktop (reason why I need this 😆)

### Resulting Gifs
Without the option:

kapture 2018-02-28 at 13 10 36

With the option:
kapture 2018-02-28 at 13 14 56

### User experience
[Gif](https://www.dropbox.com/s/iqayfrwj3rgznsq/Kapture%202018-02-28%20at%2013.12.23.gif?dl=0) showing what actually happens when you record it (too large to upload)

### New Preferences Option (defaults to false)
screen shot 2018-02-28 at 13 34 38

### DIFF
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -2,6 +2,7 @@ import {ipcRenderer, remote} from 'electron';
 import EventEmitter from 'events';
 import {default as createAperture, audioDevices} from 'aperture';
 import _ from 'lodash';
+import desktopIcons from 'hide-desktop-icons';
 
 import {init as initErrorReporter, report as reportError} from '../../common/reporter';
 import {log} from '../../common/logger';
@@ -135,6 +136,8 @@ document.addEventListener('DOMContentLoaded', () => {
       }
     }
 
+    if (app.kap.settings.get('hideDesktopIcons')) await desktopIcons.hide();
+
     try {
       await aperture.startRecording(apertureOpts);
       ipcRenderer.send('did-start-recording');
@@ -157,6 +160,7 @@ document.addEventListener('DOMContentLoaded', () => {
     ipcRenderer.send('will-stop-recording');
 
     const filePath = await aperture.stopRecording();
+    if (app.kap.settings.get('hideDesktopIcons')) desktopIcons.show();
     ipcRenderer.send('stopped-recording');
     ipcRenderer.send('open-editor-window', {filePath});
     setMainWindowSize();
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -136,7 +136,9 @@ document.addEventListener('DOMContentLoaded', () => {
       }
     }
 
-    if (app.kap.settings.get('hideDesktopIcons')) await desktopIcons.hide();
+    if (app.kap.settings.get('hideDesktopIcons')) {
+      await desktopIcons.hide();
+    }
 
     try {
       await aperture.startRecording(apertureOpts);
@@ -160,7 +162,11 @@ document.addEventListener('DOMContentLoaded', () => {
     ipcRenderer.send('will-stop-recording');
 
     const filePath = await aperture.stopRecording();
-    if (app.kap.settings.get('hideDesktopIcons')) desktopIcons.show();
+
+    if (app.kap.settings.get('hideDesktopIcons')) {
+      desktopIcons.show();
+    }
+
     ipcRenderer.send('stopped-recording');
     ipcRenderer.send('open-editor-window', {filePath});
     setMainWindowSize();
--- a/app/package.json
+++ b/app/package.json
@@ -31,7 +31,7 @@
     "file-size": "^1.0.0",
     "first-run": "^1.2.0",
     "got": "^8.2.0",
-    "hide-desktop-icons": "^0.1.3",
+    "hide-desktop-icons": "^0.3.0",
     "insight": "^0.10.0",
     "jquery": "^3.3.1",
     "lodash": "^4.17.5",
--- a/app/package.json
+++ b/app/package.json
@@ -31,7 +31,7 @@
     "file-size": "^1.0.0",
     "first-run": "^1.2.0",
     "got": "^8.2.0",
-    "hide-desktop-icons": "^0.1.2",
+    "hide-desktop-icons": "^0.1.3",
     "insight": "^0.10.0",
     "jquery": "^3.3.1",
     "lodash": "^4.17.5",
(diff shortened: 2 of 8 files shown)


app/src/renderer/js/main.js
```diff
@@ -135,6 +136,10 @@ document.addEventListener('DOMContentLoaded', () => {
       }
     }
 
+    if (app.kap.settings.get('hideDesktopIcons')) {
+      await desktopIcons.hide();
+    }
+
```
albinekb: Maybe we should use something like http://npmjs.com/package/unload to show the icons if the app crashes during a recording (from here until it shows them again)? or will the child swift app die automatically? 🤔

albinekb: Yes it will? https://github.com/sindresorhus/execa#cleanup

karaggeorge: I think `child_process` dies if the parent dies (unless you use `detached`). So if the main process dies, I would expect the child one to die. When the script is killed, the swift window is closed.

I will test it tho and get back to you

sindresorhus:
Looks like you're not building the binary correctly. Remember, you need to explicitly include the Swift runtime: https://github.com/karaggeorge/mac-windows/blob/ece82fda7272af541a5dc91ed9390b1cb5fb68f1/package.json#L8

sindresorhus:
This is super cool @karaggeorge :)

sindresorhus:
What's the purpose of this https://github.com/karaggeorge/hide-desktop-icons/commit/3f14370c4fd81681149774d1d7a24b509dc4f0d7 ?

karaggeorge:
@sindresorhus:
- Not really that good with swift yet. What are the flags I need for the swiftc command to make it good for release?
- The desktop is different. [These](https://developer.apple.com/documentation/appkit/nsimagescaling) are the options for scaling the `NSImageView` and none of them will really fill it. They scale down to fit the parent. I figured it doesn't matter that much since you don't see the change in the recording. Any ideas for a different workaround?
- The purpose of that line is to make sure the icons are hidden before resolving the promise. I had a `print('ready')` in the swift script and tried listening for it with the `stdout` on the process but I had no luck. Also, that is not exactly the right place to put it. It should really be when `NSApp.run()` is ready, but I couldn't find an event for it. I tried it a bunch and a second seemed to always make sure that the icons have disappeared before the recording starts. Again any ideas would be great

sindresorhus:
Also, what is this for? https://github.com/karaggeorge/hide-desktop-icons/blob/df0b87f7d067804de90191240facc423dd5e5af6/swift/HideIcons.swift#L17 It needs a comment in the code or a descriptive constant name, if it has a purpose.

sindresorhus:
> Not really that good with swift yet. What are the flags I need for the swiftc command to make it good for release?

What you have in https://github.com/karaggeorge/mac-windows/blob/ece82fda7272af541a5dc91ed9390b1cb5fb68f1/package.json#L8 is good: `swift build -c release -Xswiftc -static-stdlib`

> The desktop is different. These are the options for scaling the NSImageView and none of them will really fill it. They scale down to fit the parent. I figured it doesn't matter that much since you don't see the change in the recording. Any ideas for a different workaround?

https://stackoverflow.com/a/46262418/64949 ?

> The purpose of that line is to make sure the icons are hidden before resolving the promise. I had a print('ready') in the swift script and tried listening for it with the stdout on the process but I had no luck

You mean you could not read the `ready` output or it was printed too early?

sindresorhus:
Here's a quick cleanup: https://github.com/karaggeorge/hide-desktop-icons/pull/1

karaggeorge:
@sindresorhus thank you for those fixes! I also added the fill effect using that stack overflow thread and fixed the swift build process. 

For the ready thing, when I ran the script by itself, I would see the 'ready' text printed. When I was running it with execa I would use something like this:
```
script = execa(script, args);
script.stdout.on('data', data => {
	data = data.toString()
});
```
But I wouldn't get anything. Also tried stderr, but didn't get a result either.

Even if that worked (with the print being right before `NSApp.run()`) it wouldn't be 100% accurate. The ready call should be when `NSApp.run()` finishes initializing, but there is no event I found that I could listen for.

karaggeorge:
Tried doing that with `didFinishLaunchingNotification` and `didBecomeActiveNotification` and trying to print something in the function, but no luck. I get nothing in the terminal.

The code I used:
```swift
final class NotifyWhenReady {
	init() {
		NotificationCenter.default.addObserver(self, selector: #selector(self.didFinishLaunching), name: NSApplication.didBecomeActiveNotification, object: nil)
	}


	@objc func didFinishLaunching() {
		print("READY")
	}
}
```

karaggeorge:
@sindresorhus this ended up doing the trick:

```swift
class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
  func applicationDidFinishLaunching(_ notification: Notification) {
    print("READY")
  }
}
```

and then passing that as the `NSApp.delegate`.

Also using `setbuf(__stdoutp, nil)` to prevent the stdout from buffering allowed me to get the data on the stdout stream and resolve the promise only when the icons are hidden

sindresorhus:
Yes, I was going to suggest that next. Glad you solved it.

karaggeorge:
Updated the dependency one more time. It should now be smooth and working well 👍

sindresorhus:
Works great! :)

Issue #223, opened 2017-09-04
Rewrite in React/Next.js


timothyis:
It would be much easier to maintain. Much easier for other to contribute.

https://leo.im/2017/electron-next

cc @leo ❤️

leo:
It's pretty easy. Check this out: https://github.com/leo/electron-next-skeleton

CodeTheory:
Yes! Super excited to get it done 🎉

skllcrn:
Moving #53 out of the current milestone/project, should be done in conjunction with the rewrite.

steveharman:
Just a note relating to #338 while it was in my head.  I get this a lot, trying to draw / position my marquee right where the KAP window is.  In 1.x I could just drag the window aside but not with 2.x.  

Seems a shame to have removed such a handy feature.

Thanks,

![kap_marquee](https://user-images.githubusercontent.com/12557929/36668776-4387e5a6-1aea-11e8-98ef-5a86ee102b52.png)

kevva:
Do we still want to keep the styles as separate CSS files or do we want to use [`styled-jsx`](https://github.com/zeit/styled-jsx)?

timothyis:
Since next.js comes equipped with styled-jsx and that I'm personally a fan of it, I'd vote for using styled-jsx!

skllcrn:
https://github.com/wulkano/kap/pull/450

skllcrn:
#494

PR #408, opened 2018-03-02
Ensure consistent code style across the project

kevva:
I was going to have a look at #223 but I thought having a clean code base from the beginning would make it easier.

Using arrow functions instead of regular ones also removes the hoisting we got going on.

### DIFF
--- a/app/src/renderer/js/editor.js
+++ b/app/src/renderer/js/editor.js
@@ -148,7 +148,7 @@ document.addEventListener('DOMContentLoaded', () => {
   };
 
   preview.addEventListener('canplay', event => {
-    aspectRatioBaseValues = [event.target.videoWidth, event.target.videoHeight];
+    aspectRatioBaseValues = [event.currentTarget.videoWidth, event.currentTarget.videoHeight];
     [inputWidth.value, inputHeight.value] = aspectRatioBaseValues;
     [lastValidInputWidth, lastValidInputHeight] = aspectRatioBaseValues;
 
@@ -173,14 +173,14 @@ document.addEventListener('DOMContentLoaded', () => {
   trimLine.addEventListener('mousemove', hover);
 
   maximizeBtn.addEventListener('click', event => {
-    event.target.classList.add('hidden');
+    event.currentTarget.classList.add('hidden');
     unmaximizeBtn.classList.remove('hidden');
     ipcRenderer.send('toggle-fullscreen-editor-window');
     $('body').classList.add('fullscreen');
   });
 
   unmaximizeBtn.addEventListener('click', event => {
-    event.target.classList.add('hidden');
+    event.currentTarget.classList.add('hidden');
     maximizeBtn.classList.remove('hidden');
     ipcRenderer.send('toggle-fullscreen-editor-window');
     $('body').classList.remove('fullscreen');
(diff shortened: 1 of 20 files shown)


app/src/renderer/js/editor.js
```diff
@@ -191,17 +244,17 @@ document.addEventListener('DOMContentLoaded', () => {
...
-  fps15Btn.addEventListener('click', function () {
+  fps15Btn.addEventListener('click', () => {
     this.classList.add('active');
     fpsMaxBtn.classList.remove('active');
     fps = 15;
   });
 
-  fpsMaxBtn.addEventListener('click', function () {
+  fpsMaxBtn.addEventListener('click', () => {
     this.classList.add('active');
```
sindresorhus: You can't use arrow function when `this` is used, like here.

kevva: Oops. Didn't catch those.

app/src/renderer/js/input-utils.js
```diff
@@ -29,18 +29,15 @@ const validateNumericInput = (input, opts) => {
   return value;
 };
 
-const handleKeyDown = event => {
+export const handleKeyDown = event => {
   const multiplier = event.shiftKey ? 10 : 1;
   const parsedValue = parseInt(this.value, 10);
```
sindresorhus: You forgot one here.

app/src/renderer/js/preferences.js
```diff
@@ -62,13 +60,14 @@ document.addEventListener('DOMContentLoaded', async () => {
   audioInputDeviceSelector.value = settingsValues.audioInputDeviceId;
 
   const tabs = $j('.prefs-nav > a');
-  tabs.on('click', function (event) {
+
+  tabs.on('click', event => {
     event.preventDefault();
     tabs.removeClass('is-active');
-    $j(this).addClass('is-active');
+    $j(event.currentTarget).addClass('is-active');
```
kevva: Don't know if there are more instances where you'd need `currentTarget` instead of `target`.

sindresorhus: Why not always use `currentTarget` to be sure? `currentTarget` is the bound element (The previous `this`), while `target` is what was clicked, which is usually the same, but if there are more element inside the target, it can be different.

kevva: Yeah, true. Let's do that.

app/src/renderer/js/main.js
```diff
@@ -220,104 +195,130 @@ document.addEventListener('DOMContentLoaded', () => {
...
     });
-  }
+  };
 
-  function handleWidthInput(event, validate) {
+  const handleWidthInput = (event, validate) => {
     // User is deleting the current value
     // to enter a new one
-    if (!this.value) {
+    if (!event.target.value) {
```
kevva: Had to keep `target` here because somewhere `currentTarget` is set to `null`. Probably because we are emitting this event manually in `handleKeyDown`.

sindresorhus:
I've been running this today and didn't notice any problems.

sindresorhus:
@kevva Thanks for the huge cleanup :)

kevva:
Sweet. Glad it didn't cause any unforeseen errors.

sindresorhus:
@kevva After this PR, the ratio selection stopped working:

screen shot 2018-03-07 at 21 56 16

PR #420, opened 2018-03-09
Add preference to toggle the record keyboard shortcut

sindresorhus:
Just a quick and dirty fix for now until we can properly add configurable keyboard shortcut in the rewrite.

Related to #150 

screen shot 2018-03-09 at 13 40 28

### DIFF
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -11,6 +11,7 @@ const DEFAULTS = {
   showCursor: true,
   highlightClicks: false,
   hideDesktopIcons: false,
+  recordKeyboardShortcut: true,
   fps: 30,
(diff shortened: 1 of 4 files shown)


sindresorhus:
// @@chriscoyier FYI

chriscoyier:
I'd love that! I'd definitely flip it off because of the CodePen conflict, but then I could leave the app open all the time I'd like I'd prefer. 

😍

karaggeorge:
Is there a way to unregister the key bind, so they wouldn't have to restart the app? But it's probably not worth it since its just a temporary solution

sindresorhus:
There is, but I didn't bother complicating it as it's just a temp solution.

PR #424, opened 2018-03-10
Add preference to active "Do Not Disturb" during recording

### LINKED ISSUE #422: Kap update notifications are pushed during recordings
**Platform:** MacOS Sierra (10.12.6)
**Kap Version:** 1.0.1 (1.0.1.170)
### Steps to reproduce
1. Wait for a Kap update to be available or maybe downgrade
2. Start recording something; wait a bit
3. A notification will appear reminding you to update kap - which, depending, will get recorded 😅 

### Expected behaviour
Kap shouldn't push update notifications if it's currently recording the screen.

### Workaround
Disable all notifications while recording.

However, the frustrating part is that I probably haven't booted Kap up in a while when I do a screen recording (I don't like to leave apps running idly by when I don't use them), so the situation arose where I had _just_ booted Kap up to do a quick recording, started recording, and then got the notification since it had been a while.

sindresorhus:
Fixes #422 

I made it true by default as I think Kap should come with good defaults.

screen shot 2018-03-11 at 02 08 23

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -15,6 +15,7 @@
   "main": "dist/main/index.js",
   "dependencies": {
     "@ffmpeg-installer/ffmpeg": "^1.0.13",
+    "@sindresorhus/do-not-disturb": "^0.1.0",
     "ajv": "^6.2.0",
     "aperture": "^4.0.0",
     "aspectratio": "^2.2.2",
--- a/app/package.json
+++ b/app/package.json
@@ -15,7 +15,7 @@
   "main": "dist/main/index.js",
   "dependencies": {
     "@ffmpeg-installer/ffmpeg": "^1.0.13",
-    "@sindresorhus/do-not-disturb": "^0.1.0",
+    "@sindresorhus/do-not-disturb": "^0.2.0",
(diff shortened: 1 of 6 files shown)


sindresorhus:
You can test it out at https://kap-artifacts.now.sh/do-not-disturb

// @Qix- @rauchg

Qix-:
@sindresorhus saw that you bumped it - if you make a new artifact I'll give it a test.

sindresorhus:
@Qix- https://kap-artifacts.now.sh/do-not-disturb

sindresorhus:
You need to disable the "Hide desktop background" feature, as it's broken.

Qix-:
Also getting application signature errors but I'll attribute that to the fact it's probably not an official build. However, didn't get them the first time I downloaded from the artifacts link.

sindresorhus:
> Also getting application signature errors but I'll attribute that to the fact it's probably not an official build. However, didn't get them the first time I downloaded from the artifacts link.

Yeah, I think only the first commit in a PR creates a signed build, for some weird reason.

PR #426, opened 2018-03-13
Update hide-desktop-icons dependency to make it work with ASAR

karaggeorge:
As per https://github.com/karaggeorge/hide-desktop-icons/issues/4

Fixed in: https://github.com/karaggeorge/hide-desktop-icons/commit/71c477e855406dc5310d1f18305054bb35b503c3

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -31,7 +31,7 @@
     "file-size": "^1.0.0",
     "first-run": "^1.2.0",
     "got": "^8.2.0",
-    "hide-desktop-icons": "^0.3.0",
+    "hide-desktop-icons": "^0.4.0",
     "insight": "^0.10.0",
     "jquery": "^3.3.1",
     "lodash": "^4.17.5",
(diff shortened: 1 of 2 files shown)


karaggeorge:
This https://github.com/karaggeorge/mac-windows/issues/1 was also implemented and published. Should I update that dependency as well in this PR?

albinekb:
I say go ahead @karaggeorge 👍

karaggeorge:
@sindresorhus has mac-windows been working fine in the packaged versions of kap? I didn't use the ASAR path in that either. Should I update that one too?

sindresorhus:
> has mac-windows been working fine in the packaged versions of kap? I didn't use the ASAR path in that either. Should I update that one too?

It's working fine since you're using `childProcess.execFile`, which is patched by Electron. Electron just left out `spawn`, which `execa` uses, which is why we have to do this dumb workaround. See: https://github.com/electron/electron/issues/9459

PR #434, opened 2018-03-21
Fix `Do Not Disturb` persistence issue

### LINKED ISSUE #433: On Mac, Notification Center "Do Not Disturb" setting becomes permanent
I updated to v2.1.0. I noticed the "DND" option was checked by default, so I unchecked it immediately. I recorded something. 

A few minutes later, I notice I'm not getting any notifications. I see that Notification Center is disabled, "until midnight". I enable it. I restart my Mac. Notification Center is once again disabled.

I open `~/Library/Preferences/ByHost/com.apple.notificationcenterui.<guid>.plist`. I see that `dndStart` is set to `0` (indicating 0:00 on a 24-hour clock, I assume) and `dndEnd` is set to `1440` (indicating 24:00). This means that I won't be getting notifications forever. I manually change both values to `0` (or alternatively, you can delete them), and I restart my Mac. Notification Center is enabled. The [post here](https://discussions.apple.com/thread/7604851) was useful in fixing this.

This started with this update.

karaggeorge:
Fixes #433

### Changes:
- Don't turn DND on if the user already has it on
- Default DND setting to off
- Add 'Experimental' flag

### DIFF
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -11,7 +11,7 @@ const DEFAULTS = {
   showCursor: true,
   highlightClicks: false,
   hideDesktopIcons: false,
-  doNotDisturb: true,
+  doNotDisturb: false,
   record60fps: false,
   recordKeyboardShortcut: true,
   recordAudio: false,
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -139,7 +139,10 @@ document.addEventListener('DOMContentLoaded', () => {
 
     if (app.kap.settings.get('doNotDisturb')) {
       wasDoNotDisturbAlreadyEnabled = await doNotDisturb.isEnabled();
-      doNotDisturb.enable();
+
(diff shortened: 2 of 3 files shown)


karaggeorge:
Also opened a PR for do-not-disturb package to try a less evasive way of enabling DND, so even if for some reason we don't clean it up, a user can disable it normally in the Notification Center

https://github.com/sindresorhus/do-not-disturb/pull/4

sindresorhus:
> Don't turn DND on if the user already has it on

Should we maybe do this in the Swift package too? I mean, make `enable()`/`disable()` moot if already in that mode.

karaggeorge:
@sindresorhus I think that's a good idea: https://github.com/sindresorhus/do-not-disturb/pull/4/commits/f555da09d2935020555ad0664dda7ca83828e1e3

sindresorhus:
Thanks for fixing, @karaggeorge 🙌

PR #435, opened 2018-03-21
Include the application name in update notification

jesstelford:
Lessens reliance on recognising the iconography to identify where the notification originated from.


albinekb:
Thanks! ❤️

Issue #436, opened 2018-03-23
Confusing prompt copy when cancelling an export


jonsadka:
**Platform:**
`17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64`
**Kap Version:**
`Version 2.1.1 (2.1.1.747)`
### Steps to reproduce
- Make a recording, 
- Export as .gif
- Click on `Cancel Export` button
- Read copy on the confirm cancellation modal 
### Current behaviour
Confirm modal says `Are you sure you want to cancel the export?` and the two options of `Continue` and `Cancel Export`. It is unclear if `Continue` means I want to continue with cancelling the export or continue with the export.
<img width="443" alt="screen shot 2018-03-23 at 9 00 51 am" src="https://user-images.githubusercontent.com/6504944/37840007-33fb5d6e-2e79-11e8-8caa-5a50b8d01508.png">
### Expected behaviour
Confirm modal says `Are you sure you want to cancel the export?` and the two options should say something like `Yes` and `No` or `Do not Cancel` and `Cancel`.

skllcrn:
Ah, yes. We should change this to:

`Are you sure you want to cancel the export? `
`Cancel cancelling` or `Cancel`

On a more serious note; `Yes` or `No` is simple and obvious, but I'm not sure we need this prompt at all? You don't lose the original recording if you cancel the export.

Jamedjo:
> but I'm not sure we need this prompt at all? 

@skllcrn Maybe it helps if the user clicks `[x]` to close the dialog during a long export instead of minimizing it? If explicitly clicking `Cancel Export` I don't think it is needed.

skllcrn:
I think it's useful to have an explicit option to cancel, this is in part to future proof for queued exports. The button would allow you to cancel the currently active one and skip to the next in the queue, while closing the window would cancel all of them.

Jamedjo:
> I think it's useful to have an explicit option to cancel

@skllcrn Agreed. Perhaps I should have worded my suggestion better:

> @skllcrn Maybe *the confirm modal* helps when the user *accidentally* clicks [x] to close the *export dialog* during a long export instead of minimizing it? If explicitly clicking Cancel Export I don't think *the confirm modal* is needed.

I like your suggestion to get rid of the confirm modal for the cancel button, but am think it might be worth keeping the confirm modal when cancelling the export through the title bar.

[...]

PR #438, opened 2018-03-25
Improve copy on cancel export dialog

Jamedjo:
## What

Changes button text on cancel export modal

screen shot 2018-03-25 at 19 03 04

## Why

Closes https://github.com/wulkano/kap/issues/436

### DIFF
--- a/app/src/main/export.js
+++ b/app/src/main/export.js
@@ -22,8 +22,8 @@ export const startExport = () => {
     if (isExportInProgress) {
       const buttonIndex = dialog.showMessageBox(exportWindow, {
         type: 'question',
-        buttons: ['Cancel Export', 'Continue'],
-        defaultId: 1,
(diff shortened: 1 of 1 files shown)


jonsadka:
👏🏽

app/src/main/export.js
```diff
@@ -22,7 +22,7 @@ export const startExport = () => {
     if (isExportInProgress) {
       const buttonIndex = dialog.showMessageBox(exportWindow, {
         type: 'question',
-        buttons: ['Cancel Export', 'Continue'],
+        buttons: ['Cancel Export', 'Continue Export'],
         defaultId: 1,
```
albinekb: I think `cancel` should be the default, so you just can press enter

Jamedjo: Nice, updated!

albinekb:
Thanks!

Issue #429, opened 2018-03-18
Provide option to save source file (to avoid export for long videos being extremely slow)


Jamedjo:
### Steps to reproduce

Record long 30min video and attempt WebM export

### Problem

I was exporting a 30min recording and had to eventually abort the `WebM` export after a couple of hours. Using the inspector I checked the progress bar percentage which appeared to be stuck at 15%.

Attempting `mp4` export was slightly better: I could see progress being made but still slowly.

### Workaround

While looking around to find an exiting issue I noticed that `aperture.js` stores temporary files in `/var/folders/` and was able to find the video and copy it to my `Kaptures` folder instantly. Perhaps because this was on APFS?

### Feature request

1. Could you provide an instant export option which just copies the recorded video to the target directory?
2. ~~Could you provide more of a hint of expected export duration, or potentially a warning to avoid WebM export on long captures.~~ See: https://github.com/wulkano/kap/issues/437

### Platform

**Mac OS:** `10.13.3 (17D47)`
**Kernel:** `Darwin 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64`
**Kap Version:** `2.0.0 (2.0.0.681)`

albinekb:
>While looking around to find an exiting issue I noticed that aperture.js stores temporary files in /var/folders/

This is the source file, it's not possible to export this file directly from Kap, when you choose "MP4" export, we still convert the file trough ffmpeg - even if there are no settings changed.

skllcrn:
What FPS are you capturing at @Jamedjo? You can check your capture frame rate in preferences.

matheuss:
`1.` is easy and it'd be very nice, indeed, to have it. `2.` is very tricky and I'm not sure it's possible to implement it 😕

Jamedjo:
> What FPS are you capturing at @Jamedjo? 

I'm not entirely sure, but the preference toggle is set to off. Presumable that means it is using the default of `30 FPS` rather than the bolded `60 FPS`?

<img width="490" alt="screen shot 2018-03-25 at 13 33 18" src="https://user-images.githubusercontent.com/1028741/37875004-254a01a8-3031-11e8-9453-cd40b02d33ae.png">

Jamedjo:
> 2. is very tricky and I'm not sure it's possible to implement it 😕

[...]

PR #439, opened 2018-03-26
WIP: Add "Save Original" button

Jamedjo:
## What

Adds `Save: Original` button.

## Why

Much faster export of long videos. Takes a couple of seconds to export gigabytes instead of hours.

## Why not

- The extra button may be avoidable
- This was partly an exercise for me to get used to an electron app, node.js and back into javascript
- This doesn't currently return the original file, but instead uses `ffmpeg` without FPS or resolution and adds in `-c copy` to get a much faster conversion. I'll probably close this and do two things instead:
  1. Ensure that export `mp4` omits FPS and dimensions when they are unchanged and uses `-c copy` instead.
  2. Depending on design in https://github.com/wulkano/kap/issues/429, add a save option somewhere which copies the file instead of using ffmpeg.

## Screenshot

screen shot 2018-03-26 at 01 01 45

## Related

Closes https://github.com/wulkano/kap/issues/429

### DIFF
--- a/app/src/main/convert.js
+++ b/app/src/main/convert.js
@@ -1,7 +1,7 @@
 import path from 'path';
 import {ipcMain} from 'electron';
 import tempy from 'tempy';
-import {convertToGif, convertToMp4, convertToWebm, convertToApng} from '../scripts/convert';
+import {convertToGif, convertToMp4, convertToWebm, convertToApng, trimOriginal} from '../scripts/convert';
 import {exportProgress} from './export';
 
 // `exportOptions` => format filePath width height fps loop, defaultFileName
@@ -13,6 +13,8 @@ export default async exportOptions => {
     convert = convertToGif;
   } else if (format === 'mp4') {
     convert = convertToMp4;
+  } else if (format === 'original') {
+    convert = trimOriginal;
   } else if (format === 'webm') {
     convert = convertToWebm;
   } else if (format === 'apng') {
--- a/app/src/main/plugins.js
+++ b/app/src/main/plugins.js
@@ -152,10 +152,19 @@ class Plugins {
       ['apng', 'APNG'],
       ['gif', 'GIF'],
       ['mp4', 'MP4'],
+      ['original', 'Original'],
       ['webm', 'WebM']
     ]);
 
     return formats.get(format);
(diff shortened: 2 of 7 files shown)


sindresorhus:
> This doesn't currently return the original file, but instead uses ffmpeg without FPS or resolution and adds in -c copy to get a much faster conversion.

It should just copy the original file without going through `ffmpeg`.

sindresorhus:
I agree it should be possible to get the original file. I'm not yet decided on how it should be presented.

karaggeorge:
I think instead of making it complicated by adding to the export methods, you could just trigger the save dialogue when the button is clicked, and when the user selects a path/name just copy the file to that location

karaggeorge:
Hey @Jamedjo,

can you try adding an option under the File menu, with a label "Save Original…"?

When that is clicked, the save dialogue can open up, and after clicking save the file could just be copied to that destination.

It doesn't really fir with the export buttons since it doesn't really need any converting and it won't reflect the settings that are displayed on the left side (width/height/fps). So it makes sense to separate it from them.

Jamedjo:
> Can you try adding an option under the File menu, with a label "Save Original…"?

@karaggeorge Sure, I suggested this in https://github.com/wulkano/kap/issues/429#issuecomment-375996946 too. The only danger is if users end up expecting the export options there once `save` is added. I would add the export options too, but plugins complicate that so I'll just add save for now.

As mentioned in the description, I'll probably close this pull request in favour of one that does that.

Jamedjo:
> I think instead of making it complicated by adding to the export methods, you could just trigger the save dialogue when the button is clicked

@karaggeorge One downside of this is that it wouldn't be possible for plugins to use the output. E.g. If I want to quickly share something without waiting for export. However I don't think this is important, and a second pull request to make Mp4 export faster will help with that use case.

skllcrn:
Excellent, thank you @Jamedjo!

sindresorhus:
> can you try adding an option under the File menu, with a label "Save Original…"?

@Jamedjo Would you be able to update this PR per the above ^ ?

albinekb:
We'll implement this in Kap 3.0 👍

PR #443, opened 2018-03-31
Use npm registry

karaggeorge:
Use npm registry when installing plugins

### DIFF
--- a/app/src/main/plugins.js
+++ b/app/src/main/plugins.js
@@ -55,7 +55,7 @@ class Plugins {
   }
 
   async upgrade() {
-    await this._runNpm('install');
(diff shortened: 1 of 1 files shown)


app/src/main/plugins.js
```diff
@@ -55,7 +55,7 @@ class Plugins {
   }
 
   async upgrade() {
-    await this._runNpm('install');
+    await this._runNpm('install --registry=https://registry.npmjs.org');
```
sindresorhus: Note the signature of `_runNpm`.

karaggeorge: Oops, sorry. I'll fix it when I get to a computer

karaggeorge: 3e1fbde

PR #446, opened 2018-04-06
Fix fps undefined error

### LINKED ISSUE #444: FPS button shows `undefined`
**Platform: MacOS High Sierra 10.13.3**
**Kap Version: 2.1.1 (2.1.1.747)**

Refer to the picture.
screen shot 2018-04-05 at 2 23 50 pm

karaggeorge:
Fixes #444

When we migrated from the `fps` slider to a `record60fps` toggle, we didn't change the way we read the setting in the editor. Everyone that had the app before the change, had the fps still set in their local settings, so the broken behavior wasn't obvious, but getting a new installation of kap would show undefined.

### DIFF
--- a/app/src/renderer/js/editor.js
+++ b/app/src/renderer/js/editor.js
@@ -33,8 +33,7 @@ document.addEventListener('DOMContentLoaded', () => {
   const trimmerOut = $('#trimmer-out');
   const trimLine = $('.timeline-markers');
 
-  let maxFps = app.kap.settings.get('fps');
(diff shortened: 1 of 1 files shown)


sindresorhus:
@karaggeorge When you do `Fixes ...` you have to use the issue number, not URL. Otherwise, GitHub doesn't auto-close the referenced issue. They used to support just using the URL, but they no longer do.

karaggeorge:
Good to know!

PR #447, opened 2018-04-06
Fix dimension swapping

### LINKED ISSUE #445: Can't swap dimensions while aspect ratio is locked
**Kap Version:** 2.1.1.747
### Steps to reproduce
- create a cropper window
- make sure the aspect ratio is locked (and not 1:1)
- try to swap the dimensions
### Current behaviour
- sometimes makes the  cropper smaller
### Expected behaviour
- swaps the dimensions and reverses the ratio
### Workaround
- unlock aspect ratio, swap, lock them again

karaggeorge:
Fixes #445

Ignores the ratio lock when swapping dimensions 

Also:
- Sets a min width and height of 100px (same for both)
   - Can't enter width or height less than 100
   - Can't resize (using the mouse) to smaller than 100x100
- When ratio is locked and changing one dimension, check if the other dimension (calculated based on the ratio) is invalid (too high or too low) and if it is, invalidate the change

#### To explain the last point:
If I have a ratio of 2:1, and it's locked, even though the min value for width is 100px, entering 100px would make the height 50px (based on the ratio), which is less than the height's minimum of 100px. So the real minimum of the width when locked to the 2:1 ratio is 200px (which would lead to a 100px height).

### DIFF
--- a/app/src/renderer/js/main.js
+++ b/app/src/renderer/js/main.js
@@ -269,7 +269,7 @@ document.addEventListener('DOMContentLoaded', () => {
     dimensions.width = parseInt(event.target.value, 10);
     app.kap.settings.set('dimensions', dimensions);
 
-    if (dimensions.ratioLocked) {
+    if (dimensions.ratioLocked && !event.detail.ignoreRatioLocked) {
       dimensions.height = Math.round((second / first) * event.target.value);
       app.kap.settings.set('dimensions', dimensions);
       inputHeight.value = dimensions.height;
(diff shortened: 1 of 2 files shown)


app/src/renderer/js/main.js
```diff
@@ -341,8 +341,8 @@ document.addEventListener('DOMContentLoaded', () => {
   swapBtn.addEventListener('click', () => {
     [inputWidth.value, inputHeight.value] = [inputHeight.value, inputWidth.value];
     dimensions.ratio = dimensions.ratio.reverse();
-    inputWidth.dispatchEvent(new Event('input'));
-    inputHeight.dispatchEvent(new Event('input'));
+    inputWidth.dispatchEvent(new CustomEvent('input', {detail: {ignoreRatioLocked: true}}));
```
albinekb: Do we need to wrap it in `detail: {}`? is that standard practise? Otherwise just do `new CustomEvent('input', {ignoreRatioLocked: true})` to reduce complexity

karaggeorge: I wish, but it has to be under a `detail` key

https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events

karaggeorge:
Don't merge this yet. There is an issue with it that I'm trying to fix, and I'm also fixing another aspect ratio bug that was discovered while working on it.

sindresorhus:
🙌

PR #459, opened 2018-05-09
Fix tray window opening on the wrong screen

### LINKED ISSUE #365: Kap tray window stays on the wrong screen
Platform: macOS 10.12.6 (16G1114)
Kap Version: 2.0.0.681

This is probably related to #53

### Steps to reproduce

This requires an external display.

1. Click on the Tray icon, display Kap window.
2. Leave Kap window open,
3. Click on the tray icon on another display,

### Current behaviour

Kap window is displayed on the wrong display (on the one, that it was left visible).

### Expected behaviour

Kap window should be displayed at the same display where the tray icon was clicked.

### Workaround

Don't leave the window visible when changing screens ;-)

nnnoel:
Fixes #365

I'm still familiarizing myself with the codebase, so I'm not entirely clear on the ramifications of this fix.

Tested using the binary and the expected behavior seems fairly consistent, although occasionally I noticed the menubar events will sometimes fall a little out of sync.

All in all, I really enjoy using this app, so I'm def looking forward to future API improvements.

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -373,8 +373,8 @@ menubar.on('after-create-window', () => {
     recomputeExpectedWindowPosition();
     recomputeCurrentWindowPosition();
     if (expectedWindowPosition.x !== currentWindowPosition.x || expectedWindowPosition.y !== currentWindowPosition.y) { // This line is too long
-      menubar.setOption('x', currentWindowPosition.x);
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks, @nnnoel :)

nnnoel:
np! glad I could help :)

PR #450, opened 2018-04-18
Next.js rewrite and new cropper

karaggeorge:
Branch to work on Kap 3.0. Contains Next.js rewrite and the new cropper.

Technologies used:
- Next.js
- No gulp/babel for the main process to keep build time miinimum, still able to use es syntax on the renderer since Next.js builds it automatically
- Unstated for state management on the renderer

To download and test the current version: http://kap-artifacts.now.sh/next

To develop, download the branch and
- `yarn start` to start in development
- `yarn run pack` to get a Kap.app
- `yarn run dist` to get a Kap.dmg
- `yarn test` to lint

### Current Preview:
kapture 2018-04-18 at 16 30 17

### Features so far (when the branch was created):
- Cropper
	- Click+Drag to draw a cropper
	- Resize/Move
	- Locking aspect ratio
	- Shift + Click + Drag to resize keeping the aspect ratio
	- Windows list
	- Fullscreen
	- Switch between main and advanced action bar

### Current Backlog:
- [x] Introduce `xo` (0379749) (#46)
- [x] Don't allow the window to go off the screen while resizing with aspect ratio locked (7aedd88)
- [x] Implement the selecting a window functionality
	- [x] Main window functionality (1aefe5d
	- [x] Sorting based on last usage functionality (b9e1e9f)
- [x] Implement the width/height inputs
	- [x] Basic functionality (572758a)
	- [x] Edge cases (572758a)
	- [x] ArrowUp/ArrowDown functionality (820ed4f)
- [x] Settings (persist across relaunch) (572758a)
- [x] Implement the aspect ratio select functionality (adff3b7)
- [x] Migrate Settings to `electron-store` (057658c)
- [x] Style the action bar (06d6ff9)
- [x] Implement the Preferences window
	- [x] Settings tabs (26b2ffa)
	- [x] Plugin tab (a32cb80)
	- [ ] Simplify FPS options (#418)
- [x] Update to Electron 2.0
- [ ] Use `electron-updater` (#258) 
	- [ ] Research https://electronjs.org/docs/tutorial/updates
- [ ] Implement the Editor (#450)
	- [x] Open Kap's editor with an external video file (#224)
- [x] Introduce `aperture` and implement recording functionality
- [ ] Implement the Export window (#263, #354 #464)
- [ ] Analytics
	- [ ] Disable Sentry (`raven.js`) when toggling analytics off in settings (#471)
	- [ ] Introduce `insight` (#89)
- [ ] Plugins
[...]

PR #461, opened 2018-05-11
Center cropperWindow inside the focused display #131

### LINKED ISSUE #131: Kap always shows marquee on 'system' display.
**Platform:**
Darwin Steves-iMac.local 16.3.0 Darwin Kernel Version 16.3.0: Thu Nov 17 20:23:58 PST 2016; root:xnu-3789.31.2~1/RELEASE_X86_64 x86_64

**Kap Version:**
1.0.0 (1.0.0.134)

### Steps to reproduce
With multiple screens attached to a desktop Mac, open Kap from a Finder window on the non-System* display.  (* where the menu bar is dragged in the Mac's Display system preference)

Click on the red 'select recording area' tool in Kap.

### Current behaviour
The marquee always appears on the Mac's *system* display, not the display I launched Kap on.

### Expected behaviour
Should show the marquee on the screen I launched Kap on, where I want to do my recording.

### Workaround
Drag marquee from the System display over to one of the other displays,.

catalinmiron:
Closes #131

### DIFF
--- a/app/src/main/index.js
+++ b/app/src/main/index.js
@@ -116,22 +116,28 @@ const setCropperWindowOnBlur = (closeOnBlur = true) => {
 };
 
 const getCropperXYCoordinates = ({position = {}, width, height}) => {
-  let {x, y} = settings.get('cropperWindow.position');
-  let bounds = screen.getDisplayNearestPoint(screen.getCursorScreenPoint()).bounds;
+  const {x, y} = settings.get('cropperWindow.position');
+  const bounds = screen.getDisplayNearestPoint(screen.getCursorScreenPoint()).bounds;
 
   if (position.x === undefined && position.y === undefined) {
     if (x === 'center' && x === 'center') {
-      let centeredX = Math.ceil(bounds.x + ((bounds.width - width + cropperWindowBuffer) / 2));
-      let centeredY = Math.ceil(bounds.y + ((bounds.height - height + cropperWindowBuffer) / 2));
-
-      return {x: centeredX, y: centeredY};
-    } else {
-      return {x: x - (cropperWindowBuffer / 2), y: y - (cropperWindowBuffer / 2)};
+      return {
+        x: Math.ceil(bounds.x + ((bounds.width - width + cropperWindowBuffer) / 2)),
+        y: Math.ceil(bounds.y + ((bounds.height - height + cropperWindowBuffer) / 2))
+      };
     }
+
+    return {
+      x: x - (cropperWindowBuffer / 2),
+      y: y - (cropperWindowBuffer / 2)
+    };
   }
 
-  return {x: position.x - (cropperWindowBuffer / 2), y: position.y - (cropperWindowBuffer / 2)};
-}
+  return {
+    x: position.x - (cropperWindowBuffer / 2),
+    y: position.y - (cropperWindowBuffer / 2)
+  };
+};
 
 const openCropperWindow = (size = {}, position = {}, options = {}) => {
   options = Object.assign({}, {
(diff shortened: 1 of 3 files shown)


catalinmiron:
@karaggeorge thanks for reviewing the code. It doesn't happen for me actually. I have 3 displays (primary display, my macbook, and two other displays) and if I don't move the cropper window (storing the position) when I'm opening the cropper again in a different display it will center in that particular display. But after moving the cropper and have stored the coordinates (x, y) then, after opening the cropper again, it will open in the previous location.

Please check the uploaded video:
https://www.youtube.com/watch?v=3gbvZYtCHgY

Thanks!

catalinmiron:
If you're talking about a different os (other than macosx) I think that we can recalculate the coordinates with `getDisplayNearestPoint()` for `getCursorScreenPoint` like `screen.getDisplayNearestPoint(screen.getCursorScreenPoint()).bounds` and then recalculate the `x, y` based on that position. 

```js
const getCropperXYCoordinates = ({position, width, height}) => {
  let {x, y} = settings.get('cropperWindow.position');

  if (position.x && position.y) {
    return {...position};
  }

  // check if x and y has a predefined position, otherwise calculate the center of the focused window.
  // 'center' is the default value for x and y and so we know that there so no predefined value for them.
  if (x === 'center' && y === 'center') {
    let bounds = screen.getDisplayNearestPoint(screen.getCursorScreenPoint()).bounds;
    let centeredX = Math.ceil(bounds.x + ((bounds.width - width + cropperWindowBuffer) / 2));
    let centeredY = Math.ceil(bounds.y + ((bounds.height - height + cropperWindowBuffer) / 2));

    return {x: centeredX, y: centeredY};
  }

  return {x, y};
}

// in openCropperWindow

let {x, y} = getCropperXYCoordinates({position, width, height});

new BrowserWindow({
  // ...
  x,
  y
})
```

catalinmiron:
@karaggeorge I have to push some more commit in order to get everything up and running. I'm also thinking that we are hardcoding the `fullscreen` coordinates (x,y) values to `{x: 0, y: 0}` which is not true when using multiple displays. Please stop reviewing the code, I'll ping you once this is ready for review. Thanks!

karaggeorge:
@catalinmiron I'll take another look, I might have missed something. As for the fullscreen, if you are referring to us setting x and y to 0 right before recording (for aperture), that is the desired behavior, since we need them to be relative to that display in that case.

catalinmiron:
@karaggeorge what do you think about this approach for fullscreen? :)

image

skllcrn:
That's not a bad idea @catalinmiron, but I'd be hard pressed to know which is which based on "Secondary display 1" and "Secondary display 2", would probably need a preview. The simplest solution is to go fullscreen on the active display, the one Kap was triggered on. That's what we're doing in the `[next](http://kap-artifacts.now.sh/next)` branch too: https://github.com/wulkano/kap/pull/450

catalinmiron:
@karaggeorge everything should be fixed now. Including fullscreen and changing the ratio. For changing the ratio, I'm using `setBounds` with animation and reuse the `getCropperXYCoordinates`. Please take another look and let me know if this code look better. 

cc: @sindresorhus if you also have some time to take a look.

Thanks!

karaggeorge:
@catalinmiron I just checked out the latest commit of the PR and tried it locally but the cropper shows up in the wrong display. Please take a look at the video [here](https://www.dropbox.com/s/6hvr7llherzl88j/VID_20180516_104950.mp4?dl=0).

I think that happens because the `{x, y}` coming from the settings has the location of the last opened cropper. So, in the logic you have, it will follow the first `if` but not the second. So, it will open it in the last position.

I think you had it almost right with your initial solution. Try checking if there is an x and y in the `position` object. If there is, we use those relative to the current display. If not, try getting the x, y, width, height of the last cropper from the settings. Use those along with `screen.getDisplayMatching(rect)` to find what screen that cropper belonged to. Then find the currently active screen using `screen.getDisplayNearestPoint(screen.getCursorScreenPoint())` and compare their ids. If they are the same, use that x, y, width, height. If not, pass in `center: true`.

skllcrn:
https://github.com/wulkano/kap/pull/466

catalinmiron:
@karaggeorge @skllcrn would you like to continue working on this? I saw that there's a PR conflict. Thanks!

skllcrn:
I suggest you and @nnnoel agree on which PR is best to continue off and look into the feedback from @karaggeorge (https://github.com/wulkano/kap/pull/461#issuecomment-389551292). Then you can help each other wrap it up and review!

PR #463, opened 2018-05-15
Enhance Google Analytics

### LINKED ISSUE #89: Enhance Google Analytics

alisonmonteiro:
It aims to fix #89 

Kap is tracking `Recording ended` already. I used this as an example 😉 

#### Tracked events

- [x]   Opened Preferences
- [x]   Toggled Settings
- [x]   Opened Plugin Configuration
- [x]   Recording started
- [x]   Recording ended
- [x]   Accessed Editor/Preview
- [x]   Saved file
- [x]   Discarded file
- [x]   Exported
- [x]   Checked for updates

### DIFF
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -77,8 +77,8 @@ export const set = (key, value, {volatile = false} = {}) => {
     return objectPath.set(volatiles, key, value);
   }
 
-  if (key !== 'dimensions') {
-    track(`settings/${key}/toggled`);
+  if (key !== 'dimensions' && typeof value === 'boolean') {
+    track(`settings/${key}/toggled/${value}`);
   }
 
   settings.setSync(key, value);
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -3,6 +3,7 @@ import {app} from 'electron';
 import settings from 'electron-settings';
 import objectPath from 'object-path';
 import aperture from 'aperture';
+import {track} from '../main/analytics';
 
 const DEFAULTS = {
   kapturesDir: `${homedir()}/Movies/Kaptures`,
@@ -74,6 +75,10 @@ export const getAll = () => {
 export const set = (key, value, {volatile = false} = {}) => {
   if (volatile) {
     return objectPath.set(volatiles, key, value);
+  }
+
+  if (key !== 'dimensions') {
+    track(`settings/${key}/toggled`);
   }
 
   settings.setSync(key, value);
--- a/app/src/main/analytics.js
+++ b/app/src/main/analytics.js
@@ -6,7 +6,7 @@ import {get as getSetting} from '../common/settings-manager';
 
 const trackingCode = 'UA-84705099-2';
 const insight = new Insight({trackingCode, pkg});
-const version = parse(pkg.version, {loose: true});
+const version = parse(pkg.version);
 
 export const init = () => {
   if (firstRun()) {
--- a/app/src/main/analytics.js
+++ b/app/src/main/analytics.js
@@ -1,10 +1,12 @@
 import firstRun from 'first-run';
 import Insight from 'insight';
+import {parse} from 'semver';
 import pkg from '../../package';
 import {get as getSetting} from '../common/settings-manager';
 
 const trackingCode = 'UA-84705099-2';
 const insight = new Insight({trackingCode, pkg});
+const version = parse(pkg.version, {loose: true});
 
 export const init = () => {
   if (firstRun()) {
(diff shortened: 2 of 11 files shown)


app/src/common/settings-manager.js
```diff
@@ -76,6 +77,10 @@ export const set = (key, value, {volatile = false} = {}) => {
     return objectPath.set(volatiles, key, value);
   }
 
+  if (key !== 'dimensions') {
+    track(`settings/${key}/toggled`);
```
sindresorhus: Toggled is only useful if we know whether it was enabled or disabled.

app/src/main/index.js
```diff
@@ -312,10 +314,14 @@ const openPrefsWindow = () => {
 
   prefsWindow.on('close', () => {
     prefsWindow = undefined;
+    track('preferences/closed');
```
sindresorhus: I don't see how this is useful. We already know it was opened. Of course the user is going to close it too. Unless I'm missing something.

app/src/main/menus.js
```diff
@@ -11,6 +12,8 @@ const checkForUpdatesItem = {
         title: 'No updates available!',
         body: 'You will automatically receive updates as soon as they are available 🤗'
       })).show();
+
+      track('updates/checked');
```
sindresorhus: This should rather be more specific for reporting when users click the "Check for Updates" button. Reporting when an update is checked for is not very useful.

sindresorhus:
Some things that could be useful to know:

- What plugins users install so we can see which one is the most popular.
- What plugins users use in the export menu.
- What FPS users set.
- What file formats users use.

sindresorhus:
> Recording started
  Recording ended

What we want to know is how many recordings users are doing and how long. Just events for start and stop are not that useful.

skllcrn:
This is really helpful and a great first contribution, thank you @alisonmonteiro!

alisonmonteiro:
@skllcrn @sindresorhus thanks for the feedback! I'll send an update soon. 🙌

alisonmonteiro:
> What we want to know is how many recordings users are doing and how long.

I think that [`15835ee`](https://github.com/wulkano/kap/pull/463/commits/15835eea065842c5439559a3de477d5894947f98) solves this.

> What plugins users install so we can see which one is the most popular.
What plugins users use in the export menu.
What FPS users set.
What file formats users use.

I can do it. It will take more time for me. But I can do. 😅

app/src/main/index.js
```diff
@@ -513,6 +513,9 @@ ipcMain.on('will-start-recording', () => {
     cropperWindow.setAlwaysOnTop(true);
   }
 
+  timeStartedRecording = new Date().getTime();
```
sindresorhus: `Date.now()` ;)

skllcrn:
Would it be possible to know which version of Kap the events are coming from, preferably without losing interoperability with previous events?

alisonmonteiro:
> Would it be possible to know which version of Kap the events are coming from

By adding a prefix?

`v2.1.2/preferences/opened`
`v2.1.2/file/discarded`

or simply

`v2/preferences/opened`
`v2/file/discarded`

🤔

skllcrn:
Definitively prefer the incremental semver if that's possible!

sindresorhus:
@skllcrn I don't think we should differentiate patch versions though. Then we get too many disparate events. We could do `v2.1`, `v2.2`, etc.

alisonmonteiro:
> We could do v2.1, v2.2, etc.

Yes, cool

We could use [`semver-truncate`](https://github.com/sindresorhus/semver-truncate) and

```js
const truncate = semverTruncate('1.3.5-foo', 'minor');
//=> '1.3.0'
const version = `v${truncate.slice(0, -2)}`
//=> 'v1.3'
```

And I think we don't even need the `.slice` here.

🤔

alisonmonteiro:
If there is anything more I can do, so we can get this merged, please let me know. 🙌

app/src/main/analytics.js
```diff
@@ -1,18 +1,20 @@
 import firstRun from 'first-run';
 import Insight from 'insight';
+import {parse} from 'semver';
 import pkg from '../../package';
 import {get as getSetting} from '../common/settings-manager';
 
 const trackingCode = 'UA-84705099-2';
 const insight = new Insight({trackingCode, pkg});
+const version = parse(pkg.version, {loose: true});
```
sindresorhus: No need to use the `loose` option. We follow semver.

alisonmonteiro: Oh, I see. Thanks again for the feedback. 🙌

sindresorhus:
Excellent. Thank you, @alisonmonteiro :)

PR #466, opened 2018-05-16
Fix active cropper placement between displays

### LINKED ISSUE #131: Kap always shows marquee on 'system' display.

nnnoel:
Fixes #131 for the most part

- cropper displays in the same window the tray icon was clicked
- works while tray menu's open
- works after moving and resizing cropper
- tested with binary

However, after moving cropper around in one window followed by re-opening it in another, the initial position becomes a little off, which probably has something to do with what's going on [here](https://github.com/wulkano/kap/blob/621f1cabc6ce859923a6c07536241d781454726d/app/src/main/index.js#L189-L217)

Also open to suggestions

### DIFF
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -24,7 +24,7 @@ const DEFAULTS = {
   }
 };
 
-const DEFAULT_VOLATILES = {
+const getDefaultVolatiles = () => ({
   cropperWindow: {
     size: {
       width: 512,
@@ -35,11 +35,11 @@ const DEFAULT_VOLATILES = {
       y: 'center'
     }
   }
-};
+});
 
 export const DEFAULT_CROPPER_WINDOW_POSITION = 'center';
 
-const volatiles = JSON.parse(JSON.stringify(DEFAULT_VOLATILES));
+const volatiles = getDefaultVolatiles();
 
 // We need to sync every setting that can be modified externally
 // e.g. the `openOnStartup` setting can be modified via
@@ -86,6 +86,6 @@ export const set = (key, value, {volatile = false} = {}) => {
 export const observe = (keyPath, handler) => settings.observe(keyPath, handler);
 
 export const reset = async keyPath => {
-  await settings.setSync(keyPath, objectPath.get(DEFAULT_VOLATILES, keyPath));
+  await settings.setSync(keyPath, objectPath.get(getDefaultVolatiles(), keyPath));
   objectPath.set(volatiles, keyPath, settings.getSync(keyPath));
 };
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -83,3 +83,8 @@ export const set = (key, value, {volatile = false} = {}) => {
 };
 
 export const observe = (keyPath, handler) => settings.observe(keyPath, handler);
+
+export const reset = async (keyPath) => {
+  await settings.setSync(keyPath, objectPath.get(DEFAULT_VOLATILES, keyPath));
+  objectPath.set(volatiles, keyPath, settings.getSync(keyPath));
+};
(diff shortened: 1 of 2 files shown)


app/src/common/settings-manager.js
```diff
@@ -37,6 +37,10 @@ const volatiles = {
   }
 };
 
+export const DEFAULT_CROPPER_WINDOW_POSITION = 'center';
```
albinekb: Poor mans deep-clone? 😂

app/src/common/settings-manager.js
```diff
@@ -37,6 +37,10 @@ const volatiles = {
...
+
+const volatiles = JSON.parse(JSON.stringify(DEFAULT_VOLATILES));
```
albinekb: Poor mans deep-clone? 😂

sindresorhus: A better option is just making `DEFAULT_VOLATILES` a function that returns the object. Then there's no need for cloning as you get a fresh object each time.

nnnoel: good point!

timothyis:
Hey @nnnoel! I just tested this with my external display. It seems, if I open the cropper on my external display, it appears between both screens 😰 
image

nnnoel:
Interesting..

I tested this using a HP 27es and a HP 2311, and so far haven't been able to reproduce this. I wonder if having more vertical real estate influences the positioning behavior.

@timothyis Is the behavior intermittent?

timothyis:
@nnnoel It's continuous from the external display. If I open the cropper from the external display, I get this behavior. If I open it from my main display, it works fine.

Tested on a Dell U2715H 27"

nnnoel:
If you were to throw a log between these two lines https://github.com/nnnoel/kap/blob/a2e499398e04e3aa554f5701c570d85ae19f8f26/app/src/main/index.js#L427-L428 -- do you get anything after opening the cropper between displays?

nnnoel:
#### External HP 27es
screen shot 2018-05-19 at 12 23 54 pm 2
#### After resizing and reopening (and switching focus from built-in to external again)
screen shot 2018-05-19 at 12 34 18 pm 2
#### Switching back over to built-in again and reopening
screen shot 2018-05-19 at 12 37 22 pm


:confused:

skllcrn:
In case you missed it @nnnoel https://github.com/wulkano/kap/pull/461#issuecomment-390404815

nnnoel:
@skllcrn I didn't. I can do what @karaggeorge is recommending for the other PR. The problem is I can't reproduce the issue, so even replacing the cropper/screen detection logic, I wouldn't be able to confirm without feedback from someone else.

@catalinmiron Sounds like your fix could make use of something along the lines of what I'm doing here https://github.com/wulkano/kap/pull/466/commits/b921fa361f6f68fe079b3a7eb1af610d5800e3f4 -- which aims to revert any volatile keyPath to it's initial value whenever needed.

skllcrn:
If you're able to implement @karaggeorge's suggestions I can help test, maybe @catalinmiron would be able to help with that too!

PR #492, opened 2018-08-01
Disable allowAnalytics by default

mikaelf:
In order to conform with data protection regulation (the principle of Privacy by Default in the GDPR) I'm disabling analytics.

### DIFF
--- a/app/src/common/settings-manager.js
+++ b/app/src/common/settings-manager.js
@@ -8,7 +8,7 @@ import {track} from '../main/analytics';
 const DEFAULTS = {
   kapturesDir: `${homedir()}/Movies/Kaptures`,
   openOnStartup: false,
-  allowAnalytics: true,
+  allowAnalytics: false,
   showCursor: true,
   highlightClicks: false,
   hideDesktopIcons: false,


skllcrn:
I'm not opposed to this change, however we will lose *a lot* of valuable insight that we use to improve Kap. We don't collect any PII and the data is not used for any sort of identification or tracking, GDPR conditions for consent (Art. 7) explicitly governs *personal* data, defined as: 

‘personal data’ means any information relating to an identified or identifiable natural person (‘data subject’); an identifiable natural person is one who can be identified, directly or indirectly, in particular by reference to an identifier such as a name, an identification number, location data, an online identifier or to one or more factors specific to the physical, physiological, genetic, mental, economic, cultural or social identity of that natural person; (Art. 4) 

You can read more about what kind of information we use and our implementation of analytics in [our privacy policy](https://github.com/wulkano/kap/blob/master/PRIVACY.md). We take a lot of care to only collect the information we actually need and to retain that information for the shortest possible time.

mikaelf:
I assumed that the data could identify a person (device identifiers, IP address, etc.), in which case I would prefer opt-in, but it's great to hear that you don't process any personal data. 

Perhaps the description below the "Allow analytics" option should link to the privacy policy?

skllcrn:
Good idea! We'll make sure to add a link to the privacy policy in the next update. For the record we do not collect IP addresses or any kind of device identifiers.

PR #499, opened 2018-09-14
Refactor settings

karaggeorge:
### Tasks
- Merges the two settings tabs
- Update settings wording
- Change recording fps control to a select
- Add looping option and connect to export

### Preview
kapture 2018-09-13 at 18 08 46

Exporting with looping off:
kapture 2018-09-13 at 18 10 51

Exporting with looping on:
kapture 2018-09-13 at 18 10 20

### Questions
- Wasn't sure if looping should be enabled by default (for now I enabled it)
- Wasn't sure if we are adding the Kaptures dir (wasn't in the mockup) and what are we doing about the path thing

screen shot 2018-09-13 at 5 07 21 pm

versus 

screen shot 2018-09-13 at 5 07 08 pm


skllcrn:
- I agree, looping should be on by default for all formats that support it.
- Previously we ended up shortening the path to .../Kaptures, I believe the reasoning for that is we never know how long the path is and that could cause issues in the UI. What are your thoughts on it?

sindresorhus:
> I agree, looping should be on by default for all formats that support it.

👍 

> Previously we ended up shortening the path to .../Kaptures, I believe the reasoning for that is we never know how long the path is and that could cause issues in the UI. What are your thoughts on it?

The `../Kaptures` always confused me. It's not clear where it actually is. I think it would be better to have the path as `~/Movies/Kaptures` and if it's too long, we can just shorten it with `text-overflow: ellipsis`.

.gitignore
```diff
@@ -1,3 +1,5 @@
+.DS_Store
```
sindresorhus: This should be in your own global `.gitignore`. Only project-specific ignores should be in the repo.

skllcrn:
>  I think it would be better to have the path as ~/Movies/Kaptures and if it's too long, we can just shorten it with text-overflow: ellipsis.

Sounds good @sindresorhus, we can also make the path clickable to open the specified folder!

renderer/components/preferences/categories/settings.js
```diff
@@ -60,23 +71,21 @@ class Advanced extends React.Component {
...
-          title="Do Not Disturb"
+          title="Silence notifications"
           subtitle="Activate “Do Not Disturb” while recording"
         >
           <Switch checked={doNotDisturb} onClick={() => toggleSetting('doNotDisturb')}/>
         </Item>
         <Item
-          title="Capture at 60 FPS"
-          subtitle="The default is 30 FPS"
+          title="Loop Kaptures"
```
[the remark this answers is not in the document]
karaggeorge: "Recordings"?

karaggeorge:
@skllcrn @sindresorhus 
Sounds good.
- What should we use instead of "Kaptures" for looping. "Recordings"?
- If the path is clickable (open in finder) should it be styled differently to appear clickable? 
- Is there an easy way to get the path as `~/...` instead of `/Users/username/...`?

albinekb:
>What should we use instead of "Kaptures" for looping. "Recordings"?

GIF is the only format with support for looping right? Then maybe "Loop gifs"

sindresorhus:
> What should we use instead of "Kaptures" for looping. "Recordings"?

👍, but @skllcrn should weigh in here too.

> If the path is clickable (open in finder) should it be styled differently to appear clickable?

Style it as a link?

> Is there an easy way to get the path as ~/... instead of /Users/username/...?

https://github.com/sindresorhus/tildify

skllcrn:
I think it's a nice touch to call them "Kaptures" in the UI, we do that for the cursor settings in Preferences too, as well as the default filename of exports. 

> Style it as a link?

👍

skllcrn:
When attempting to export using a Plugin with a missing or incorrect configuration, show a dialog with options to cancel or open the configuration in question: 

Alert

Once a Plugin has been installed, send a notification (banner) to indicate that it's ready for use. In the case of a Plugin that requires configuration there's an action associated with the notification (alert) and it does not dismiss automatically: 

notifications

sindresorhus:
> Review your plugin configuration to continue exporting

Should mention the name of the plugin.

---

The alert dialog should also use a more descriptive button than `OK`. I would use `Review`.

karaggeorge:
Those alerts and notifications will probably not be part of this PR, but the next one I'm working on for the plugin configuration

skllcrn:
Awesome @karaggeorge, want me to create a separate issue for it?

sindresorhus:
@skllcrn Yes, do it so it's not forgotten.

renderer/components/preferences/categories/settings.js
```diff
@@ -80,10 +80,10 @@ class Settings extends React.Component {
           <Switch checked={doNotDisturb} onClick={() => toggleSetting('doNotDisturb')}/>
         </Item>
         <Item
-          title="Loop Kaptures"
+          title="Loop Recordings"
```
sindresorhus: @skllcrn wanted `Kaptures` to stay in the UI, but let's keep using `recordings` for anything non-UI, like variable names and the settings.

main/common/settings.js
```diff
@@ -13,7 +13,7 @@ const store = new Store({
     highlightClicks: false,
     hideDesktopIcons: false,
     record60fps: false,
-    loopKaptures: true,
+    loopRecordings: true,
```
albinekb: Isn't it weird calling this `recordings`? The recordings are not looped, it's the exported Kaptures that are.

PR #500, opened 2018-09-14
Refactor plugins

karaggeorge:
### Tasks
- Split plugins into two pages
- Don't run `npm prune` when uninstalling but run on startup
- Add screens for no plugins installed and for network error

### Preview

kapture 2018-09-13 at 20 15 02
kapture 2018-09-13 at 20 25 58


skllcrn:
Really appreciate your attention to detail mate, it's looking fantastic! Do you think not running `prune` on uninstall might stop people from being able to fix potential issues by simply reinstalling plugins? Might not be clear that it's necessary to restart the entire app.

main/index.js
```diff
@@ -29,6 +29,7 @@ app.on('ready', async () => {
   app.dock.hide();
 
   // Ensure all plugins are up to date
+  plugins.prune().catch(() => {});
   plugins.upgrade().catch(() => {});
```
sindresorhus: I don't think we should run both these commands at the same time. They might conflict.

main/common/plugins.js
```diff
@@ -79,8 +79,9 @@ class Plugins {
     this._modifyMainPackageJson(pkg => {
       delete pkg.dependencies[name];
     });
+  }
 
-    // Intentionally not awaited as it can just finish in the background
+  async prune() {
     this._runNpm('prune');
```
sindresorhus: It should be awaited here, as logic for deciding to await or not is up to the consumer of this method.

sindresorhus:
When on the "Installed" tab, and clicking the checkbox on a plugin, the plugin disappears too abruptly in the middle of the animation. Maybe it could wait 0.5s and then fade out?

renderer/components/preferences/categories/plugins/index.js
```diff
@@ -0,0 +1,187 @@
...
+                    onClick={install}/>
+                )
+              }
+            </div>
+            <div className="tab">
+              {
+                pluginsInstalled.length === 0 ? (
+                  <EmptyTab
+                    title="No plugins, yet!"
+                    subtitle="Customize Kap your liking with Plugins."
```
sindresorhus: `Plugins` => `plugins`

renderer/components/preferences/categories/plugins/index.js
```diff
@@ -0,0 +1,187 @@
...
+                    disabled={Boolean(pluginBeingInstalled)}
+                    onClick={install}/>
+                )
+              }
+            </div>
+            <div className="tab">
+              {
+                pluginsInstalled.length === 0 ? (
+                  <EmptyTab
+                    title="No plugins, yet!"
```
sindresorhus: > No plugins yet

renderer/components/preferences/categories/plugins/index.js
```diff
@@ -0,0 +1,187 @@
...
+
+          .plugins-nav div {
+            margin-right: 16px;
+            height: 100%;
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            padding-bottom: 2px;
+            font-size: 1.2rem;
+            color: #007AFF;
```
sindresorhus: Use lowercase HEX for consistency.

renderer/containers/preferences.js
```diff
@@ -43,34 +42,43 @@ export default class PreferencesContainer extends Container {
     })();
   }
 
+  fetchFromNpm = async () => {
+    this.plugins.getFromNpm().then(plugins => {
```
sindresorhus: Use `await`

main/index.js
karaggeorge: Hmm, it was doing it `onTransitionEnd`. Maybe make the transition longer? Or just wait a bit after the transition?

karaggeorge:
@skllcrn I think if you uninstall/install, even without running prune, the install command should get the plugin to the latest version. If there are no new versions it probably won't do much, but I don't see how that would fix any issues. It would remove and install the same code (I think)?

@sindresorhus About the uninstalling animation, it's removing it from the list `onTransitionEnd`. Maybe make the transition longer? Or just wait a bit after the transition?

renderer/components/preferences/categories/plugins/index.js
albinekb: `prettier` does this automatically, should we add it?

karaggeorge: I usually use `xo --fix` if I have lots of errors, and it takes care of most of them

albinekb: But it doesn't format the hex colors.

skllcrn:
Oops, commented on the wrong issue 🙈 https://github.com/wulkano/kap/pull/499#issuecomment-421936501

karaggeorge:
kapture 2018-09-17 at 14 15 25

Adding `500 ms` delay after the uninstall animation

renderer/components/preferences/categories/plugins/index.js
karaggeorge: We can discuss pros/cons about adding prettier, but I think we can do that in a different Issue/PR 👍

sindresorhus:
This is looking great ✨

PR #503, opened 2018-09-19
Upgrade to Electron 3

sindresorhus:
Electron 3 comes with newer Node.js version, so we can use object spread in the main process now too.

I've been playing with it for a while and everything seems to be working.

### DIFF
--- a/main/export.js
+++ b/main/export.js
@@ -62,7 +62,7 @@ class Export {
           });
         }
         resolve();
-      } catch (error) {
+      } catch (_) {
         reject();
       }
     });
--- a/main/share-service-context.js
+++ b/main/share-service-context.js
@@ -43,7 +43,7 @@ class ShareServiceContext {
       return;
     }
 
-    const request = got(url, Object.assign({}, options, {useElectronNet: false}));
+    const request = got(url, options);
 
     this.requests.push(request);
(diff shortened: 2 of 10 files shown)


main/share-service-context.js
```diff
@@ -43,7 +43,7 @@ class ShareServiceContext {
       return;
     }
 
-    const request = got(url, Object.assign({}, options, {useElectronNet: false}));
+    const request = got(url, options);
```
sindresorhus: `useElectronNet: false` is the default already in the got version we're using.

main/export.js
```diff
@@ -62,7 +62,7 @@ class Export {
           });
         }
         resolve();
-      } catch (error) {
+      } catch (_) {
         reject();
```
albinekb: Why not reject with the error here? 🤔

sindresorhus: I was going to ask @karaggeorge about that. I changed it to `_` to make it clear that the error is ignored.

albinekb: Yep, better for another PR perhaps

karaggeorge: Hmm, I don't remember. Probably should `reject(error)` but yeah we can leave that for a separate PR if you want

PR #504, opened 2018-09-19
Plugin config UI

### LINKED ISSUE #286: Settings button for plugins that provide it
When doing the plugins pane, we forgot about adding a `Settings` button to plugins in the list that provide settings. This button should simply open the config file.

@skllcrn Any opinions on how it should look?

### LINKED ISSUE #501: Plugin notifications and alerts
Via #500 (https://github.com/wulkano/kap/pull/500#issuecomment-421945971) 

> When attempting to export using a Plugin with a missing or incorrect configuration, show a dialog with options to cancel or open the configuration in question: 
> 
> Alert
> 
> Once a Plugin has been installed, send a notification (banner) to indicate that it's ready for use. In the case of a Plugin that requires configuration there's an action associated with the notification (alert) and it does not dismiss automatically: 
> 
> Notifications

### LINKED ISSUE #505: The plugin discover tab needs an empty state
When you have installed all the plugins, this is what you see:

screen shot 2018-09-20 at 12 08 19

We could have something fun, like:

> Wow. You sure like plugins.

karaggeorge:
Fixes #286 
Fixes #501 
Fixes #505

Adds configuration window for plugins that provide a schema and checks for valid config before exporting otherwise redirects to the config page for that plugin

Also adds notifications around installing plugins


sindresorhus:
- I think the pensil button should have some kind of slight hover effect. It's not entirely clear right now that it's clickable.
- It's not entirely clear what the warning icon is for for new users. Would be good to have some kind of tooltip explaining that required config is missing.
- Clicking the warning icon should also open the config modal.
- "Open file" => "Open config file" Better to be explicit
- "View on GitHub" => "View plugin on GitHub" (I first thought it would show me the config file on GitHub)
- The "Open file" button has a folder icon. Should find something more suitable.
- Would be nice if "View on GitHub" used a octocat icon. Though not super important.
- Would be nice to fade in a dark transparent overlay over preferences window when the modal opens. To give focus on the modal.
- The `kap-now` plugin has a link in it's config description: `Create one here https://zeit.co/account/tokens`. Would be nice if links in descriptions were linkified. (https://github.com/sindresorhus/linkify-urls)
- The field `title` should be distinguished from the `description`. You can see they're identical here: screen shot 2018-09-20 at 12 06 42
- Plugins need a place to specify general plugin info, like where to see the API docs, etc. Ideas for how this could be specified in a plugin? The info would be shown at the top of the plugin config modal.
- It should not show the config button when the config object is empty. Like here: https://github.com/lucaperret/kap-transfer/blob/baeb95439e959131b02b5a84cfc992f970689176/index.js#L22

main/config.js
```diff
@@ -0,0 +1,37 @@
...
+
+  loadRoute(configWindow, 'config');
+
+  configWindow.webContents.on('did-finish-load', async () => {
+    ipc.callRenderer(configWindow, 'plugin', pluginName);
+    configWindow.show();
+  });
+
+  return new Promise(resolve => {
+    configWindow.on('closed', resolve);
```
sindresorhus: https://github.com/sindresorhus/p-event is good for this kind of thing.

main/export-list.js
```diff
@@ -120,6 +121,23 @@ class ExportList {
       }
     }
 
+    if (!newExport.plugin.isConfigValid()) {
+      const result = dialog.showMessageBox({
+        type: 'error',
+        buttons: ['Configure', 'Cancel'],
+        defaultId: 0,
+        message: 'Error in plugin configuration',
+        detail: 'Review your plugin configuration for "now" to continue exporting',
```
sindresorhus: Slight tweaked. Better?

> Review the config for the `now` plugin to continue exporting

main/export-list.js
```diff
@@ -120,6 +121,23 @@ class ExportList {
       }
     }
 
+    if (!newExport.plugin.isConfigValid()) {
+      const result = dialog.showMessageBox({
+        type: 'error',
+        buttons: ['Configure', 'Cancel'],
+        defaultId: 0,
+        message: 'Error in plugin configuration',
```
sindresorhus: `configuration` => `config`

main/preferences.js
```diff
@@ -31,8 +32,11 @@ const openPrefsWindow = () => {
 
   loadRoute(prefsWindow, 'preferences');
 
-  ipcMain.once('preferences-ready', () => {
-    prefsWindow.show();
+  return new Promise(resolve => {
+    ipcMain.once('preferences-ready', () => {
```
sindresorhus: Could use `p-event` here too.

renderer/components/config/index.js
```diff
@@ -0,0 +1,174 @@
...
+
+          .service-nav div {
+            margin-right: 16px;
+            height: 100%;
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            padding-bottom: 2px;
+            font-size: 1.2rem;
+            color: #007AFF;
```
sindresorhus: We should really make use of CSS variables. I see this color used constantly. Using CSS variables will be important for adding dark mode support too. // @skllcrn

sindresorhus:
There's a proptype error in the preference window:

```
main.js:664 Warning: Failed prop type: Invalid prop `selected` of type `boolean` supplied to `Select`, expected `string`.
    in Select (at settings.js:117)
    in Settings (at index.js:21)
    in Consumer (created by Subscribe)
    in Subscribe (at index.js:14)
    in Unknown (at index.js:39)
    in div (at index.js:34)
    in Categories (at index.js:21)
    in Consumer (created by Subscribe)
    in Subscribe (at index.js:14)
    in Unknown (at preferences.js?entry:24)
    in Provider (created by Consumer)
    in Consumer (created by Provider)
    in Provider (at preferences.js?entry:20)
    in div (at preferences.js?entry:19)
    in PreferencesPage (created by Container)
    in AppContainer (created by Container)
    in Container (created by App)
    in div (created by App)
    in App
```

renderer/components/config/index.js
karaggeorge: I'll open a new PR that takes out the colors into variables 👍

main/export-list.js
karaggeorge: Wow `now` was also hardcoded instead of using the pluginName 😆

karaggeorge:
@sindresorhus 

> The kap-now plugin has a link in it's config description: Create one here https://zeit.co/account/tokens. Would be nice if links in descriptions were linkified. (sindresorhus/linkify-urls)

Any way of using this without using `dangerouslySetInnerHTML`? xo doesn't like it

> Plugins need a place to specify general plugin info, like where to see the API docs, etc. Ideas for how this could be specified in a plugin? The info would be shown at the top of the plugin config modal.

Maybe a reserved `description` key on the config object that we extract before we create the schema?

karaggeorge:
@sindresorhus I think I fixed almost all those comments. The only thing left is that plugin description which we'll have to figure out and not sure what to do about the `dangerouslySetInnerHTML` with the `linkify-urls`. For now I've disabled the eslint rule for that line

sindresorhus:
> The only thing left is that plugin description which we'll have to figure out and not sure what to do about the dangerouslySetInnerHTML with the linkify-urls. For now I've disabled the eslint rule for that line

Use the `type` option in `linkify-urls` to get a DOM element.

sindresorhus:
> Maybe a reserved description key on the config object that we extract before we create the schema?

How about a top-level `configMeta` key? (Open to suggestion for a better name)

```js
const giphy = {
	title: 'Share to GIPHY',
	formats: ['gif'],
	action,
	configMeta: {
		description: 'sdasdas'
	},
	config: {
		apiKey: {
			title: 'API key',
			type: 'string',
			minLength: 13,
			default: '',
			required: true
		}
	}
};
```

sindresorhus:
Can you fade in and out the overlay? It's a bit abrupt now. Can probably use CSS animations for that.

sindresorhus:
The linkified URL currently opens inside the modal. You want to add `target="_blank"`.

sindresorhus:
The red color used to show the validation errors is too bright.

karaggeorge:
@sindresorhus 
> Use the type option in linkify-urls to get a DOM element.

Will that work with React? Can you pass a DOM element as a component?

About the configMeta, that idea doesn't sound bad to me. Can we leave that for later/another PR since that will need doc changes etc? 

I'll update the overlay fade in/out 👍

sindresorhus:
> About the configMeta, that idea doesn't sound bad to me. Can we leave that for later/another PR since that will need doc changes etc?

Yup. I'll open an issue.

sindresorhus:
> Will that work with React? Can you pass a DOM element as a component?

Yes: https://stackoverflow.com/questions/49525599/how-to-pass-a-javascript-generated-dom-element-to-react-render-function

sindresorhus:
The config modal page needs some bottom padding so the button at the end (View plugin on GitHub) are not so close to the "Done" button when fully scrolled down.

sindresorhus:
The GitHub icon is larger than the file icon, since the file icon path has some padding.

screen shot 2018-09-20 at 22 56 01

sindresorhus:
The "all the things" image is blurry on retina screens. @skllcrn Do you have a higher resolution version of it?

sindresorhus:
> It should not show the config button when the config object is empty. Like here: lucaperret/kap-transfer:index.js@baeb954#L22

It still shows the config button.

sindresorhus:
Maybe https://stackoverflow.com/a/43902326/64949

But probably easier to just use something more React-y, like https://github.com/tasti/react-linkify

renderer/components/preferences/item/index.js
```diff
@@ -75,6 +76,7 @@ class Item extends React.Component {
             display: flex;
             max-width: 100%;
             padding: ${onClick ? '16px' : '32px'} 16px;
+            margin-bottom: ${last ? '16px' : '0'};
```
sindresorhus: Why not just do this with pure CSS?

```css
.container:last-child {
	margin-bottom: 16px;
}
```

karaggeorge: This was to push that github button up a bit. This class is used all over settings, so I could do it like that but I would still have to add a class so it can be toggled

sindresorhus: Ok. I missed the context. It's fine then.

sindresorhus:
It's perfect now. Super work, @karaggeorge (as always!)

skllcrn:
> I think the pencil button should have some kind of slight hover effect. It's not entirely clear right now that it's clickable.

Let's go with a subtly darker blue: #006BE0

> It's not entirely clear what the warning icon is for for new users. Would be good to have some kind of tooltip explaining that required config is missing.

Can we use an `alt` attribute for this? Agree that the error icon (and surrounding area) should open the config modal.

PR #507, opened 2018-09-22
Add keyboard shortcut setting

albinekb:
### Todo
- [x] Implement design
- [x] Implement unregister/register all when shortcuts are toggled
- [x] Implement validation
- [x] Shake animation?
- [x] Add X icon

This should be backwards compatible with the old setting too.


main/global-accelerators.js
```diff
@@ -0,0 +1,53 @@
...
+  try {
+    const shortcutAccelerator = shortcutToAccelerator(shortcut);
+    globalShortcut.register(shortcutAccelerator, openCropperWindow);
+  } catch (error) {
+    console.error('Error registering shortcut', shortcut);
+  }
+};
+
+const registrerFromStore = () => {
+  if (store.get('recordKeyboardShortcut')) {
```
karaggeorge: We could probably rename this to something generic like `useKeyboardShortcuts`

main/global-accelerators.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+const {globalShortcut} = require('electron');
+const ipc = require('electron-better-ipc');
+const shortcutToAccelerator = require('../renderer/utils/shortcut-to-accelerator');
+const store = require('./common/settings');
+const {openCropperWindow} = require('./cropper');
+
+const registerShortcut = shortcut => {
```
karaggeorge: Could have this accept two arguments, `shortcut` and `action` so we can re-use it if we have different shortcuts in the future. And then we can do `registerShortcut(cropperShortcut, openCropperWindow)` in the `registerFromStore` method

main/global-accelerators.js
```diff
@@ -0,0 +1,53 @@
...
+    if (shortcut) {
+      registerShortcut(shortcut);
+    }
+    store.set(setting, shortcut);
+  });
+
+  // Register keyboard shortcuts from store
+  registrerFromStore();
+
+  store.events.on('change', registrerFromStore);
```
karaggeorge: Can use [onKeyChange](https://github.com/sindresorhus/electron-store#ondidchangekey-callback) to only register again if the specific key changes, and not if any store key changes.

Also if we listen for changes like this, we probably don't even need the `ipc` event as long as we update the setting in the store in the prefs window

albinekb: Yeah, I was thinking that too but I didn't know about onKeyChange! Nice!

main/global-accelerators.js
albinekb: yep, just didnt want to add too much features here since there is only one shortcut now

sindresorhus:
When I press `Option` key, it shows the `Option` key and also an empty white box.

renderer/utils/shortcut-to-accelerator.js
```diff
@@ -0,0 +1,16 @@
+const shortcutToAccelerator = shortcut => {
+  const {metaKey, altKey, ctrlKey, shiftKey, char} = shortcut;
+  if (!char) {
+    throw new Error(`shortcut needs char ${JSON.stringify(shortcut)}`);
```
sindresorhus: > Shortcut needs character

albinekb:
>When I press Option key, it shows the Option key and also an empty white box.

Fixed

main/global-accelerators.js
```diff
@@ -0,0 +1,67 @@
...
+      console.error('Error unregestering old shortcutAccelerator', error);
+    }
+    store.set(setting, shortcut);
+    if (shortcut) {
+      registerShortcut(shortcut, handlers[setting]);
+    }
+  });
+
+  ipc.answerRenderer('toggle-shortcuts', ({enabled}) => {
+    if (enabled) {
```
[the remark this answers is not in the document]
albinekb: Yep, that will work 👍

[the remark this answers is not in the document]
albinekb: Hmm, one thing with that is that it could become a race condition, so if you disabled it in settings, and the store hasn't synced yet, it will say it's disabled/enabled wrongly, I think it's best to keep it as-is.

PR #510, opened 2018-09-23
Add auto-update functionality

### LINKED ISSUE #258: Use `electron-updater`
Right now we're using the built-in updater in Electron with a custom server. The built-in updater in Electron requires a server and a lot of boilerplate code in Kap.

I suggest we switch to [`electron-updater`](https://www.npmjs.com/package/electron-updater). It's part of `electron-builder` and will let us just build on Travis or CircleCI and publish to GitHub. No need for us to maintain a server. It just works. It also now supports delta updates for smaller updates, [staged rollouts](https://www.electron.build/auto-update#staged-rollouts), prerelease support, etc.

I've been using it with great success on Caprine for a long time.

This is not high priority. Just opening this since I got the idea now.

Thoughts?

### LINKED ISSUE #451: Clicking upgrade notification closes Kap without saving
**Platform:** macOS 10.13.4
**Kap Version:** 2.1.2

### Steps to reproduce

1. Start and finish a recording, keeping the window open and don't save yet
2. Click the notification to get the latest version of Kap
3. Watch your work disappear 😀 

### Current behaviour

1. Simply quits without saving

### Expected behaviour

1. Maybe prompt me to save before closing

### Workaround

1. Think before pressing "Upgrade" 😂

### LINKED ISSUE #481: „Check for updates…“ option has no feedback
Clicking on „Check for updates...“ contextual menu of the Kap menu icon, checks if a new version is available or not.

I think, if it doesn’t find a new version, the menu item is only grayed out/disabled.

It should send a user feedback in any case - whether a new version is available or also if not. It‘s confusing triggering that contextual menu manually and not receiving any user feedback on the action‘s result.

sindresorhus:
Fixes #258 
Fixes #451
Fixes #481

It checks for new updates on startup and once every hour. If there is an update, it's downloaded in the background and installed on the next launch.


main/index.js
```diff
@@ -33,6 +37,21 @@ const initializePlugins = async () => {
   }
 };
 
+const checkForUpdates = () => {
+  if (is.development) {
+    return false;
+  }
+
+  autoUpdater.logger = log;
+  autoUpdater.logger.transports.file.level = 'info';
```
sindresorhus: This is so we can debug auto-update functionality in Console.app

albinekb: Maybe add a comment for the future?

karaggeorge:
Does this take care of the whole notifying the user about updating and restarting the app?

main/index.js
```diff
@@ -33,6 +37,22 @@ const initializePlugins = async () => {
...
+  if (is.development) {
+    return false;
+  }
+
+  // For auto-update debugging in Console.app
+  autoUpdater.logger = log;
+  autoUpdater.logger.transports.file.level = 'info';
+
+  setInterval(() => {
+    autoUpdater.checkForUpdates();
```
albinekb: I think this should use `checkForUpdatesAndNotify`? 

>Asks the server whether there is an update, download and notify if update available.


https://www.electron.build/auto-update

skllcrn: #481, #451

sindresorhus:
> Does this take care of the whole notifying the user about updating and restarting the app?

No, I thought we had concluded to have silent auto-updates. Per #481.

karaggeorge:
I'm okay with no option to check. I mean it does happen automatically. But without a notification when does the update happen? On restart?

sindresorhus:
> though I still think it's useful to have the option to manually check for updates

What practical purpose would it have? We automatically check every hour. If you have a button, users will think they need to click it, so they will waste time doing so. Better to just have good automatic behavior whenever possible.

sindresorhus:
> But without a notification when does the update happen? On restart?

See my PR description.

PR #512, opened 2018-09-23
Add Sentry

### LINKED ISSUE #471: Disable Sentry error logging when Analytics if toggled off
Disable `raven.js` when the analytics setting in preferences is off.

sindresorhus:
I've confirmed it to work:

- main: https://sentry.io/wulkano-l0/kap/issues/698345992/?query=is:unresolved
- renderer: https://sentry.io/wulkano-l0/kap/issues/698347095/?query=is:unresolved

Fixes #471


renderer/pages/config.js
```diff
@@ -5,6 +5,8 @@ import {ConfigContainer} from '../containers';
 import Config from '../components/config';
 import WindowHeader from '../components/window-header';
 
+import '../utils/sentry';
```
albinekb: I wonder if we can add it to _app.js, and that way dont need to add it to every page 🤔 See https://github.com/zeit/next.js#custom-app

albinekb: https://github.com/zeit/next.js/blob/canary/examples/with-sentry/pages/_app.js

sindresorhus:
Updated

Confirmed working: https://sentry.io/wulkano-l0/kap/issues/699034690/?query=is:unresolved

PR #521, opened 2018-09-24
Close preferences window before recording

karaggeorge:
This is to avoid a number of weird bugs around changing recording settings while actually recording

### DIFF
--- a/main/common/aperture.js
+++ b/main/common/aperture.js
@@ -6,6 +6,7 @@ const dnd = require('@sindresorhus/do-not-disturb');
 const createAperture = require('aperture');
 
 const {openEditorWindow} = require('../editor');
+const {closePrefsWindow} = require('../preferences');
 const {setRecordingTray, disableTray} = require('../tray');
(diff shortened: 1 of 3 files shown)


karaggeorge:
@sindresorhus how can you get to it while recording? The action bar hides and the tray icon turns into the stop recording button

sindresorhus:
I can also right-click the menu bar icon during recording to get to `Preferences…`.

karaggeorge:
@sindresorhus that's true.

This is a different approach: https://github.com/wulkano/kap/pull/523

If we still want to keep this and hide the Preferences, I'll work on updating it but it doesn't need to be in the beta

sindresorhus:
> If we still want to keep this and hide the Preferences, I'll work on updating it but it doesn't need to be in the beta

What's your opinion? I think it's sensible to disable `Preferences…` during recording. We can actually just disable right-click on the icon altogether during recording instead.

karaggeorge:
Yeah I think that's the easiest way to eliminate all edge cases

karaggeorge:
@sindresorhus This should be good now

PR #523, opened 2018-09-24
Fix desktop icons and do not disturb behavior

karaggeorge:
We can save what the setting was when we start recording and use that when we stop, instead of pulling it from the store again, to avoid mismatch if the settings changed mid-recording.

This way we don't have to hide the Preferences window


sindresorhus:
> This way we don't have to hide the Preferences window

I still think we should hide it during recording though. You wouldn't want it in the recording and there might be other synchronization issues we haven't thought of.

PR #526, opened 2018-09-24
Ensure muting in the editor removes audio from the export

### LINKED ISSUE #525: Audio included even when disabled in the Editor
**macOS version:**    10.13.6
**Kap version:**      3.0.0-beta.1

#### Steps to reproduce
* Enable Audio Recording in Preferences
* Create a Kapture
* Select a format that supports audio
* Mute the audio in the Editor
* Export 

#### Current behavior
* Audio is still included in the export 

#### Expected behavior
* Audio should not be included in the export

#### Workaround
* Completely disable the Audio Recording setting

karaggeorge:
Fixes #525

### DIFF
--- a/renderer/containers/editor.js
+++ b/renderer/containers/editor.js
@@ -158,7 +158,7 @@ export default class EditorContainer extends Container {
         fps,
         startTime,
         endTime,
-        muted: isMuted
+        isMuted
       },
       inputPath: filePath,
       pluginName,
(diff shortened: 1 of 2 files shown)


renderer/containers/editor.js
```diff
@@ -158,7 +158,7 @@ export default class EditorContainer extends Container {
         fps,
         startTime,
         endTime,
-        muted
+        muted: isMuted
```
sindresorhus: I think `muted` should be renamed to `isMuted`, both here and on the main process side. Better to be consistent.

karaggeorge: eeae224

PR #527, opened 2018-09-26
Refactor export options in the editor window

### LINKED ISSUE #238: Default share service
Many people would like to set a default export service for each format. That way you could "streamline" the pipeline from a recorded kapture to shared *insert format* file. 

For example: I use https://github.com/albinekb/kap-draggable to drag-n-drop my gifs onto github PRs, if I had a way to select this drag-n-drop as the default share service for GIFs, I could save having to do the tedious task to select the same share setting every time.

Here's my initial version of this, however, we'd like to improve on this idea and make it even better when we implement it: https://github.com/wulkano/kap/pull/198

### LINKED ISSUE #516: Update Export options after installing a plugin
If you have the Editor open, and install a plugin, that plugin is not available (you need to close the window and open again). We should somehow make it update so the plugin is usable when it's installed.

karaggeorge:
Fixes #238 
Fixes #516

### DIFF
--- a/main/common/plugins.js
+++ b/main/common/plugins.js
@@ -9,7 +9,7 @@ const ipc = require('electron-better-ipc');
 const {app, Notification} = electron;
 
 const Plugin = require('../plugin');
-const {updateExportOptions} = require('../editor');
+const {updateExportOptions} = require('../export-options');
 const {openConfigWindow} = require('../config');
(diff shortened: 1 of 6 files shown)


karaggeorge:
@sindresorhus good catch. Updated

PR #528, opened 2018-09-26
Refactor main bar inputs

### LINKED ISSUE #539: Crop window's starting point slightly off
Here's a video: https://photos.app.goo.gl/wsu9oFeGJ648QNMNA

**macOS Version:** 10.13.6
**Kap Version:** 3.0.0-beta.1.1006

#### Steps to reproduce
- Start Kap
- Click on the icon in the menubar
- Try to select an area

#### Current behaviour
The starting point is slightly off

#### Expected behaviour
The starting point should be where I clicked
_________________________
It's not fully consistent but happens quite regularly

karaggeorge:
Aims to fix the weird behaviors experienced with the main bar width/height/ratio inputs

---

Fixes #539

### DIFF
--- a/renderer/components/action-bar/controls/advanced.js
+++ b/renderer/components/action-bar/controls/advanced.js
@@ -15,7 +15,8 @@ import {
   handleWidthInput,
   handleHeightInput,
   buildAspectRatioMenu,
-  handleInputKeyPress
+  handleInputKeyPress,
+  RATIOS
 } from '../../../utils/inputs';
 
 const advancedStyles = css`
@@ -35,6 +36,8 @@ const stopPropagation = event => event.stopPropagation();
 class Left extends React.Component {
   state = {}
 
+  select = React.createRef();
+
   static getDerivedStateFromProps(nextProps, prevState) {
     const {ratio, isResizing, setRatio} = nextProps;
 
@@ -48,6 +51,21 @@ class Left extends React.Component {
     return null;
   }
 
+  openMenu = () => {
+    const {ratio} = this.props;
+    const boundingRect = this.select.current.getBoundingClientRect();
+    const {top, left} = boundingRect;
+    const selectedRatio = ratio.join(':');
+    const index = RATIOS.findIndex(r => r === selectedRatio);
+    const positioningItem = index > -1 ? index : RATIOS.length;
+
+    this.state.menu.popup({
+      x: Math.round(left),
+      y: Math.round(top),
+      positioningItem
+    });
+  }
+
   render() {
     const {toggleAdvanced, toggleRatioLock, ratioLocked, ratio = []} = this.props;
 
@@ -56,7 +74,7 @@ class Left extends React.Component {
         <div className="back">
           <BackIcon onClick={toggleAdvanced}/>
         </div>
-        <div className="select" onClick={() => this.state.menu.popup({})} onMouseDown={stopPropagation}>
+        <div ref={this.select} className="select" onClick={this.openMenu} onMouseDown={stopPropagation}>
           <span>{ratio[0]}:{ratio[1]}</span>
           <DropdownArrowIcon size="18px"/>
         </div>
--- a/renderer/components/action-bar/controls/advanced.js
+++ b/renderer/components/action-bar/controls/advanced.js
@@ -130,22 +130,32 @@ class Right extends React.Component {
     this.heightInput = React.createRef();
   }
 
-  onWidthChange = event => {
+  onWidthChange = (event, {ignoreEmpty} = {}) => {
     const {x, y, setBounds, ratioLocked, ratio, setWidth} = this.props;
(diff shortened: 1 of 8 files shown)


renderer/components/action-bar/controls/advanced.js
```diff
@@ -223,18 +225,38 @@ class Right extends React.Component {
 Right.propTypes = {
   x: PropTypes.number,
   y: PropTypes.number,
-  width: PropTypes.number,
-  height: PropTypes.number,
+  width: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
sindresorhus: It feels a bit dirty that `width` can be both a number and a string...

karaggeorge:
@sindresorhus updated 👍

renderer/containers/action-bar.js
```diff
@@ -17,18 +17,18 @@ export default class ActionBarContainer extends Container {
 
     this.settings = this.remote.require('./common/settings');
     this.state = {
-      cropperWidth: 0,
-      cropperHeight: 0
+      cropperWidth: '0',
+      cropperHeight: '0'
```
sindresorhus: Why did you go with saving them as strings and not numbers? Just curious. I haven't looked closely at the code yet. Just seems like the obvious would be to store as numbers.

karaggeorge: Because the validation is `debounce`d, so the user could enter a string that is not a number. Those values there are what the user sees in the input field. The `width` and `height` in the other container are the actual dimensions of the cropper. And they are synced when the debounced func runs

karaggeorge:
These are good points. We've talked about most of them in the past and I've worked on some but last time I did was when I first started the re-write. I'll work on it a bit more, see how it goes, and if it seems like it's going to be a lot of extra work, I'll see about splitting into two PRs

karaggeorge:
This also fixes #539 now

renderer/components/action-bar/controls/main.js
```diff
@@ -85,6 +97,18 @@ MainControls.Left = connect(
 )(Left);
 
 class Right extends React.Component {
+  icon = React.createRef();
+
+  openMenu = () => {
```
sindresorhus: Should really make a reusable component for this as a Electron dropdown menu is used in many places.

karaggeorge:
@sindresorhus it's all of them (hopefully) 😄 I went ahead and checked them though

sindresorhus:
> Also, for the ratio menu I used the itemPositioning option of the menu api to show the selected one under the cursor. I saw it was osx specific, so I'm guessing is standard osx practice?

Not cursor, but yes it's common practise to show it under the element that triggered the menu. It needs to be better aligned though. When the menu is open, the active menu item should be vertically centered to our white dropdown widget.

sindresorhus:
I'm getting this error every time I click the Kap menu bar icon:

screen shot 2018-09-30 at 02 45 20

sindresorhus:
> Selecting an aspect ratio should active the width/height lock. Otherwise, what's the point of selecting aspect ratio if it's not enforced when you change width or height.

Doesn't look like this is done. If I select an item in the aspect ratio dropdown, the lock button is not highlighted.

skllcrn:
I can't recreate the issue @sindresorhus is having when clicking the menubar icon, I can however confirm that selecting an spect ratio does not lock the dimensions, which I think it should. 

Overall this is such a massive improvement, after testing the branch for a few days I haven't had a single issue with it and it feels great to use. The only thing I might add is for the inputs to only allow number input, however that's not really common practice. In a future update a nice to have would be the ability to do simple math in the inputs, e.g. 64+64 would change the input to 128 on enter/blur.

sindresorhus:
@karaggeorge Can you fix the merge conflict?

sindresorhus:
I'm no longer getting the `focus` error when clicking the menu bar icon. Weird...

sindresorhus:
The only thing left is https://github.com/wulkano/kap/pull/528#issuecomment-425671151

karaggeorge:
@sindresorhus I'll fix the merge conflicts when I get some down-time. About the other comment, I tried to get it as close to centered as I could. Bringing it one more pixel makes it even worse. Any idea of a consistent way to do it?

sindresorhus:
> About the other comment, I tried to get it as close to centered as I could. Bringing it one more pixel makes it even worse. Any idea of a consistent way to do it?

Hmm, actually it looks enough for me now. Must have been something weird when I tried earlier as I could swear it was not aligned then.

karaggeorge:
Should be all good. Also, I found a bug where the validation was pulling the `primary` display screen's size instead of the one we sent through ipc. So in external displays, the cropper was limited to the primary display size. Should be fixed now

sindresorhus:
Looks great!

PR #531, opened 2018-09-26
Upgrade dependencies

sindresorhus:
Most of the changes are XO fixes, like removing `async` for functions that does not need to be `async` (not using `await`).

### DIFF
--- a/main/common/aperture.js
+++ b/main/common/aperture.js
@@ -43,7 +43,7 @@ const startRecording = async options => {
     cropArea: cropperBounds,
     showCursor,
     highlightClicks,
-    displayId: String(displayId)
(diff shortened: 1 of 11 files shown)


main/common/aperture.js
```diff
@@ -43,7 +43,7 @@ const startRecording = async options => {
     cropArea: cropperBounds,
     showCursor,
     highlightClicks,
-    displayId: String(displayId)
+    screenId: displayId
```
sindresorhus: This is for the Aperture upgrade. See: https://github.com/wulkano/aperture/commit/38f82228c02e1e357369768d739792f835b07944

PR #532, opened 2018-09-27
Add custom "about" window

### LINKED ISSUE #511: Link to privacy statement/policy in about
https://github.com/wulkano/kap/pull/492#issuecomment-409608352
 
https://github.com/wulkano/kap/blob/master/PRIVACY.md

### LINKED ISSUE #524: Improve about window focus
**macOS Version:** 10.13.6
**Kap Version:** 3.0.0-beta.1.1006

#### Steps to reproduce
- Open Kap (default recording interface)
- Click the ... icon and then "About Kap Beta"

#### Current behaviour
- The About window is opened
- It is not focused or interact-able because the recording interface takes mouse focus
- Cannot close or interact with the About window until you exit the recording interface

#### Expected behaviour
- Exit the recording interface
- Open the About window
- Basically just use the same behaviour as the Preferences window

#### Screenshot
<p align="center">
screen shot 2018-09-24 at 11 36 20 am
</p>

It feels natural that a newly created window should receive focus and is not hidden behind the recording interface.

karaggeorge:
Fixes #511 
Fixes #524

### DIFF
--- a/main/about.js
+++ b/main/about.js
@@ -1,0 +1,48 @@
+const electron = require('electron');
+const {is} = require('electron-util');
+const ipc = require('electron-better-ipc');
+const loadRoute = require('./utils/routes');
+
+const {getAppIcon} = require('./utils/icon');
+
+const {BrowserWindow} = electron;
+
+let aboutWindow;
+
+const openAboutWindow = () => {
+  if (aboutWindow) {
+    aboutWindow.show();
+    return;
+  }
+
+  aboutWindow = new BrowserWindow({
+    width: 280,
+    height: 200,
+    resizable: false,
+    maximizable: false,
+    minimizable: false,
+    fullscreenable: false,
+    title: '',
+    center: true,
+    webPreferences: {
+      webSecurity: !is.development // Disable webSecurity in dev to load video over file:// protocol while serving over insecure http, this is not needed in production where we use file:// protocol for html serving.
+    },
+    show: false
+  });
+
+  loadRoute(aboutWindow, 'about');
+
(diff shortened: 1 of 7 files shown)


sindresorhus:
Why are we creating a custom about window? Why not just use the native one? Seems like an easier solution to the focus issue would be to just close the cropper view when the about window is opened.

sindresorhus:
The about window should be slightly wider to exactly match the default macOS about window.

sindresorhus:
I think we should also include a `Website` link in there, on the left side of the Policy link.

sindresorhus:
The icon should not be draggable. Use `pointer-event: none` on it.

karaggeorge:
Is there a way to execute code when the about window is opened? It seemed like it was an accelerator so there was no way to close the croppers before opening it.

main/about.js
```diff
@@ -0,0 +1,46 @@
...
+const openAboutWindow = () => {
+  closeAllCroppers();
+
+  if (aboutWindow) {
+    aboutWindow.show();
+    return;
+  }
+
+  aboutWindow = new BrowserWindow({
+    width: 280,
```
sindresorhus: Native about window is `284`

sindresorhus:
I didn't think of that. If it's an accelerator, then no.

sindresorhus:
The about window should not show until it's completely rendered. This is what I see in the built app (not dev):

kap-about 2018-09-27 19_57_57

karaggeorge:
Not sure why that happened. I called `show` on `componentDidUpdate`, but as you said I still noticed the flash. Awaiting `delay(100)` seems to solve it.

PR #546, opened 2018-09-27
Reactivate some XO rules

sindresorhus:


renderer/containers/preferences.js
```diff
@@ -123,6 +123,8 @@ export default class PreferencesContainer extends Container {
   }
 
   toggleSetting = (setting, value) => {
+    // TODO: Fix this ESLint violation
+    // eslint-disable-next-line react/no-access-state-in-setstate
```
karaggeorge: I still don't get why this is bad if we are talking about a "toggle" function. I guess a solution would be to pass in the `!state.value` from where it's called

sindresorhus: @karaggeorge React state can be async, so the state might be changed between `this.state[setting]` is used and `this.setState({[setting]: newVal})` is finished.

albinekb: Can you use `this.setState(state => nextState)`?

sindresorhus: Yes, using the callback version of setState is the correct way to fix this. It ensures the updates are atomic.

karaggeorge: Does unstated support that?

albinekb: Yes @karaggeorge https://github.com/jamiebuilds/unstated#setstate

```setState() in Container mimics React's setState() method as closely as possible.
```

Issue #538, opened 2018-09-27
Exporting to APNG fails in master branch


sindresorhus:
Master meaning https://github.com/wulkano/kap/commit/285a26dac425a35ab759ef76b629584e6a5fb4a6

I just get "Export failed" in the Exports window. No error message in devtools.

When fixing this we should ensure that if it fails, at least it shows a more detailed error message in devtools.

skllcrn:
I realise that you're referring to `master`, but just to be clear I'm not able to reproduce this in the latest beta. Wonder what could be causing it?

karaggeorge:
I just tested this in the current master and it exported APNG successfully. I'll open a PR for showing any export errors in the devtools however, so we can figure out what it was if it's something that doesn't happen often

sindresorhus:
Thanks to #547, here's the error I'm getting:

```
index.js:2178 Error: ffmpeg exited with code: 1

ffmpeg version N-91158-g63c4a4b0d6-tessus Copyright (c) 2000-2018 the FFmpeg developers
  built with Apple LLVM version 9.1.0 (clang-902.0.39.1)
  configuration: --cc=/usr/bin/clang --prefix=/opt/ffmpeg --extra-version=tessus --enable-avisynth --enable-fontconfig --enable-gpl --enable-libaom --enable-libass --enable-libbluray --enable-libfreetype --enable-libgsm --enable-libmodplug --enable-libmp3lame --enable-libmysofa --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopus --enable-librubberband --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libtheora --enable-libtwolame --enable-libvidstab --enable-libvo-amrwbenc --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libx264 --enable-libx265 --enable-libxavs --enable-libxvid --enable-libzimg --enable-libzmq --enable-libzvbi --enable-version3 --pkg-config-flags=--static --disable-ffplay
  libavutil      56. 18.102 / 56. 18.102
  libavcodec     58. 19.104 / 58. 19.104
  libavformat    58. 17.100 / 58. 17.100
  libavdevice    58.  4.100 / 58.  4.100
  libavfilter     7. 24.100 /  7. 24.100
  libswscale      5.  2.100 /  5.  2.100
  libswresample   3.  2.100 /  3.  2.100
  libpostproc    55.  2.100 / 55.  2.100
[...]

PR #547, opened 2018-09-27
Export window updates

### LINKED ISSUE #537: Clicking an item in the Exports window should reveal the file in Finder
That's what I expected to happen. It's weird that it currently has a hover effect, but nothing happens when I click it.

### LINKED ISSUE #543: Add a way to open export window
**Kap version:** 3 Beta 1

#### Current behaviour
Can't open export window after closing it

#### Expected behaviour
I can open up the export window in the menu bar or cropper controls:
image

#### Workaround
Record & export another recording to get the export window to show

karaggeorge:
Fixes #543 
Fixes #537 

Partially addresses #538 by printing export errors in the devtools when an export fails


sindresorhus:
The `Export completed` text should not be selectable.

sindresorhus:
I exported a GIF to disk and when clicking the item in the export window, nothing happens still.

sindresorhus:
Another thing I noticed is that the menu shown when clicking `...` is incorrectly placed based on mouse location and not the `...` location. So https://github.com/wulkano/kap/pull/528#discussion_r221130376 would be useful here too.

karaggeorge:
Hmm, @sindresorhus the click to open was working for me and now it's not. After inspecting and grabbing the filepath, seems like the file doesn't exist any more. Is it possible that because it's in a temp directory it gets cleared at some point?

sindresorhus:
@karaggeorge Temp directories doesn't clear out that fast, but it is moved: https://github.com/wulkano/kap/blob/8caadd8e3abdd1dcae1d51a72a2029f182c1ee87/main/save-file-service.js#L14

karaggeorge:
😮 I don't know how I missed that. That's why it was working before. I was using a different export method. I can change that so it copies instead 👍 

Good catch!

karaggeorge:
@sindresorhus This should be ready and now fixes #538 as well. The only thing left out is the `...` menu not opening right which I'll fix with the reusable select in the other PR

sindresorhus:
>  I don't know how I missed that. That's why it was working before. I was using a different export method. I can change that so it copies instead 👍

I didn't mean that you should change the moving. The point of the "revealing in Finder" is to show it where it ended up. For "Save to Disk" this should use `targetFilePath`, for other services, we can just reveal it in the temp directory.

sindresorhus:
👌

PR #570, opened 2018-10-04
fix #556 set initial width and height of cropper to 400px

### LINKED ISSUE #556: Use a better initial size when setting the aspect ratio first
If I select an aspect ratio after starting Kap, it will default the cropper to the minimum possible size with that aspect ratio. I think we should have a better default. No one will want to record that small. I think a good default would be 400px width and whatever the aspect ratio makes the height.

masquel:


renderer/containers/cropper.js
```diff
@@ -77,8 +77,8 @@ export default class CropperContainer extends Container {
       displayId: id,
       x: x || screenWidth / 2,
       y: y || screenHeight / 2,
-      width: width || 0,
-      height: height || 0,
+      width: width || 400,
+      height: height || 400,
       ratio: ratio || [1, 1]
```
skllcrn: I suggest setting the initial/default aspect ratio to 4:3, that aspect ratio works better across devices and services you're likely to upload to. 

```
width: width 400,
height: height 300,
ratio: ratio [4, 3]
```

sindresorhus:
I don't think I was clear enough in #556. I meant we should set a larger initial size of the cropper **after** choosing an aspect ratio in the dropdown, not should show the cropper rectangle by default on startup.

sindresorhus:
@masquel Still interested in finishing this? :)

masquel:
@sindresorhus hello, so what do you want now for the initial size?

PR #571, opened 2018-10-04
Fix the audio toggle button in the editor for unmute cases

### LINKED ISSUE #561: The audio icon should be more obvious when it's not clickable
When I have audio recording disabled in the preferences, record something, and in the editor the audio (mute) icon shows muted like normal, but nothing happens when I click it. I had forgotten that I had audio recording disabled and thought it was a bug. We should make it look like a disabled control when it's not possible to change it.

Right now it looks exactly like the play/pause icon which is clickable:

screen shot 2018-10-04 at 12 50 31

wcastand:
Fixes #561 

I used a color i found in the code, i'm not sure that how you wanted the disabled button to look like. i think we should improve also the hover state for a clickable button because right now nothing happen when you hover them, not even the hand so might be cool to have a change of color and the hand for example. Maybe in an other PR x)

Also, in the contributing.md you wrote to test the binary you need to run `yarn pack`. For me, yarn pack just created a tarball with no dist and no app.

So i run `yarn build`and `yarn dist` to get the executable.

### DIFF
--- a/renderer/components/editor/controls/right.js
+++ b/renderer/components/editor/controls/right.js
@@ -17,17 +17,16 @@ class RightControls extends React.Component {
   }
 
   render() {
-    const {isMuted, mute, unmute, format, duration} = this.props;
+    const {isMuted, mute, unmute, format, duration, hasAudio} = this.props;
     const {recordAudio} = this.state;
     const canUnmute = !['gif', 'apng'].includes(format);
     const unmuteColor = canUnmute && recordAudio ? '#fff' : 'rgba(255, 255, 255, 0.40)';
-
     return (
       <div className="container">
         <div className="time">{formatTime(duration)}</div>
         <div className="mute">
           {
-            isMuted || !recordAudio ?
+            isMuted || !recordAudio || !hasAudio ?
               <VolumeOffIcon shadow fill={unmuteColor} hoverFill={unmuteColor} onClick={canUnmute && recordAudio ? unmute : undefined}/> :
               <VolumeHighIcon shadow fill="#fff" hoverFill="#fff" onClick={mute}/>
           }
@@ -61,6 +60,7 @@ class RightControls extends React.Component {
 }
 
 RightControls.propTypes = {
+  hasAudio: PropTypes.bool,
   isMuted: PropTypes.bool,
   mute: PropTypes.func,
   unmute: PropTypes.func,
@@ -70,6 +70,6 @@ RightControls.propTypes = {
 
 export default connect(
   [VideoContainer, EditorContainer],
-  ({isMuted, duration}, {format}) => ({isMuted, format, duration}),
+  ({isMuted, duration, hasAudio}, {format}) => ({hasAudio, isMuted, format, duration}),
   ({mute, unmute}) => ({mute, unmute})
 )(RightControls);
--- a/renderer/components/editor/controls/right.js
+++ b/renderer/components/editor/controls/right.js
@@ -10,7 +10,7 @@ import formatTime from '../../../utils/format-time';
 class RightControls extends React.Component {
   state = {}
 
-  componentDidMount(){
+  componentDidMount() {
     const settings = electron.remote.require('./common/settings');
     const recordAudio = settings.get('recordAudio');
     this.setState({recordAudio});
--- a/renderer/components/editor/controls/right.js
+++ b/renderer/components/editor/controls/right.js
@@ -10,6 +10,7 @@ class RightControls extends React.Component {
   render() {
     const {isMuted, mute, unmute, format, duration} = this.props;
     const canUnmute = !['gif', 'apng'].includes(format);
+    const unmuteColor = canUnmute ? '#fff' : 'rgba(255, 255, 255, 0.25)';
 
     return (
       <div className="container">
@@ -17,7 +18,7 @@ class RightControls extends React.Component {
         <div className="mute">
           {
             isMuted ?
-              <VolumeOffIcon shadow fill="#fff" hoverFill="#fff" onClick={canUnmute ? unmute : undefined}/> :
+              <VolumeOffIcon shadow fill={unmuteColor} hoverFill={unmuteColor} onClick={canUnmute ? unmute : undefined}/> :
               <VolumeHighIcon shadow fill="#fff" hoverFill="#fff" onClick={mute}/>
           }
         </div>
--- a/renderer/components/editor/controls/right.js
+++ b/renderer/components/editor/controls/right.js
@@ -1,4 +1,3 @@
-import electron from 'electron';
 import PropTypes from 'prop-types';
 import React from 'react';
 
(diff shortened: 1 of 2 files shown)


skllcrn:
Good catch @alonalon, we could completely hide the icon if audio is disabled?

wcastand:
oh good catch, i think we can show the disabled button too. So people know it's not unavailable and not a "bug" in the interface. what do you think ?

skllcrn:
Sounds good to me @wcastand, I might also make the icon slightly more opaque to make sure it's sufficiently visible on both solid black and white backgrounds. Maybe `0.40`?

image

renderer/components/editor/controls/right.js
```diff
@@ -7,18 +8,27 @@ import {connect, VideoContainer, EditorContainer} from '../../../containers';
 import formatTime from '../../../utils/format-time';
 
 class RightControls extends React.Component {
+  state = {}
+
+  componentDidMount() {
+    const settings = electron.remote.require('./common/settings');
```
albinekb: @karaggeorge Do we have the information about if this recording was recorded with audio or not here? Since loading the setting when the controls mount, you could have changed the settings between doing the recording and this happening, for example when opening up an existing file or an old recording via the export UI.

wcastand: I have no idea, first time contributing to this repo :)

The editor is mounted once the recording is done so i'm not sure you can change the settings between the recording and the moment the editor is mounted

wcastand: Then we need the video player to give us that settings right ? But how do i pass the video settings from tag to the control component, from what i can see the video component and controls are at the same depth in video player and Kap has no global store from what i could see.

I'm not sure i can have access to the video file in the controls to check is audio is available or not. If this is possible, let me know and i will update the PR :)

albinekb: I will have a look tonight how it could be done and give you some pointers :)

karaggeorge: I would look into something like this: 
https://stackoverflow.com/questions/21446804/find-if-video-file-has-audio-present-in-it

I used ffmpeg to find the fps of a video (which is set to the max exported fps). If we could do something like that at the same spot (I think there's a util file and then it's used in the editor file in the main process) that would probably cover all the cases

skllcrn: I saw you might be interested in helping out with this issue @hooskers, feel free to chime in!

sindresorhus:
@wcastand Still interested in finishing this? :)

wcastand:
Yeah, why not, sorry i'm moving to london for a job so a bit busy and didn't find a way to fix the last issue, but if anyone has a clue, i'll be happy to finish :)

renderer/components/editor/controls/right.js
sindresorhus: According to https://stackoverflow.com/a/31022391/64949, we can use this to detect whether the video has audio:

```js
const hasAudio = video => {
	return video.webkitAudioDecodedByteCount > 0 ||
	Boolean(video.audioTracks && video.audioTracks.length > 0);
};
```

wcastand:
I looked again and it seems i was right in my last msg, the issue is not to find out if the video has audio or not but to access the video object in the editor/controls/right where we need it.

Anyone know a way to access the video object from this file or a way to make it accessible ? Once we found this, finishing this PR should be okay since #571 is probably the right solution once we have the video object.

wcastand:
I added hasAudio in VideoContainer state and used it in the controls player. It seemed to worked as intended for me. someone might need to double check to be sure :) but i think we are good now.

renderer/components/editor/controls/right.js
```diff
@@ -7,17 +8,26 @@ import {connect, VideoContainer, EditorContainer} from '../../../containers';
...
+    const recordAudio = settings.get('recordAudio');
+    this.setState({recordAudio});
```
karaggeorge: We shouldn't need the state, recordAudio, or electron at all now. The hasAudio coming from the container should be enough

renderer/components/editor/controls/right.js
```diff
@@ -7,17 +8,26 @@ import {connect, VideoContainer, EditorContainer} from '../../../containers';
...
+  }
+
   render() {
-    const {isMuted, mute, unmute, format, duration} = this.props;
+    const {isMuted, mute, unmute, format, duration, hasAudio} = this.props;
+    const {recordAudio} = this.state;
     const canUnmute = !['gif', 'apng'].includes(format);
-
+    const unmuteColor = canUnmute && recordAudio ? '#fff' : 'rgba(255, 255, 255, 0.40)';
```
karaggeorge:
```suggestion
    const unmuteColor = canUnmute && hasAudio ? '#fff' : 'rgba(255, 255, 255, 0.40)';
```

renderer/components/editor/controls/right.js
```diff
@@ -7,17 +8,26 @@ import {connect, VideoContainer, EditorContainer} from '../../../containers';
...
     return (
       <div className="container">
         <div className="time">{formatTime(duration)}</div>
         <div className="mute">
           {
-            isMuted ?
-              <VolumeOffIcon shadow fill="#fff" hoverFill="#fff" onClick={canUnmute ? unmute : undefined}/> :
+            isMuted || !recordAudio || !hasAudio ?
+              <VolumeOffIcon shadow fill={unmuteColor} hoverFill={unmuteColor} onClick={canUnmute && recordAudio ? unmute : undefined}/> :
```
karaggeorge:
```suggestion
              <VolumeOffIcon shadow fill={unmuteColor} hoverFill={unmuteColor} onClick={canUnmute && hasAudio ? unmute : undefined}/> :
```

renderer/containers/video.js
```diff
@@ -57,7 +58,10 @@ export default class VideoContainer extends Container {
     video.addEventListener('loadedmetadata', () => {
       const {videoWidth, videoHeight, duration} = video;
       this.editorContainer.setDimensions(videoWidth, videoHeight);
-      this.setState({duration, startTime: 0, endTime: duration});
+      const hasAudio = video.webkitAudioDecodedByteCount > 0 ||
+        Boolean(video.audioTracks && video.audioTracks.length > 0);
+      console.log(hasAudio);
```
karaggeorge:
```suggestion
```

renderer/components/editor/controls/right.js
```diff
@@ -7,17 +8,26 @@ import {connect, VideoContainer, EditorContainer} from '../../../containers';
...
+  componentDidMount() {
+    const settings = electron.remote.require('./common/settings');
+    const recordAudio = settings.get('recordAudio');
+    this.setState({recordAudio});
+  }
+
   render() {
-    const {isMuted, mute, unmute, format, duration} = this.props;
+    const {isMuted, mute, unmute, format, duration, hasAudio} = this.props;
+    const {recordAudio} = this.state;
```
karaggeorge:
```suggestion
```

renderer/components/editor/controls/right.js
```diff
@@ -7,17 +8,26 @@ import {connect, VideoContainer, EditorContainer} from '../../../containers';
...
-
+    const unmuteColor = canUnmute && recordAudio ? '#fff' : 'rgba(255, 255, 255, 0.40)';
     return (
       <div className="container">
         <div className="time">{formatTime(duration)}</div>
         <div className="mute">
           {
-            isMuted ?
-              <VolumeOffIcon shadow fill="#fff" hoverFill="#fff" onClick={canUnmute ? unmute : undefined}/> :
+            isMuted || !recordAudio || !hasAudio ?
```
karaggeorge:
```suggestion
            isMuted || !hasAudio ?
```

wcastand:
Any updates ?
Do i need to do something more ? :)

karaggeorge:
Hey @wcastand, this looks much cleaner. Can you just remove that `console.log` from the container code? It's the last review comment left unresolved: https://github.com/wulkano/kap/pull/571#discussion_r252732086

wcastand:
@karaggeorge done, sorry i forgot this one. we should be okay now :)

sindresorhus:
@wcastand Thanks for contributing :)

Issue #565, opened 2018-10-04
The title of the editor window is not centered


sindresorhus:
It should be centered in the window, ignoring the traffic lights.

<img width="880" alt="screen shot 2018-10-04 at 13 10 32" src="https://user-images.githubusercontent.com/170270/46455849-f90b5b80-c7d6-11e8-98c9-9774d59816fc.png">

PR #573, opened 2018-10-05
Issue #565: Title not centered

joviij:
Add margin-left: -72px to make title ignore the size of traffic lights plus the margins associated with the traffic light

The total size of traffic light (image + margins) is 72px thus the -72px.

Another option is to make the title have position: absolute and give it a negative z-index to make sure it doesn't mess with the traffic lights.

The final result:
screen shot 2018-10-04 at 5 44 03 pm

### DIFF
--- a/renderer/components/editor/index.js
+++ b/renderer/components/editor/index.js
@@ -66,6 +66,7 @@ export default class Editor extends React.Component {
             justify-content: center;
             font-size: 1.4rem;
             color: #fff;
+            margin-left: -72px;
           }
         `}</style>
       </div>


renderer/components/editor/index.js
```diff
@@ -66,6 +66,7 @@ export default class Editor extends React.Component {
             justify-content: center;
             font-size: 1.4rem;
             color: #fff;
+            margin-left: -72px;
```
skllcrn: This is fine for now, setting it to `absolute` is definitively an option, but we can look into that when we make changes to the traffic lights or titlebar in the future. Is this consistent with how we're doing this in other windows that have custom traffic lights and titlebars? Thank you!

skllcrn: That makes sense, thank you @joviij. I was just double checking, we should *only* be using the custom traffic lights for this window.

sindresorhus:
@joviij Tip: Next time, write `Fixes #565` (switch out the issue number, of course) in the pull request description and the issue will be automatically closed when the pull request is merged ;)

PR #577, opened 2018-10-06
Standardizes the size of SVG icons

### LINKED ISSUE #564: The cropper control icons looks like they have different sizes
Would be good to visually unify them.

screen shot 2018-10-04 at 13 03 40

Slack discussion:

screen shot 2018-10-04 at 13 04 11

pauloedurezende:
Fix #564

### DIFF
--- a/renderer/components/action-bar/controls/main.js
+++ b/renderer/components/action-bar/controls/main.js
@@ -52,7 +52,7 @@ class Left extends React.Component {
     return (
       <div className="main">
         <div className="crop">
-          <CropIcon viewBox="-1.75 -1.75 28 28" onClick={toggleAdvanced}/>
+          <CropIcon onClick={toggleAdvanced}/>
         </div>
         <IconMenu onOpen={menu && menu.popup}><ApplicationsIcon active={Boolean(selectedApp)}/></IconMenu>
         <style jsx>{mainStyle}</style>
@@ -93,11 +93,11 @@ class Right extends React.Component {
         <div className="fullscreen">
           {
             isFullscreen ?
-              <ExitFullscreenIcon active viewBox="2 2 20 20" onClick={exitFullscreen}/> :
-              <FullscreenIcon viewBox="2 2 20 20" onClick={enterFullscreen}/>
+              <ExitFullscreenIcon active onClick={exitFullscreen}/> :
+              <FullscreenIcon onClick={enterFullscreen}/>
           }
         </div>
-        <IconMenu onOpen={electron.remote.require('./menus').cogMenu.popup}><MoreIcon viewBox="2 2 20 20"/></IconMenu>
+        <IconMenu onOpen={electron.remote.require('./menus').cogMenu.popup}><MoreIcon/></IconMenu>
         <style jsx>{mainStyle}</style>
         <style jsx>{`
           .fullscreen {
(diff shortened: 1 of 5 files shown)


skllcrn:
Thank you @pauloedurezende, and congrats on your first! Does this edit the actual SVG's/icons themselves, or just the bounds? We've been very careful with using customised icons as it can quickly lead us down the path of inconsistency. If this does tweak the actual paths we should make sure to contain it to the toolbar and leave sufficient context. Could you post a Kapture which also shows the hover and active states?

sindresorhus:
I think it would be better to set a correct `viewBox` attribute on all those icons. Then we can just use the width/height attributes.

pauloedurezende:
> Thank you @pauloedurezende, and congrats on your first! Does this edit the actual SVG's/icons themselves, or just the bounds? We've been very careful with using customised icons as it can quickly lead us down the path of inconsistency. If this does tweak the actual paths we should make sure to contain it to the toolbar and leave sufficient context. Could you post a Kapture which also shows the hover and active states?

Thank you! I had only changed the sizes of the icons in SVG, I left them all with a fixed size (18px if I'm not mistaken).

> I think it would be better to set a correct `viewBox` attribute on all those icons. Then we can just use the width/height attributes.

I agree that as I had done, it was not the best way, I returned the icons the way they were and added the `viewBox` as you mentioned.

It ended up having the same aesthetic result as the GIFs below...

Using the edited SVG icons:
edited svg

Using `viewBox`:
viewbox

sindresorhus:
@pauloedurezende The `viewBox` attribute should be added to the components themselves, not where they are used.

sindresorhus:
@pauloedurezende Still interested in finishing this? :)

pauloedurezende:
@sindresorhus  Yes I am, sorry for the delay...

But one thing I can not understand, the SVG components, such as `crop.js`, it gets the props and goes to the `svg.js` component. In this component, it has its default properties, where it ends up passing to `<svg></svg>`. So, it would not be more logical to end up using the `crop.js` component, for example, this way `<CropIcon viewBox="-1.75 -1.75 28 28" onClick={toggleAdvanced} />`, instead of having to change the component itself, to stay this way `<Svg {...props} viewBox="-1.75 -1.75 28 28">[...]</Svg>`?

sindresorhus:
@pauloedurezende I don't fully understand your question, but the `viewBox` should be added to the icon components themselves as it's internal data that you should not need to know outside the specific icon component. I would recommend reading up on the `viewBox` attribute.

sindresorhus:
It looks better now than what's in master branch, but I still feel the sizes could be improved. Not sure yet what bothers me.

Here's how it looks now:

screen shot 2018-12-18 at 15 32 06

@skllcrn Thoughts?

skllcrn:
In that last screenshot both the cropper and fullscreen icon areblurry at @1x and not particularly well balanced in terms of visual weight. They should stick to the standard icon size and sit inside 24x24 bounds. We can [try using another icon](https://github.com/google/material-design-icons/blob/master/image/svg/production/ic_crop_free_24px.svg) for the fullscreen option if that's the one that's bothering you.

sindresorhus:
Closing as this doesn't correctly fix the issue and we're not really sure in which way to fix it ourselves. We still need to do something about the icons, but I'm not sure what...

PR #578, opened 2018-10-06
Open cropper window at app launch

### LINKED ISSUE #544: Open cropper view at app launch
I think it would be nice if Kap would show the cropper view at launch. I only have Kap open when I actually want to use it and this would make it quicker for me to start recording.

It should not however be opened if Kap was launched at system startup. We can detect this with `app.getLoginItemSettings`, which returns an object with `wasOpenedAtLogin`. 

https://electronjs.org/docs/api/app#appgetloginitemsettingsoptions-macos-windows

theshortcut:
Adds a setting (defaulted to true) to open the cropper window on launch.

Only applies if kap is not opened at login and if it is not opening files at launch.

Fixes #544

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -11,6 +11,7 @@ const {initializeTray} = require('./tray');
 const plugins = require('./common/plugins');
 const {initializeAnalytics} = require('./common/analytics');
 const initializeExportList = require('./export-list');
+const {openCropperWindow} = require('./cropper');
 const {openEditorWindow} = require('./editor');
 const {track} = require('./common/analytics');
 const {initializeGlobalAccelerators} = require('./global-accelerators');
@@ -74,6 +75,12 @@ const checkForUpdates = () => {
   for (const file of filesToOpen) {
     track('editor/opened/startup');
     openEditorWindow(file);
+  }
+
+  const {wasOpenedAtLogin} = app.getLoginItemSettings();
+  const isOpeningFile = filesToOpen.length > 0;
+  if (!isOpeningFile && !wasOpenedAtLogin) {
+    openCropperWindow();
   }
 
   checkForUpdates();


main/common/settings.js
```diff
@@ -24,7 +24,8 @@ const store = new Store({
       ctrlKey: false,
       shiftKey: true,
       character: '5'
-    }
+    },
+    showCropperOnLaunch: true
```
sindresorhus: Nitpick, but I would call it `showCropperAtLaunch`

renderer/components/preferences/categories/settings.js
```diff
@@ -146,6 +147,9 @@ class Settings extends React.Component {
         <Item title="Save to…" subtitle={kapturesDirPath} tooltip={kapturesDir} onSubtitleClick={this.openKapturesDir}>
           <Button title="Choose" onClick={pickKapturesDir}/>
         </Item>
+        <Item subtitle="Show cropper on launch">
```
sindresorhus: `on` => `at`

sindresorhus: You need a title too. We only do subtitle-only items when they are subitems of the above item.

albinekb: Or just change it to title without subtitle.

theshortcut: 👍 It could probably use a subtitle explaining when it wouldn’t open despite the setting.

theshortcut:
Updated:

image

Would appreciate any better wording here :)

theshortcut:
@sindresorhus any additional changes you'd like to see here?

renderer/components/preferences/categories/settings.js
```diff
@@ -146,6 +147,9 @@ class Settings extends React.Component {
         <Item title="Save to…" subtitle={kapturesDirPath} tooltip={kapturesDir} onSubtitleClick={this.openKapturesDir}>
           <Button title="Choose" onClick={pickKapturesDir}/>
         </Item>
+        <Item title="Show cropper at launch" subtitle="Only when not launched at system startup or launching editor">
```
sindresorhus: `launching editor` => `launching the editor`

sindresorhus:
Sorry about the delay in feedback. We had many back and forth discussions and decided that we would like to try to make this behavior the default without any setting. It's easy to add settings, but hard to remove them. Kap has ended up with a lot of settings now, so we'd like to be more strict about what we add. I'm sorry that you wasted your time on that part, but I think it's for the better. Would you be able to update the PR to not include the setting and instead just make it the default behavior?

theshortcut:
@sindresorhus no worries. PR updated to remove the setting.

sindresorhus:
Looks good now. Thank you! 🙌

PR #579, opened 2018-10-07
Position menu above IconMenu near screen bottom

### LINKED ISSUE #563: Incorrect menu position for buttons with popup menus
If I move the cropper controls view to the bottom of the screen and then click the `...` button, the menu opens over the button instead of above it as expected.

screen shot 2018-10-04 at 13 00 53

This happens with the window selector button too, and probably anything that uses the `IconMenu` component.

theshortcut:
Attempts to fix #563 but since electron provides no control or feedback for whether the menu will display above or below this approach isn't perfect.

It works just fine when all the way at the bottom of the screen:
image

The issue is that if we check the position compared to the bottom of the screen in an attempt to guess if the menu will display upwards, then we move the menu attachment up to the top of the bounding rect of the IconMenu, we've now given the menu more space to display downwards and it may decide to use that and display downwards still, once again covering the IconMenu component.

Example:
image

The threshold I've set this to attempts to make the amount of space this could possible happen in either direction as small as possible, but it is still possible.

### DIFF
--- a/renderer/components/action-bar/controls/main.js
+++ b/renderer/components/action-bar/controls/main.js
@@ -54,7 +54,7 @@ class Left extends React.Component {
         <div className="crop">
           <CropIcon onClick={toggleAdvanced}/>
         </div>
-        <IconMenu onOpen={menu && menu.popup}><ApplicationsIcon active={Boolean(selectedApp)}/></IconMenu>
+        <IconMenu itemCount={menu && menu.items.length} onOpen={menu && menu.popup}><ApplicationsIcon active={Boolean(selectedApp)}/></IconMenu>
         <style jsx>{mainStyle}</style>
         <style jsx>{`
           .crop {
@@ -87,6 +87,7 @@ MainControls.Left = connect(
 class Right extends React.Component {
   render() {
     const {enterFullscreen, exitFullscreen, isFullscreen} = this.props;
+    const {cogMenu} = electron.remote.require('./menus');
 
     return (
       <div className="main">
@@ -97,7 +98,7 @@ class Right extends React.Component {
               <FullscreenIcon onClick={enterFullscreen}/>
           }
         </div>
-        <IconMenu onOpen={electron.remote.require('./menus').cogMenu.popup}><MoreIcon/></IconMenu>
+        <IconMenu itemCount={cogMenu.items.length} onOpen={cogMenu.popup}><MoreIcon/></IconMenu>
         <style jsx>{mainStyle}</style>
         <style jsx>{`
           .fullscreen {
--- a/renderer/components/icon-menu.js
+++ b/renderer/components/icon-menu.js
@@ -1,19 +1,26 @@
+import electron from 'electron';
 import React from 'react';
 import PropTypes from 'prop-types';
 
+const MENU_UP_THRESHOLD = 60;
+
 class IconMenu extends React.Component {
+  remote = electron.remote || false
+
   container = React.createRef();
 
   openMenu = event => {
     const boundingRect = this.container.current.getBoundingClientRect();
-    const {bottom, left} = boundingRect;
+    const {screen} = this.remote;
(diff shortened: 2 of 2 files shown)


sindresorhus:
If this component knew how many menu items there were, it could maybe calculate the threshold more accurately based on that information.

sindresorhus:
I think we should open an Electron issue about being able to specify the direction the menu should open. You can do that natively.

theshortcut:
@sindresorhus looking through Apple's documentation for any details on how this would be handled on the electron side, 

https://developer.apple.com/documentation/appkit/nsmenu/1518212-popup

I don't see any way to control or introspect the behavior there, are you familiar with this? If opening an issue upstream I'd like to provide some more information. 

Additionally I'm not sure the OS actually takes the number of menu items into consideration when displaying up or down either since the `…` popover and the windows list popover both decide to open upwards at the same position on screen despite having a very different number of menu items.

I believe this is electron's handling of MacOS popup menus here:
https://github.com/electron/electron/blob/master/atom/browser/api/atom_api_menu_mac.mm#L91

They do some custom horizontal positioning but vertical direction looks to be OS determined.

theshortcut:
Updated based on feedback from the electron folks. Exposes the item count to the `IconMenu` component and when opening upwards based on our threshold we now force positioning using the `positioningItem` argument to `Menu.popUp`. This allows us to force the menu to open upwards based on our own logic rather than leaving that to the OS.

sindresorhus:
FYI, this is waiting for response from Electron team here: https://github.com/electron/electron/issues/15096#issuecomment-429959619

sindresorhus:
I've opened a duplicate https://github.com/electron/electron/issues/16008 as the Electron team ignored my follow-up comment.

sindresorhus:
Not sure if it's Electron's fault or a bug in the calculation, but the menu goes down here, while it would make more sense for it to go upwards:

screen shot 2018-12-21 at 01 38 51

skllcrn:
On my display where I've set the dock to display at all times the menu appears exactly like @sindresorhus showed in his comment, but probably for a different reason; it doesn't flow over the "safe zone" around the dock. I'm guessing that's something the calculation would need to account for?

skllcrn:
What are your thoughts on ways to wrap this one up @theshortcut and @sindresorhus?

theshortcut:
I was hoping we would see some motion on https://github.com/electron/electron/issues/16008 but since we haven’t yet I think the options are:

1. Move ahead with the original lower threshold option (not the current code in this PR)
2. Stick with the selected index approach (current PR) with a little tweaking to the threshold 
3. Close this out and revisit when the upstream issue is resolved

karaggeorge:
It looks like the logic between opening down or up isn't if it fits all the items, but if it fits a set amount (I think it's about 2 items). Otherwise it shows part of the menu with an arrow. We could (by manually testing) figure out what that size is, and use that as a static threshold for when to show it above or below.

Further thinking about it, there will always be a point where we would think that the items can't fit, we would say open above, but opening above would give it more space and it would open below, even though we are showing it over the icon. So no matter what threshold we choose, there'll always be a range of pixels that will make the menu show over the icon.

Not sure what electron could offer either. Even if they tell us if it opened up or down, at that point it has already opened, so we have already positioned it.

sindresorhus:
> 1. Move ahead with the original lower threshold option (not the current code in this PR)

I would go with this. I don't see it being fixed in Electron anytime soon.

> Not sure what electron could offer either. Even if they tell us if it opened up or down, at that point it has already opened, so we have already positioned it.

The Electron issue is about getting ability to choose where it opens before it opens and also the menu size.

skllcrn:
What do you think @theshortcut? As @karaggeorge pointed out there are a lot of cases to keep in mind.

sindresorhus:
Closing this for lack of activity.

PR #584, opened 2018-10-10
Add graphite appearance to traffic lights

theshortcut:
Kap feels out of place when using graphite appearance mode.

This updates the TrafficLights to read from the user's appearance preferences and updates their appearance to match the rest of the system more closely.

kapture 2018-10-10 at 16 53 09


sindresorhus:
This is great. Thank you, @theshortcut :)

PR #585, opened 2018-10-14
Close export window if not open already

### LINKED ISSUE #560: Clicking "Export" button in the editor and then "Cancel" should close export window
...unless the "Export" window was already open.

Say I click "Export", then realize I want some different settings, there's no point in keeping the "Export" window open if it was not already open. It just forces me to click extra to close it away.

joviij:
Fixes #560 

Issue: When we click the "Export" button in the Editor, and then hit "Cancel," it should close the export window if it was not open already.

It seemed that in the code we never explicitly close the export window but rather we hide it. The way I went about it was to check if the export window was hidden when the "Export" button was clicked and if it was hidden (indicating that it was not open previously or closed previously), then on "Cancel" click, it would close the exports window.

**If exports window was not open previously**
cancelclick_notopen

**If exports window was open previously**
cancel_open

### DIFF
--- a/main/export-list.js
+++ b/main/export-list.js
@@ -101,7 +101,7 @@ class ExportList {
   }
 
   async addExport(options) {
-    const exportWindowVisible = getExportsWindow().isVisible(); // Check if export window is visible
+    const isExportsWindowVisible = getExportsWindow().isVisible();
     toggleExportMenuItem(true);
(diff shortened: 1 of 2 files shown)


sindresorhus:
Looks great. Thanks, @joviij 👍

PR #587, opened 2018-10-21
Add keyboard control to editor and refactor into shared component

### LINKED ISSUE #562: The number inputs in the editor should be changeable with arrow keys
In the cropper view, you can change the number inputs with arrow up/down keys, and you can also press <kbd>Shift</kbd> to increase/decrease by 10. Would be nice if this worked in the editor window too, for the width, height, and FPS inputs. Should probably make it a reusable component.

bringking:
This PR closes #562 and does the following - 

* Creates a shared component `KeyboardNumberInput` which uses the `onKeyDown` event to change the input value using the arrow keys. This uses the existing `handleInputKeyPress` with some tweaking. The component takes `min, max and onKeyDown` as props
* Updates `handleInputKeyPress` to accept `min, max` as parameters to make it less opinionated and usable in both the cropper and the editor
* Sets the `min,max` values in the cropper to be the `minHeight` and `minWidth` values in the `utils/inputs.js` and the screen height/width
* Sets the `mix,max` values in the editor to be the `minHeight` and `minWidth` values in the `utils/inputs.js` and the original height/width

### DIFF
--- a/renderer/components/editor/options/left.js
+++ b/renderer/components/editor/options/left.js
@@ -1,8 +1,38 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-
+import css from 'styled-jsx/css';
 import {connect, EditorContainer} from '../../../containers';
+import {minHeight, minWidth} from '../../../utils/inputs';
+import KeyboardNumberInput from '../../keyboard-number-input';
 import Slider from './slider';
+
+const {className: keyboardInputClass, styles: keyboardInputStyles} = css.resolve`
+  height: 24px;
+  background: hsla(0, 0%, 100%, 0.1);
+  text-align: center;
+  font-size: 12px;
+  box-sizing: border-box;
+  border: none;
+  padding: 4px;
+  border-bottom-left-radius: 4px;
+  border-top-left-radius: 4px;
+  width: 48px;
+  color: white;
+
+  input + input {
+    border-bottom-left-radius: 0;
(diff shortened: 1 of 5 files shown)


renderer/components/editor/options/left.js
```diff
@@ -56,7 +61,7 @@ class LeftOptions extends React.Component {
             width: 32px;
           }
 
-          input {
+          :global(input) {
```
sindresorhus: Are there no better ways to achieve this than using `:global()`?

bringking: @sindresorhus, yeah it feels wrong to me, but I couldn't find a better way in the docs. I will dig in again to be sure there isn't a better way.

sindresorhus: @bringking Did you find anything better? Otherwise, I guess we can say it's good enough for now.

bringking: @sindresorhus, I didn't dig in yet, sorry about that. I can carve some time for this tomorrow

bringking: @sindresorhus I found the relevant docs for styled/jsx and got those globals cleaned up!

sindresorhus: Awesome!

sindresorhus:
Nice work and high quality PR, @bringking 👌✨ Thank you :)

PR #588, opened 2018-10-30
Add confirm prompt when closing the editor window

### LINKED ISSUE #566: Quitting Kap while the editor window is open should ask the user first
If the user presses Cmd+Q, Cmd+W, red traffic light, or "Quit" in the top menu, the user should be prompted whether they want to quit and what will happen if they choose to quit. This is a regression for Kap 2 where we had this implemented. Refer to the 2.0 source for the text and behavior of the prompt.

alonalon:
Fixes #566

tried to solve the task without change the events behavior and also used same dialog text as in kap 2.0 :)

### DIFF
--- a/main/editor.js
+++ b/main/editor.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const {BrowserWindow, app} = require('electron');
+const {BrowserWindow, app, dialog} = require('electron');
 const ipc = require('electron-better-ipc');
 const {is} = require('electron-util');
 
@@ -9,7 +9,8 @@ const loadRoute = require('./utils/routes');
 
 const editors = new Map();
 let exportOptions;
-
+let editorWindow;
+let forceClose = false;
 const OPTIONS_BAR_HEIGHT = 48;
 const VIDEO_ASPECT = 9 / 16;
 const MIN_VIDEO_WIDTH = 768;
@@ -23,7 +24,7 @@ const openEditorWindow = async (filePath, recordFps) => {
   }
   const fps = recordFps || await getFps(filePath);
 
-  const editorWindow = new BrowserWindow({
+  editorWindow = new BrowserWindow({
     minWidth: MIN_VIDEO_WIDTH,
     minHeight: MIN_WINDOW_HEIGHT,
     width: MIN_VIDEO_WIDTH,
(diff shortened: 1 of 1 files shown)


main/editor.js
```diff
@@ -41,7 +42,30 @@ const openEditorWindow = async (filePath, recordFps) => {
...
+
+    event.preventDefault();
+
+    dialog.showMessageBox(editorWindow, {
+      type: 'question',
+      buttons: ['Discard', 'Cancel'],
+      defaultId: 1,
+      message: 'Are you sure that you want to discard this recording?',
+      detail: 'You will no longer be able to edit and export the original recording.'
+    }, buttonIndex => {
```
sindresorhus: Don't use the callback. We actually want this dialog to block the app.

main/editor.js
```diff
@@ -41,7 +42,30 @@ const openEditorWindow = async (filePath, recordFps) => {
...
+  editorWindow.on('close', event => {
+    if (!editorWindow || forceClose) {
+      return;
+    }
+
+    event.preventDefault();
+
+    dialog.showMessageBox(editorWindow, {
+      type: 'question',
+      buttons: ['Discard', 'Cancel'],
```
sindresorhus: Put the array items on separate lines.

main/editor.js
```diff
@@ -41,7 +42,30 @@ const openEditorWindow = async (filePath, recordFps) => {
...
+      defaultId: 1,
+      message: 'Are you sure that you want to discard this recording?',
+      detail: 'You will no longer be able to edit and export the original recording.'
+    }, buttonIndex => {
+      if (buttonIndex === 0) {
+        forceClose = true;
```
sindresorhus: Instead of the `forceClose` thing, you can just unsubscribe to the `close` event here.

alonalon:
Did some fixes does it looking better now? :)

main/editor.js
```diff
@@ -41,7 +66,9 @@ const openEditorWindow = async (filePath, recordFps) => {
 
   loadRoute(editorWindow, 'editor');
 
+  editorWindow.on('close', discardVideo);
   editorWindow.on('closed', () => {
+    editorWindow = null;
```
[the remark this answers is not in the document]
alonalon: Good catch, Fix is coming!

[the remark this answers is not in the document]
kevva: Yeah, since it'll be different functions when adding/removing the event listener I guess it's ok. I don't know a way to remove the event listener inside a function that's called with arguments without doing some dirty hacks.

sindresorhus:
Nice work, @alonalon 🎉

Issue #136, opened 2017-01-31
UX Improvement: Update control panel tabindex


onordbo:
**As a keyboard shortcut power user**, I would love to be able to jump between input fields `width` and `height` with a single `TAB` press.

Currently, the `tabindex` is completely linear, so there's **three** tab presses to get from `width` to `height`, although the two steps in between doesn't have any actual interactions.

_e.g: There's no highlighted state for the two buttons in between `width` and `height`, and I can't trigger either button with `Enter`._

skllcrn:
A thousand times this.

alokpant:
Is this still available? Can I look into it? @skllcrn

skllcrn:
Sure is @alokpant! We've implemented some, but there's still loads of room for improvement.

alokpant:
I read that this is part of rewrite. If there is anything I can help with, then please let me know. I will look into issues section.

skllcrn:
I'll also add that any elements such as icons without text should have an `alt` attribute, and text that needs more context or that can be ellipsed/truncated have their full content in a `title` attribute.

isaiahnields:
Are we still looking for someone to work on these improvements?

I noticed the following in v 3.0.0:

1. There is no highlighted / selected state for any of the buttons on the action bar.
2. Pressing the tab button immediately after the action bar is displayed takes the user to the size selection and some glitches seem to appear here with the record button.

If no one else is working on this, I'm happy to give it a go.

skllcrn:
Would love your help @isaiahnields, let me know if there's anything you need!

isaiahnields:
@skllcrn I made a pull request that fixes the tabbing for the width and height inputs. It doesn't add the ability to tab over the buttons, however.

skllcrn:
Being able to tab through the UI is great for both accessibility and "power users". I agree that the default focus states might not look great, but is is the standard for native macOS apps, so I'd ague that we should adhere to that. Any thoughts here @sindresorhus?

sindresorhus:
Kap already does a lot of custom UI, so I don't see anything wrong by adding a custom focus state style. I think users should be able to tab through buttons too.

PR #593, opened 2018-11-23
Fix tabbing glitches on the action bar

isaiahnields:
This pull request fixes animation / GUI errors that occur when a user presses the tab key immediately after the action bar is displayed.

Fixes part of #136.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ node_modules
 /renderer/.next
 /app/dist
 /dist
-.idea/*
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules
 /renderer/.next
 /app/dist
 /dist
+.idea/*
```
sindresorhus: This should be in your own global gitignore file and not here.

skllcrn:
Are you still interested in finishing this PR @isaiahnields?

Issue #559, opened 2018-10-04
The `...` button in the export window should use the `IconMenu` component


sindresorhus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2480%20Funded-%2300A156.svg" />](https://issuehunt.io/r/wulkano/kap/issues/559)

https://github.com/wulkano/kap/blob/master/renderer/components/icon-menu.js

- [ ] I also noticed that clicking close to the edge of the square does not trigger the menu. This is it is expected you could click anywhere in the square. Maybe the square should have some kind of hover effect too, to make it totally clear it is clickable.

- [ ] I also don't think the export item should be hoverable during export as it makes me think I can click it to do something. Maybe the text with the filename should also look disabled (more gray)?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($80.00)
### Submitted pull Requests
- [#742 The `...` button in the export window should use the `IconMenu` component #559](https://issuehunt.io/r/wulkano/kap/pull/742)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/wulkano/kap/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

karaggeorge:
It is using it:
https://github.com/wulkano/kap/blob/master/renderer/components/exports/export.js#L54

> it is expected you could click anywhere in the square.
I tried this before. I can bring it back. But that means the menu icon should open underneath the whole square right? or just under the `...` again?

sindresorhus:
> It is using it:
/renderer/components/exports/export.js@master#L54

Hmm, weird. I was sure it was opening under my mouse. I must have been on the wrong branch or something.

> I tried this before. I can bring it back. But that means the menu icon should open underneath the whole square right? or just under the ... again?

Good point. Maybe if we just make it clearer where the click area of the `...` button is? For example, give `...` a hover effect.

karaggeorge:
[...]

PR #594, opened 2018-11-23
Exports GUI Improvements

isaiahnields:
This fixes the GUI to match the ideas laid out in #559.

### DIFF
--- a/renderer/components/exports/export.js
+++ b/renderer/components/exports/export.js
@@ -50,8 +50,8 @@ export default class Export extends React.Component {
           <div className="icon">
             {
               cancelable ?
-                <CancelIcon fill="white" hoverFill="white" onClick={cancel}/> :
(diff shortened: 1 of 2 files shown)


renderer/components/exports/export.js
```diff
@@ -50,8 +50,8 @@ export default class Export extends React.Component {
           <div className="icon">
             {
               cancelable ?
-                <CancelIcon fill="white" hoverFill="white" onClick={cancel}/> :
-                <IconMenu onOpen={menu && menu.popup}><MoreIcon fill="white" hoverFill="white"/></IconMenu>
+                <div className="icon" onClick={cancel}><CancelIcon fill="white" hoverFill="white"/></div> :
+                <IconMenu fullScreen onOpen={menu && menu.popup}><MoreIcon fill="white" hoverFill="white"/></IconMenu>
```
sindresorhus: I don't think `fullScreen` is the correct naming for this.

PR #603, opened 2018-12-22
Drag and drop from export history

### LINKED ISSUE #602: Drag and drop from export history
Ability to drag and drop kaptures directly from the export window

albinekb:
Fixes #602 

I can not get https://electronjs.org/docs/api/native-image#nativeimagecreatefromdataurldataurl to work with the kapture image, any ideas why or how to get the correct image when dragging?

Do we want the screenshot or a static image for all dragged files?


main/export-list.js
```diff
@@ -193,6 +195,19 @@ class ExportList {
       openEditorWindow(exp.inputPath, exp.originalFps);
     }
   }
+
+  dragExport(createdAt) {
+    track('export/history/drag/recording');
+    const exportsWindow = getExportsWindow();
+
+    const exp = this.exports.find(exp => exp.createdAt === createdAt);
```
sindresorhus: `exp` => `export`

main/export-list.js
```diff
@@ -2,6 +2,7 @@
 'use strict';
 
 const {dialog} = require('electron');
+const {join: pathJoin} = require('path');
```
sindresorhus: There's no benefit of using destructuring here. `const path = require('path')` would be better.

main/export-list.js
```diff
@@ -19,6 +20,7 @@ const {toggleExportMenuItem} = require('./menus');
 const Export = require('./export');
 
 const ffmpegPath = util.fixPathForAsarUnpack(ffmpeg.path);
+const fileIconPath = pathJoin(__dirname, '../', 'static/fileIcon.png');
```
sindresorhus:
```suggestion
const fileIconPath = pathJoin(__dirname, '../static/fileIcon.png');
```

main/export-list.js
```diff
@@ -193,6 +195,19 @@ class ExportList {
...
   }
+
+  dragExport(createdAt) {
+    track('export/history/drag/recording');
+    const exportsWindow = getExportsWindow();
+
+    const exp = this.exports.find(exp => exp.createdAt === createdAt);
+    if (exp) {
+      exportsWindow.webContents.startDrag({
+        file: exp.context.targetFilePath,
```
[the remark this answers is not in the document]
sindresorhus: Why do you think it's incorrect? (I haven't really looked into this yet)

sindresorhus:
> Do we want the screenshot or a static image for all dragged files?

Screenshot

main/export-list.js
albinekb: This is copy-pasted from above, should i change that too? or make a function to get an export by createdAt?

main/export-list.js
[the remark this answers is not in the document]
albinekb: It seems to only be set by the default file export share plugin, see:
https://github.com/wulkano/kap/blob/d021f0cc1e936782928c12501ea4c733e9aaa4ee/main/export-list.js#L110-L131

main/export-list.js
albinekb: I agree 👍

main/export-list.js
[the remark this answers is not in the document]
karaggeorge: Yeah, the rest of the plugins only use the temp file(s). Only the default plugin actually copies it to the "target file path"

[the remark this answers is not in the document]
karaggeorge: Pretty sure you can use `export.filePath` to get the path of the file after the conversion. It gets the value after the conversion is completed so we would have to make sure you can't drag/drop before it's done:

https://github.com/wulkano/kap/blob/c0baee3659b97e6e9c38868416ab4748cb1a3646/main/export.js#L111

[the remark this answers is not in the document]
albinekb: Let's try that 👍 https://superuser.com/questions/187071/in-macos-how-often-is-tmp-deleted

>Short answer: by default, files that aren't accessed in three days are deleted from /tmp

main/export-list.js
sindresorhus: > or make a function to get an export by createdAt?

👍

albinekb: `export`  seems to be restricted keyword?
image
image

karaggeorge: I remember that haha. Can't use export as a variable name because it's reserved

albinekb:
@sindresorhus I fixed your comments. We can't use `export` as the variable name as it's restricted. 

I can't get the screenshot to work, any ideas on where i should get the sceenshot image to show when dragging?

The only thing i got to work was generating the screenshot just when you start dragging, which causes the drag to appear stuttering in the beginning. When I am using the same screenshot data uri with electron's api to convert it to an NativeImage, the app crashes when dragging the screenshot.

main/export-list.js
sindresorhus: Ugh, I hate that. In Swift, you can at least wrap the variable name in backticks to allow using reserved names.

sindresorhus: `exp` => `exportItem` then?

albinekb: yep, that's better

sindresorhus:
Closing in favor of #641.

PR #608, opened 2019-01-23
Ignore .DS_Store

LitoMore:
Ignore `.DS_Store` files :)


sindresorhus:
This should be in your own global gitignore. It’s not project-specific.

LitoMore:
Good point. Thanks for reminding me that.

PR #616, opened 2019-01-31
Add tab support to the cropper

### LINKED ISSUE #136: UX Improvement: Update control panel tabindex

karaggeorge:
Fixes #136 

Used the last commit of #593 as the first ones seemed to all be reverted and built on it


karaggeorge:
@sindresorhus Should be fixed

karaggeorge:
- Ah, I didn't know the <kbd>↓</kbd> was standard behavior. I can add it. Do we have to remove the <kbd>↵</kbd> button, or can that stay? 
- I would think <kbd>↵</kbd> would be better than <kbd>space</kbd>
- Is that something we want? / what would that focused state look

karaggeorge:
To me, space feels like it's for selecting apps (because of how close this behavior is to the native screenshot tool), but I could go either way.

Also, what's the default `focus` state for the record button? I has none right now

sindresorhus:
> * I think we should remove ↵ as that's typically reserved for making a selection within the `select` and to submit the entire form @karaggeorge, but I'll defer to @sindresorhus on this.

Correct. Only `↓`.

skllcrn:
For the record button I think it makes the most sense to invert it @karaggeorge
 
image

karaggeorge:
@skllcrn @sindresorhus 

Changed the menus so now they are opened by the arrow and not enter or space. (might be hard to identify since the regular buttons and the dropdowns don't look any different), and also added focused state to the record button.

Do we still want the global enter/space key to start recording? Or are those okay to being applicable when the record button is focused?

skllcrn:
For now I suggest just supporting <kbd>↵</kbd> when focused on the record button. We can look more into a dedicated keyboard shortcut to start the recording separately.

sindresorhus:
I'm not seeing any record button focus state.

sindresorhus:
If I click the record button right after starting up, it looks a bit weird:

screen shot 2019-02-07 at 02 22 46

karaggeorge:
@sindresorhus should be good

sindresorhus:
Looks good 👍

PR #621, opened 2019-02-01
Resize from center

### LINKED ISSUE #612: Add keyboard shortcuts to resize the cropper from the center
On macOS, you can resize while pressing <kbd>Shift</kbd> to resize from the center. And <kbd>Shift</kbd><kbd>Option</kbd> to resize from the center while keeping the aspect ratio. Would be useful if the cropper in Kap supported that. It could be useful for being able to expand the cropper for a window to also include the shadow or just more context.

These shortcuts are not even well known for most macOS users, so we need to document it somewhere. Maybe as a Tips section in the preferences?

karaggeorge:
Fixes #612 

Adds logic to resize from the center of the cropper when pressing the <kbd>Option</kbd> key, and supports using both that and locking ratio at the same time

This is a WIP. Getting a PR up so other people can download the build and test it

### DIFF
--- a/renderer/components/cropper/overlay.js
+++ b/renderer/components/cropper/overlay.js
@@ -47,9 +47,6 @@ class Overlay extends React.Component {
       'no-transition': isResizing || isMoving || !isActive
     });
 
-    const centerX = Math.round(x + (width / 2));
-    const centerY = Math.round(y + (height / 2));
-
     return (
       <div
         className={contentClassName}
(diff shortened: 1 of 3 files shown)


sindresorhus:
The resizing logic seems to work perfectly. I was not able to make it break.

I noticed there's a red dot in the middle now. Is that intentional? The size text is a bit too close to the red dot and the text is invisible on light backgrounds.

screen shot 2019-02-03 at 22 12 36

skllcrn:
The red dot etc. is just for testing purposes @sindresorhus, can you remove that now @karaggeorge, if we’re all done testing?

karaggeorge:
@sindresorhus Yes, that dot was for debugging purposes. It helped me verify that the center was indeed not changing when resizing. I've removed that part of the code. Should be good to go

PR #623, opened 2019-02-02
Use input type password

ochorocho:
Use input type password for config fields in plugins containing "password" in name.

Example Plugin config

```
...
password: {
      title: 'Password',
      type: 'string',
      required: true
},
...
```
Result:

bildschirmfoto 2019-02-02 um 19 12 54

### DIFF
--- a/renderer/components/config/tab.js
+++ b/renderer/components/config/tab.js
@@ -9,11 +9,11 @@ import {OpenOnGithubIcon, OpenConfigIcon} from '../../vectors';
 const ConfigInput = ({name, type, schema, value, onChange, hasErrors}) => {
   if (type === 'string') {
     const className = hasErrors ? 'has-errors' : '';
-    const password = name.match("password") ? "password" : "text";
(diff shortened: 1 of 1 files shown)


sindresorhus:
It's too fragile to just check the title. Having a password field might not be what the author wants, even if the title contains the word "password".

Instead, we should add a custom [`ajv`](https://github.com/epoberezkin/ajv) password type.

I imagine it would look like this then:

```js
password: {
      title: 'Password',
      type: 'password',
      required: true
},
```

sindresorhus:
It's not enough to just create the keyword. We also need to check for it in the settings window renderer and use a `<input type="password">` instead of normal input if it's a `type: 'password'`.

skllcrn:
Just checking in, are you still wanting to wrap this one up @ochorocho?

ochorocho:
Well, i can't get this working. Due to lag of Javascript knowledge. But it would be great to have such a addtional field.

sindresorhus:
@ochorocho Can you open a new issue about this instead?

PR #626, opened 2019-02-06
Support HEVC files in the editor

### LINKED ISSUE #341: Use HEVC instead of H264
We can't do for now, as Chrome doesn't yet support it, and they have no plans to support it yet. Just opening this issue so we can do it eventually.

karaggeorge:
Fixes #341 

Adds some type of HEVC support. When a file is opened, we use ffmpeg to check its encoding and if it is HEVC, we use ffmpeg to convert it to a temp H264 file and open that in the editor.

Also makes the opening file a shared utility and uses it everywhere (tray/right-click -> open, menus) to get the same behavior everywhere.

TODOs: 
- [x] ~One question at the moment is that since we open the editor with the temp H264 file, the conversions queued will use that as input instead of the original HEVC. If that sacrifices quality, we can switch the editor implementation to use one file for preview but a different one for conversions.~ Done. Uses the original file for conversions and the converted only for preview
- [ ] Record in HEVC if the hardware supports it. Waiting on https://github.com/wulkano/aperture-node/pull/6

### DIFF
--- a/main/editor.js
+++ b/main/editor.js
@@ -19,16 +19,16 @@ const MIN_WINDOW_HEIGHT = MIN_VIDEO_HEIGHT + OPTIONS_BAR_HEIGHT;
 
 const getEditorName = (filePath, isNewRecording) => isNewRecording ? `New Recording ${moment().format('YYYY-MM-DD')} at ${moment().format('H.mm.ss')}` : path.basename(filePath);
 
-const openEditorWindow = async (filePath, recordFps, {isNewRecording} = {isNewRecording: false}) => {
+const openEditorWindow = async (filePath, {recordedFps, isNewRecording, originalFilePath} = {}) => {
   if (editors.has(filePath)) {
     editors.get(filePath).show();
     return;
   }
 
-  const fps = recordFps || await getFps(filePath);
+  const fps = recordedFps || await getFps(filePath);
 
   const editorWindow = new BrowserWindow({
-    title: getEditorName(filePath, isNewRecording),
+    title: getEditorName(originalFilePath || filePath, isNewRecording),
     minWidth: MIN_VIDEO_WIDTH,
     minHeight: MIN_WINDOW_HEIGHT,
     width: MIN_VIDEO_WIDTH,
@@ -70,7 +70,7 @@ const openEditorWindow = async (filePath, recordFps, {isNewRecording} = {isNewRe
 
   editorWindow.webContents.on('did-finish-load', async () => {
     ipc.callRenderer(editorWindow, 'export-options', exportOptions);
-    await ipc.callRenderer(editorWindow, 'file', {filePath, fps});
+    await ipc.callRenderer(editorWindow, 'file', {filePath, fps, originalFilePath});
     editorWindow.show();
   });
 };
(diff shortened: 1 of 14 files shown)


karaggeorge:
@sindresorhus @skllcrn 

How do we want to continue on this? It seems like the results of exports will always be h264 even if you opened or recorded a hevc file.

I think the ability to open hevc files is still useful, but do we want to keep the recording with hevc when the hardware supports it? And maybe add a save original option somewhere if the user wants to get the original recording with no alterations?

sindresorhus:
> It seems like the results of exports will always be h264 even if you opened or recorded a hevc file.

Can you add some context here from the Slack discussions. The problem was that you could not get `ffmpeg` to change the size/FPS of HEVC and then export as HEVC, right? If so, we can ignore that for now, and you can just open an issue about fixing that in the future. It should be fixable somehow, but not essential for this PR.

> I think the ability to open hevc files is still useful

👍 

> but do we want to keep the recording with hevc when the hardware supports it?

👍 

> And maybe add a save original option somewhere if the user wants to get the original recording with no alterations?

👍 In the `File` menu.

karaggeorge:
Yeah, so the issue was that when converting `ffmpeg`, if you don't specify a video codec it will use the default it has for that type. For mp4s etc it is `h264`. If we want to tell it to use `hevc` for encoding, we can either do:
	-  `-vcodec copy` which will keep the encoding at `hevc`, but since the video stream never gets decoded, you can't change dimensions, fps or anything else
	- `-vcodec hevc` which in theory should work but it fails on my machine. I can't find examples for the options I need to use for that or test any of it since I can't get it to work. It might be because my hardware can't handle hevc to begin with, even though ffmpeg says I have that encoder. 

If the same criteria for recording hevc exist for encoding to hevc, we might be able to use that same check to use the encoder, but until someone can test it and find the right `ffmpeg` arguments, we can't export in hevc format.

I'll open an issue with this info so we can track it further

karaggeorge:
Added a `Save Original…` option to the file menu when an editor is focused and tested that it saves it with `hevc` format if the source was `hevc`

main/editor.js
```diff
@@ -81,8 +87,13 @@ const setOptions = options => {
 
 const getEditors = () => editors.values();
 
+ipc.answerRenderer('save-original', ({inputPath, outputPath}) => {
+  cpFile(inputPath, outputPath);
```
sindresorhus: You can just use the native method here. `fs.copyFile()`. Make sure you pass the `fs.constants.COPYFILE_FICLONE` to make copying fast.

karaggeorge: e3168b3

main/menus.js
```diff
@@ -121,6 +122,16 @@ const applicationMenuTemplate = [
       {
         type: 'separator'
       },
+      {
+        label: 'Save Original…',
```
sindresorhus: It should have the Command+S accelerator.

karaggeorge: 41c7479

sindresorhus:
🥇

sindresorhus:
> I'll open an issue with this info so we can track it further

Don't forget to open that issue.

PR #631, opened 2019-02-15
Fix Kap preventing sleep

karaggeorge:
When an editor was open, Kap would prevent the computer from going sleep. This was caused by chrome's functionality of preventing sleep when there's media playing on the page. Until a better solution is found, we decided to at least pause the editor window when it is blurred, and resume when focused.

This indeed removes Kap from the list of Apps that prevent sleep a couple of seconds after blurring. The issue, of course, will remain if you keep the editor focused.

Questions:
- While doing this, I was also wondering if it would be worth utilizing that functionality and preventing the computer from sleeping while aperture is recording. I think I've heard of a couple of issues of putting the computer to sleep while recording, but I'm not 100% sure.

### DIFF
--- a/main/editor.js
+++ b/main/editor.js
@@ -72,8 +72,14 @@ const openEditorWindow = async (filePath, {recordedFps, isNewRecording, original
     });
   }
 
-  editorWindow.on('blur', () => editorEmitter.emit('blur'));
-  editorWindow.on('focus', () => editorEmitter.emit('focus'));
+  editorWindow.on('blur', () => {
+    editorEmitter.emit('blur');
+    ipc.callRenderer(editorWindow, 'blur');
+  });
+  editorWindow.on('focus', () => {
+    editorEmitter.emit('focus');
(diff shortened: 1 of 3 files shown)


sindresorhus:
If I pause the video, blur the window, and then focus the window again it starts playing, but it should be paused like it was before being blurred.

karaggeorge:
@sindresorhus 

Good catch, updated. Also, fixed the preferences categories color issue. It will need some more touching up (the gray background focus state) but I'm working on a follow up PR to add tab accessibility to the preferences window, so I'll address further there.

sindresorhus:
> * While doing this, I was also wondering if it would be worth utilizing that functionality and preventing the computer from sleeping while aperture is recording. I think I've heard of a couple of issues of putting the computer to sleep while recording, but I'm not 100% sure.

Better to do that at the source: https://github.com/wulkano/Aperture/pull/64

PR #640, opened 2019-03-01
Replace PropTypes.func with PropTypes.elementType

### LINKED ISSUE #633: Use the new `elementType` PropType
See https://github.com/facebook/prop-types/pull/211

We should use it instead of `PropType.func` in most places.

maddieeiddam:
Closes #633


maddieeiddam:
@skllcrn @karaggeorge @sindresorhus Hmm - so I can see that the build is failing because it's seeing elementType as an invalid prop type...did I make an error, do I need to change something else, or is elementType just not supported yet?

sindresorhus:
The build failure is because of https://github.com/yannickcr/eslint-plugin-react/issues/2188. I've disabled the rule in `master`. Just rebase from `master`.

ljharb:
That rule enumerates the actual `prop-types` library; did perhaps someone forget to update `prop-types`? I see this repo has a lockfile; that's a common bug when using one.

sindresorhus:
Yes, it was the lockfile. I thought it was not as I had already tried to uninstall and then reinstall `eslint-plugin-react`, but it seemed I had to force regenerate the lockfile to make it work.

sindresorhus:
Looks good to me.

@maddieeiddam Can you fix the merge conflict?

sindresorhus:
Thanks, @maddieeiddam 🙌

PR #642, opened 2019-03-09
Fix aperture stop error

karaggeorge:
Sentry error: [KAP-3G3](https://sentry.io/organizations/wulkano-l0/issues/826041989/?project=255536&query=is%3Aunresolved+release%3AKap-Beta3.0.0-beta.5&statsPeriod=14d&utc=false)

Take steps to avoid `stopRecording` from being called if we're not recording

This also fixes a bug that the current beta has, where if you clear your accelerator but keep the toggle enabled and try to record, when you stop the app crashes, so aperture keeps recording until you force quite Kap

### DIFF
--- a/main/common/aperture.js
+++ b/main/common/aperture.js
@@ -119,34 +119,43 @@ const startRecording = async options => {
 };
 
 const stopRecording = async () => {
+  // Ensure we only stop recording once
+  if (!past) {
+    return;
+  }
+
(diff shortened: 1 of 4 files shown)


sindresorhus:
One step closer to no Sentry errors ❤️

skllcrn:
Sentry issue: [KAP-3KR](https://sentry.io/organizations/wulkano-l0/issues/908567479/?project=255536&referrer=github_plugin)

PR #643, opened 2019-03-10
Clean up Preferences and add tab support

karaggeorge:
Updates the tab display on the Preferences window and adds tab support.

Also adds tab support to the plugin config window.

Screen Shot 2019-03-09 at 7 13 07 PM

Finally, it attempts to fix this bug. Haven't been able to reproduce in this branch's build yet. I think it had something to do with the `Item` components not being the same when the page was server-rendered and when it was being rendered on the app. I gave the components a key so hopefully that fixed it

image

### DIFF
--- a/main/utils/macos-version.js
+++ b/main/utils/macos-version.js
@@ -1,0 +1,4 @@
+const macosVersion = require('macos-version');
+
+console.log(macosVersion);
+module.exports = macosVersion;
--- a/main/utils/macos-version.js
+++ b/main/utils/macos-version.js
@@ -1,4 +1,3 @@
 const macosVersion = require('macos-version');
 
-console.log(macosVersion);
 module.exports = macosVersion;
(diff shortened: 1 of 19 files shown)


sindresorhus:
In Plugins => Discover, if I tab until the last plugin in the list, and then tab again, it seems to select something hidden. If I tab yet another time, it selects the "General" tab.

sindresorhus:
Otherwise, looks good.

skllcrn:
This needs some design tweaks, but we can address that in #644.

karaggeorge:
@sindresorhus Yeah, that happens in all windows. I noticed it first when I was working on the cropper and it drove me crazy. The element that receives focus after the last one is the `body` of the page. It's a known thing in chrome/chromium apparently. I tried a bunch of things like setting `tabIndex` to -1 on it but nothing worked. So for all pages it takes one extra tab to go back to the start. Not sure if there's anything we can do about it at this point

sindresorhus:
@karaggeorge Ah ok. Can you open an issue so we can track it? Maybe some random person passing by have a good workaround. Would be good to include a link to the Chrome issue too.

PR #644, opened 2019-03-10
Dark Mode 🎉

### LINKED ISSUE #478: Dark Mode support
Blocked by: 

https://github.com/electron/electron/issues/12726
https://github.com/electron/electron/issues/13175
https://www.chromestatus.com/feature/5679432723333120

karaggeorge:
Closes #478 

Support dark mode and switching between live/dark while the windows are open 😄 

Screen Shot 2019-03-10 at 3 13 21 AM

Kapture 2019-03-10 at 3 11 52


karaggeorge:
This should wait for #643 since some styles introduced in that PR will need the dark/light css vars

sindresorhus:
It looks beautiful! ❤️

sindresorhus:
The blue link color is not readable enough:

Screen Shot 2019-03-10 at 21 31 13

sindresorhus:
Screen Shot 2019-03-10 at 21 32 13

skllcrn:
I've been experimenting with making all blue elements white when in dark mode @sindresorhus, I think it's a decent solution since we rarely if ever rely on colour alone to indicate states. Thoughts?

sindresorhus:
@skllcrn I think that would make it look very bland. I like the blue, it just need to be slightly more legible.

karaggeorge:
I agree that I like the blue. But yeah the one we are using is a bit too much for dark mode. I already added some variables for that as well, so it should be pretty easy to switch it

karaggeorge:
@skllcrn @sindresorhus Gave the editor slider a dark mode version and switched the blue with the "kap-light" color from the Abstract palette for dark mode:

Screen Shot 2019-03-10 at 3 56 29 PM

let me know if that makes it more legible or if we need to go a bit lighter than that

sindresorhus:
I feel like the blue links could still be a bit more legible.

sindresorhus:
You also need to set https://electronjs.org/docs/api/system-preferences#systempreferencessetapplevelappearanceappearance-macos when changing to/from dark mode.

Example: https://github.com/sindresorhus/caprine/blob/ac689dec73e1e24b847971484dab484077629fb0/source/index.ts#L441

sindresorhus:
The toggle component in light mode is broken:

Screen Shot 2019-03-14 at 23 09 17

sindresorhus:
The tab here are broken and moves when you click them:

Screen Shot 2019-03-14 at 23 09 58

sindresorhus:
No visible text on the button here:

Screen Shot 2019-03-14 at 23 10 41

sindresorhus:
The links here are no longer colored, which makes it hard to see that they are links:

Screen Shot 2019-03-14 at 23 11 36

karaggeorge:
@sindresorhus All those visual bugs happened because of a wrong css var setting which I fixed. If you check again, all those should be good.


> You also need to set electronjs.org/docs/api/system-preferences#systempreferencessetapplevelappearanceappearance-macos when changing to/from dark mode.

Doesn't [this](https://github.com/wulkano/kap/blob/0f4a583707cb473b6bde72a3f93df4ca928c1fa8/package.json#L121) cover that, since we always want to follow the system setting: 

For me the packaged version's window bars switch correctly between dark/light. Unless that setting covers something else

sindresorhus:
> Doesn't this cover that, since we always want to follow the system setting?

Ah yes, I forgot we don't have a toggle to override dark/light mode.

onordbo:
> No visible text on the button here:
> 
> Screen Shot 2019-03-14 at 23 10 41

Looking at this one, we need some more padding on the right side of the text string to account for long objects like _"Mute Dropbox desktop notification for the upload"_

skllcrn:
Just for additional context, you're on the latest macOS @onordbo? In your screenshot the plugin configuration window is also appearing at a position further down on the screen than it should.

karaggeorge:
@skllcrn anything else left for this one?

skllcrn:
We've got some styling issues left @karaggeorge: 

- The blue isn't legible enough on the dark background, especially for small regular fonts, we can either go monochrome or tweak our palette to pass accessibility tests. 
- The plugin configuration window spawns 2 scrollbars 
- Window backgrounds globally seem to be 10-20% darker that they are intended to be and feels inconsistent with other apps, are you seeing that?

karaggeorge:
@skllcrn 
> The blue isn't legible enough on the dark background, especially for small regular fonts, we can either go monochrome or tweak our palette to pass accessibility tests.

I'm thinking pick a different blue? Can we try some hexcodes?

> The plugin configuration window spawns 2 scrollbars

I'm not seeing this locally. I'll try to download the built version and see if there's something there

> Window backgrounds globally seem to be 10-20% darker that they are intended to be and feels inconsistent with other apps, are you seeing that?

What should the bg color be? I think we set it together on a call from Abstract

karaggeorge:
@sindresorhus This is ready for a final review. @skllcrn and I got on a call and fixed almost everything.

There's two things left, but we can handle them in separate PRs:

- About window needs a dark mode title bar
- Changing the main bar transition when going between regular/advanced from a slide animation to a fade in/out

I'll work on those in the next few days

sindresorhus:
> About window needs a dark mode title bar

It already has that when built for production: 
Screen Shot 2019-07-08 at 22 53 15

sindresorhus:
It's not obvious that the links in the about window are links. They're just white like the other text. It was clearer when they were blue.

Other than that, this looks very good.

PR #649, opened 2019-03-16
Work around the "Save" dialog bug in Electron

### LINKED ISSUE #368: File browser prompt expand/collapse not working
**Platform:** MacOS 10.13.3
**Kap Version:** 2.0.0 (2.0.0.681)

### Steps to reproduce
When saving an exported format, the file browser only shows the collapsed version. Attempts to expand to view file list do not produce the file list.
1. After capturing the screen, view the playback
2. select Export GIF from controls on the bottom right of the screen
3. this will reveal the file browser prompt

### Current behaviour
click expor gif shows the file system prompt:
screen shot 2018-02-01 at 2 02 09 pm

after clicking arrow down to expand (at the end of Save As line)
screen shot 2018-02-01 at 2 03 10 pm

**the screen does not expand to show the contents of the folder.**

### Expected behaviour
After clicking arrow down to expand (at the end of Save As line), screen expands to show file system (as this other example):
screen shot 2018-02-01 at 2 05 01 pm

### Workaround
Did not encounter a method to expand the screen.

karaggeorge:
(Hopefully) Fixes #368 

Tested locally on all 3 places we use a save dialog, and it's working fine

Used the workaround mentioned here: https://github.com/wulkano/kap/issues/368#issuecomment-473526414

`promisify` epxects an error-first callback, while electron's `showSaveDialog` calls it just with the file path or `undefined` if the user cancels. I used a catch block to get the file path, not sure if there's a better way, or if I should just write a small utility to implement `promisify` for this specific use-case

### DIFF
--- a/main/editor.js
+++ b/main/editor.js
@@ -94,7 +94,18 @@ const setOptions = options => {
 
 const getEditors = () => editors.values();
 
-ipc.answerRenderer('save-original', ({inputPath, outputPath}) => promisify(fs.copyFile)(inputPath, outputPath, fs.constants.COPYFILE_FICLONE));
+ipc.answerRenderer('save-original', async ({inputPath}) => {
+  const now = moment();
+
+  try {
+    await promisify(dialog.showSaveDialog)(BrowserWindow.getFocusedWindow(), {
+      defaultPath: `Kapture ${now.format('YYYY-MM-DD')} at ${now.format('H.mm.ss')}.mp4`
+    });
+    // eslint-disable-next-line unicorn/catch-error-name
+  } catch (path) {
+    return promisify(fs.copyFile)(inputPath, path, fs.constants.COPYFILE_FICLONE);
+  }
+});
 
 module.exports = {
   openEditorWindow,
--- a/main/editor.js
+++ b/main/editor.js
@@ -21,6 +21,7 @@ const MIN_VIDEO_HEIGHT = MIN_VIDEO_WIDTH * VIDEO_ASPECT;
 const MIN_WINDOW_HEIGHT = MIN_VIDEO_HEIGHT + OPTIONS_BAR_HEIGHT;
 const editorEmitter = new EventEmitter();
 
+const showSaveDialog = pify(dialog.showSaveDialog, {errorFirst: false});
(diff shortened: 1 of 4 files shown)


sindresorhus:
Use https://github.com/sindresorhus/pify with the `{errorFirst: false}` option.

karaggeorge:
@sindresorhus I knew there was something out there, but I couldn't find it. Updated 👍

main/editor.js
```diff
@@ -94,7 +94,17 @@ const setOptions = options => {
...
-ipc.answerRenderer('save-original', ({inputPath, outputPath}) => promisify(fs.copyFile)(inputPath, outputPath, fs.constants.COPYFILE_FICLONE));
+ipc.answerRenderer('save-original', async ({inputPath}) => {
+  const now = moment();
+
+  const path = await pify(dialog.showSaveDialog, {errorFirst: false})(BrowserWindow.getFocusedWindow(), {
+    defaultPath: `Kapture ${now.format('YYYY-MM-DD')} at ${now.format('H.mm.ss')}.mp4`
+  });
+
+  if (path) {
+    return pify(fs.copyFile)(inputPath, path, fs.constants.COPYFILE_FICLONE);
```
sindresorhus: Better to await than return for readability reasons.

main/editor.js
```diff
@@ -94,7 +94,17 @@ const setOptions = options => {
 
 const getEditors = () => editors.values();
 
-ipc.answerRenderer('save-original', ({inputPath, outputPath}) => promisify(fs.copyFile)(inputPath, outputPath, fs.constants.COPYFILE_FICLONE));
+ipc.answerRenderer('save-original', async ({inputPath}) => {
+  const now = moment();
+
+  const path = await pify(dialog.showSaveDialog, {errorFirst: false})(BrowserWindow.getFocusedWindow(), {
```
sindresorhus: Would be more readable to promisify it at the top-level (top of the file):

```js
const showSaveDialog = pify(dialog.showSaveDialog, {errorFirst: false});
```

sindresorhus:
Works great 👌

PR #655, opened 2019-04-06
Fix incorrect filename being displayed after export

### LINKED ISSUE #651: Incorrect filename displayed after export
**macOS version:**
ProductName:	Mac OS X
ProductVersion:	10.14.3
BuildVersion:	18D109

**Kap version:**
3.0.0 Beta5

#### Steps to reproduce

- Create a recording
- Edit the automatically generated filename to be "Demo.mp4"
- Export as MP4

#### Current behaviour
Following export, Kap shows the automatically generated filename as "successfully exported", not the edited name

#### Expected behaviour
In this case, Kap should show "Demo.mp4" as the exported filename, not the automatically generated name as a file of that name was not exported.

#### Workaround
Live with it.  ;-)

kap

blandonfrank:
When using the save to file service, the filename is placed in the context target path. Using that to display under export list - rather than using the generated default one.

Fixes #651

### DIFF
--- a/main/export.js
+++ b/main/export.js
@@ -38,7 +38,7 @@ class Export {
 
   get data() {
     return {
-      defaultFileName: this.defaultFileName,
+      defaultFileName: this.defaultFileName && (this.isDefault ? this.context.targetFilePath.replace(/^.*[\\\/]/, '') : this.defaultFileName),
       text: this.text,
       status: this.status,
       percentage: this.percentage,
--- a/main/export.js
+++ b/main/export.js
@@ -7,6 +7,7 @@ const {track} = require('./common/analytics');
 const {convertTo} = require('./convert');
 const ShareServiceContext = require('./share-service-context');
 const Plugin = require('./plugin');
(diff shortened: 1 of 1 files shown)


sindresorhus:
You should use `path` methods instead of regex to modify the path.

blandonfrank:
sindresorhus - you're right. I'm a total JS beginner. I went ahead and changed it to use `path` instead.

main/export.js
```diff
@@ -38,7 +39,7 @@ class Export {
 
   get data() {
     return {
-      defaultFileName: this.defaultFileName,
+      defaultFileName: this.defaultFileName && (this.isDefault ? path.basename(this.context.targetFilePath) : this.defaultFileName),
```
karaggeorge:
```suggestion
      defaultFileName: this.isDefault ? path.basename(this.context.targetFilePath) : this.defaultFileName,
```

sindresorhus:
Thank you, @blandonfrank :)

PR #664, opened 2019-04-26
Upgrade to Electron 5 and upgrade other dependencies

sindresorhus:


sindresorhus:
The only issue I could see is that the export items in the export window doesn't show up, but I'm not sure why, so I'll open an issue after this PR is merged. I think it has to do with `electron-better-ipc` being upgraded and the existing logic having some kind of race issue, but not sure.

karaggeorge:
Hm, yeah that logic is a bit old and complex. I can look into it on this PR if you want and try to fix it if it's something small

skllcrn:
Maybe best to address that in a separate PR @sindresorhus and @karaggeorge?

sindresorhus:
Merging this now so we can continue testing Electron 5 for other things. I'll open a new issue if it's not something you can easily fix.

PR #671, opened 2019-05-09
Give the cropper an initial size

### LINKED ISSUE #556: Use a better initial size when setting the aspect ratio first

masquel:
Closes #556

Improved #570 

Added setSize to cropper container. Using setSize in action bar container, when toggle advanced.

### DIFF
--- a/renderer/containers/action-bar.js
+++ b/renderer/containers/action-bar.js
@@ -91,8 +91,14 @@ export default class ActionBarContainer extends Container {
 
   toggleAdvanced = () => {
     if (!this.cropperContainer.state.isFullscreen) {
-      const {advanced} = this.state;
+      const {advanced, screenWidth, screenHeight} = this.state;
       this.updateSettings({advanced: !advanced});
+      if (!advanced) {
+        this.cropperContainer.setSize({
+          width: screenWidth * 0.2,
+          height: screenHeight * 0.2
+        });
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


renderer/containers/action-bar.js
```diff
@@ -91,8 +91,14 @@ export default class ActionBarContainer extends Container {
 
   toggleAdvanced = () => {
     if (!this.cropperContainer.state.isFullscreen) {
-      const {advanced} = this.state;
+      const {advanced, screenWidth, screenHeight} = this.state;
       this.updateSettings({advanced: !advanced});
+      if (!advanced) {
+        this.cropperContainer.setSize({
+          width: screenWidth * 0.2,
```
albinekb: Maybe this should use the same logic as the other stuff to get 4:3 aspect?

masquel: added height calculation depending from width

renderer/containers/cropper.js
```diff
@@ -123,6 +123,16 @@ export default class CropperContainer extends Container {
     this.setState(updates);
   }
 
+  setSize = ({width: defaultWidth, height: defaultHeight}) => {
+    let {width, height} = this.state;
```
sindresorhus:
```suggestion
    const {width = defaultWidth, height = defaultHeight} = this.state;
```

masquel: Are you sure? In this case every time you click on crop area button, it will set the default size. For example, if user set size, return to main menu and click  crop button again, it will reset his settings to default.

sindresorhus: I'm not sure. It was just a suggestion when looking at the code without running it.

PR #672, opened 2019-05-14
Fix devices error

karaggeorge:
Fixes Sentry error [KAP-3G8](https://sentry.io/organizations/wulkano-l0/issues/832030699/?project=255536&query=is%3Aunresolved&statsPeriod=7d)

`TypeError: devices.some is not a function`

### DIFF
--- a/main/common/settings.js
+++ b/main/common/settings.js
@@ -94,7 +94,7 @@ const audioInputDeviceId = store.get('audioInputDeviceId');
 
   if (!Array.isArray(devices)) {
     const Sentry = require('../utils/sentry');
-    Sentry.captureMessage(`devices is not an array: ${JSON.stringify(devices)}`);
+    Sentry.captureException(new Error(`devices is not an array: ${JSON.stringify(devices)}`));
     return;
   }
 
--- a/main/common/settings.js
+++ b/main/common/settings.js
@@ -90,7 +90,7 @@ const audioInputDeviceId = store.get('audioInputDeviceId');
 (async () => {
   const devices = await audioDevices();
 
-  if (!devices.some(device => device.id === audioInputDeviceId)) {
+  if (devices && !devices.some(device => device.id === audioInputDeviceId)) {
     const [device] = devices;
     if (device) {
       store.set('audioInputDeviceId', device.id);
--- a/main/common/settings.js
+++ b/main/common/settings.js
@@ -82,6 +82,8 @@ const store = new Store({
   }
 });
 
+module.exports = store;
+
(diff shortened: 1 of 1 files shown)


sindresorhus:
I agree with the change, but it would be good to figure out why `devices` is not an array in the first place.

main/common/settings.js
```diff
@@ -90,7 +90,7 @@ const audioInputDeviceId = store.get('audioInputDeviceId');
 (async () => {
   const devices = await audioDevices();
```
sindresorhus: Note https://github.com/wulkano/aperture-node/blob/ef5cd98b9cc6496405bf36177dea79a11cf4952f/index.js#L165, so we could maybe do:

```js
if (!Array.isArray(devices)) {
	// Sentry logging here
	// console.log(devices);

	return
}
```

main/common/settings.js
```diff
@@ -90,12 +92,16 @@ const audioInputDeviceId = store.get('audioInputDeviceId');
...
 
+  if (!Array.isArray(devices)) {
+    const Sentry = require('../utils/sentry');
+    Sentry.captureMessage(`devices is not a string: ${JSON.stringify(devices)}`);
```
sindresorhus: You mean not an array?

```suggestion
    Sentry.captureMessage(`devices is not an array: ${JSON.stringify(devices)}`);
```

karaggeorge: 🤦‍♂ Yes

main/common/settings.js
```diff
@@ -90,12 +92,16 @@ const audioInputDeviceId = store.get('audioInputDeviceId');
 (async () => {
   const devices = await audioDevices();
 
+  if (!Array.isArray(devices)) {
+    const Sentry = require('../utils/sentry');
+    Sentry.captureMessage(`devices is not an array: ${JSON.stringify(devices)}`);
```
sindresorhus: It's unclear from the [docs](https://github.com/getsentry/sentry-electron) if Sentry saves the message here when there's no exception. I think `captureMessage` is meant to add additional context to an error, but there's no error here. Let's use `captureException` just to be sure it's saved.

sindresorhus: // @karaggeorge  ⬆️

karaggeorge: Sorry, have been super busy at work lately 😅. Used `captureException`, their types say that it accepts `any` but the annotation says 

> an exception-like object

So, I passed in a new error. That should work right?

sindresorhus: Yup, I would assume so.

PR #684, opened 2019-07-07
Better error handling around recording

### LINKED ISSUE #677: Error alert (could not start recording within 5 seconds) and app freezes
**macOS version:**
**Kap version:**

#### Steps to reproduce
1. Click Kap icon. 
2. Create capture window. 
3. Hit record button

#### Current behaviour
Nothing happens for about 5 seconds, then an alert pops up stating it could not record within 5 seconds. The app then freezes. I have to force quit through the Activity Monitor.

https://www.dropbox.com/s/1n22v6r7unrjbr4/KapBug.mov?dl=0

karaggeorge:
This PR adds some error handling around recording. It seems like we are having more issues (especially with the new macOS), but Kap doesn't handle them gracefully. The cropper windows stay open and you have to force kill the app to get rid of them. This PR allows Kap to reset the recording UI if something goes wrong and show the user a message about the error.

Also, disables HEVC recording, until we can figure out the details with ffmpeg and HEVC encoding

Fixes #677

### DIFF
--- a/main/common/aperture.js
+++ b/main/common/aperture.js
@@ -10,12 +10,17 @@ const {closePrefsWindow} = require('../preferences');
 const {setRecordingTray, disableTray, resetTray} = require('../tray');
 const {disableCroppers, setRecordingCroppers, closeAllCroppers} = require('../cropper');
 const {setCropperShortcutAction} = require('../global-accelerators');
+
+// eslint-disable-next-line no-unused-vars
 const {convertToH264} = require('../utils/encoding');
+
 const settings = require('./settings');
 const {track} = require('./analytics');
 
 const aperture = createAperture();
 const {audioDevices, videoCodecs} = createAperture;
+
+// eslint-disable-next-line no-unused-vars
 const recordHevc = videoCodecs.has('hevc');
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
While recording, if I accidentally click close the DevTools window for the recorder, the whole window get's white and it's impossible to quit Kap. Not relevant for the user, but super annoying when it happens while developing Kap:

Screen Shot 2019-07-08 at 22 44 06

karaggeorge:
Yeah, I noticed that too. I'm not sure what's happening with it, but I'll try to see if I can work around it.

It looks like the window is losing transparency. If that's a BrowserWindow level, might not be much I can do, if that's in the React code, I can track it down

sindresorhus:
Cool. I'm gonna merge this as my above comment is kinda unrelated and this looks good.

karaggeorge:
Not sure what's causing this. The page seems to be working for me, so I'm guessing the react part is working. Also I can close it by clicking anywhere in the overlay area (as in, my app doesn't get stuck). But I have no idea why that happens. Might be an electron issue

PR #685, opened 2019-07-07
Action bar animation: fade in/out instead of scrolling

karaggeorge:
This PR changes the animation of the menu-bar when switching between the main and advanced view as discussed with @skllcrn


sindresorhus:
Good choice. It does indeed look better.

PR #686, opened 2019-07-07
Improve plugin discoverability

### LINKED ISSUE #648: Plugin discoverability
It's hard for people to find plugins if they don't know to look for it. @karaggeorge suggested we add an option to view available plugins within the editor export options. 

I suggest only showing that hint if there are *no* plugins installed, just as a little nudge in the right direction. So we add e.g. a `Get plugins...` in the `select`, that when you click it opens preferences on the plugin overview.

karaggeorge:
Fixes #648

### DIFF
--- a/renderer/components/editor/options/right.js
+++ b/renderer/components/editor/options/right.js
@@ -11,6 +11,13 @@ class RightOptions extends React.Component {
     const formatOptions = options ? options.map(({format, prettyFormat}) => ({value: format, label: prettyFormat})) : [];
     const pluginOptions = options ? options.find(option => option.format === format).plugins.map(plugin => ({value: plugin.title, label: plugin.title})) : [];
 
+    if (pluginOptions.length < 2) {
+      pluginOptions.push({
+        label: 'Get plugins...',
+        value: 'open-plugins'
+      });
+    }
+
     return (
       <div className="container">
         <div className="label">Destination</div>
--- a/renderer/components/editor/options/right.js
+++ b/renderer/components/editor/options/right.js
@@ -13,7 +13,7 @@ class RightOptions extends React.Component {
 
     if (pluginOptions.length < 2) {
       pluginOptions.push({
-        label: 'Get plugins...',
+        label: 'Get plugins…',
         value: 'open-plugins'
       });
     }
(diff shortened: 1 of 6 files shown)


renderer/components/editor/options/right.js
```diff
@@ -11,6 +11,13 @@ class RightOptions extends React.Component {
     const formatOptions = options ? options.map(({format, prettyFormat}) => ({value: format, label: prettyFormat})) : [];
     const pluginOptions = options ? options.find(option => option.format === format).plugins.map(plugin => ({value: plugin.title, label: plugin.title})) : [];
 
+    if (pluginOptions.length < 2) {
+      pluginOptions.push({
+        label: 'Get plugins...',
```
sindresorhus:
```suggestion
        label: 'Get plugins…',
```

renderer/components/editor/options/right.js
```diff
@@ -11,6 +11,13 @@ class RightOptions extends React.Component {
     const formatOptions = options ? options.map(({format, prettyFormat}) => ({value: format, label: prettyFormat})) : [];
     const pluginOptions = options ? options.find(option => option.format === format).plugins.map(plugin => ({value: plugin.title, label: plugin.title})) : [];
 
+    if (pluginOptions.length < 2) {
+      pluginOptions.push({
```
sindresorhus: I think we should have a separator here.

sindresorhus:
In this PR, when I go to the Plugins tab, the "Discover" sub-tab is not highlighted:

Screen Shot 2019-07-08 at 22 27 12

renderer/components/editor/options/right.js
karaggeorge: Seems like there's no native way to add a separator to an html select. I used [this](https://stackoverflow.com/questions/899148/html-select-option-separator) and it shows up like this:
Screenshot 2019-07-08 11 53 27

Ideally, I would change that to use the electron menu, but that would be a bit more of a refactor, since that component is used in a couple of places

sindresorhus: Hmm, that doesn't look very good. We can just skip the separator until we use the Electron API.

karaggeorge:
@sindresorhus Fixed that issue for the tab. I agree about the separator, we can work on switching that to the native one in a separate PR (and when we figure out how to handle screen positioning better)

PR #687, opened 2019-07-08
Save the last used directory

### LINKED ISSUE #659: Export should remember last location
**macOS version:**    18.5.0 (darwin)
**Kap version:**      3.0.0-beta.5

#### Steps to reproduce
Record a video, choose Gif as export target, hit export, it is always in the Kaptures folder.

#### Current behavior
No matter where I saved previously, it starts in the Kaptures folder.

#### Expected behavior
It remembers the previous save location.

#### Workaround
I use this to add gifs to PR's many times a day, and dragging from my desktop is just the easiest way for me to do that outside of always having a finder opened to the kaptures directory, which I don't. As it stands now, I have to remember to change the directory, or forget and hunt around to find that directory. It would be great to offer an option to not always preset to it but instead the last used directory.

karaggeorge:
Fixes #659

### DIFF
--- a/main/export-list.js
+++ b/main/export-list.js
@@ -23,7 +23,7 @@ const Export = require('./export');
 
 const ffmpegPath = util.fixPathForAsarUnpack(ffmpeg.path);
 const showSaveDialog = pify(dialog.showSaveDialog, {errorFirst: false});
-let lastSavedDirectory = null;
+let lastSavedDirectory;
 
 const filterMap = new Map([
   ['mp4', [{name: 'Movies', extensions: ['mp4']}]],
--- a/main/export-list.js
+++ b/main/export-list.js
@@ -2,6 +2,7 @@
 'use strict';
 const {dialog, BrowserWindow} = require('electron');
 const fs = require('fs');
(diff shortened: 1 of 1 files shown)


main/export-list.js
```diff
@@ -22,6 +23,7 @@ const Export = require('./export');
 
 const ffmpegPath = util.fixPathForAsarUnpack(ffmpeg.path);
 const showSaveDialog = pify(dialog.showSaveDialog, {errorFirst: false});
+let lastSavedDir = null;
```
sindresorhus:
```suggestion
let lastSavedDirectory;
```

main/export-list.js
```diff
@@ -22,6 +23,7 @@ const Export = require('./export');
 
 const ffmpegPath = util.fixPathForAsarUnpack(ffmpeg.path);
 const showSaveDialog = pify(dialog.showSaveDialog, {errorFirst: false});
+let lastSavedDirectory = null;
```
sindresorhus: There's no reason to initialize with `null`.
```suggestion
let lastSavedDirectory;
```

PR #688, opened 2019-07-08
Use original file name for exports when possible

### LINKED ISSUE #673: When opening a video and exporting to gif, filename doesn't stay the same
**macOS version:** 10.14.4
**Kap version:** 3.0.0-beta.5

#### Steps to reproduce

- Open video file. e.g. "hello.mp4"
- Select Gif format and press export.

#### Current behaviour
Filename is "Kapture....gif"
#### Expected behaviour

- I expected "hello.gif".
- Usually tools that convert things keep the filename

(Love this feature in the beta BTW! ❤️)

karaggeorge:
Fixes #673 

Screenshot 2019-07-08 11 22 10

### DIFF
--- a/main/export.js
+++ b/main/export.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const path = require('path');
+const {basename, parse} = require('path');
 const PCancelable = require('p-cancelable');
 const moment = require('moment');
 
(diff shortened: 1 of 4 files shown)


main/export.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const path = require('path');
+const {basename, parse} = require('path');
```
sindresorhus: Generally don't use destructuring for `path`. Most of the names are too generic without context and that makes the code harder to read. `parse` could mean so many things, but when you see `path.parse` it's much clearer.

karaggeorge: That makes a lot of sense 👍

PR #689, opened 2019-07-08
Upgrade dependencies

sindresorhus:
I'm upgrading to Electron 5.0.4 and not 5.0.6 (latest) as 5.0.5 and later has a bug that makes the tray icon have half opacity.


renderer/components/action-bar/controls/main.js
```diff
@@ -74,7 +74,8 @@ class Left extends React.Component {
 
 Left.propTypes = {
   toggleAdvanced: PropTypes.elementType.isRequired,
-  selectApp: PropTypes.elementType.isRequired,
+  // TODO: Is this actuall unused? If so, remove it.
+  selectApp: PropTypes.elementType.isRequired, // eslint-disable-line react/no-unused-prop-types
```
sindresorhus: @karaggeorge Can this be removed?

karaggeorge: Yes that can get removed. We moved that logic in the main process since it has to update all the cropper windows, and I guess I forgot to remove it from the props

PR #700, opened 2019-08-07
Allow plugins to overwrite default file format

### LINKED ISSUE #698: Add ability for a plugin to override the file type it receives
Use-case: I'm working on a Kap plugin for [Gifski](https://github.com/sindresorhus/Gifski) and need to present it as a GIF option in the export window, but I actually need the converted MP4 file, not GIF.

`context.filePath()` returns the chosen file format, so we could maybe add the ability to override that. Something like `await context.filePath({fileType: 'mp4'})`.

karaggeorge:
Fixes #698 
Fixes #700

This is the initial implementation for this feature. We'll need to add a `kapVersion` field to the plugins and handle some edge cases before this is ready to go


sindresorhus:
You have to ensure that even though a plugin requests a different format than what the user picked, the original format is still displayed in the UI, for example, in the export list.

main/convert.js
```diff
@@ -184,6 +184,11 @@ const converters = new Map([
 const convertTo = (opts, format) => {
   const outputPath = path.join(tempy.directory(), opts.defaultFileName);
   const converter = converters.get(format);
+
+  if (!converter) {
+    throw new Error('Invalid file format');
```
sindresorhus:
```suggestion
    throw new Error(`Invalid file format: ${format}`);
```

sindresorhus:
You need to document it in https://github.com/wulkano/kap/blob/master/docs/plugins.md

karaggeorge:
This is a minimum working version. As in, I re-used the error UI we had for plugins that need configuration, but I updated the reasoning and disabled the toggle when appropriate. 

In the future, we can look into some more helpful UI, for example checking if the version required is higher than the current one and prompt them to update Kap to the latest version etc

sindresorhus:
I think we should disable the toggle switch when the plugin is `npm link`'d (when it's a symlink).

sindresorhus:
> In the future, we can look into some more helpful UI, for example checking if the version required is higher than the current one and prompt them to update Kap to the latest version etc

Yeah, that's a good follow-up task for other contributors. Can you open an issue about it?

sindresorhus:
A plugin is correctly marked as incompatible when I use `"kapVersion": ">=4.0.0"`, but if I use `"kapVersion": ">=3.0.0-beta.7"`, it's not. When the latest version is `3.0.0-beta.7`.

sindresorhus:
When I export to my plugin using the overridden `fileType`. The item in the export list should not be clickable. Right now it opens the `.mp4` file, which is not wanted, as the user picked GIF and the `.mp4` file is an implementation detail.

sindresorhus:
Otherwise, it worked fine. I've finished my plugin and I'm just waiting for this to be merged.

docs/plugins.md
```diff
@@ -21,6 +21,7 @@ Tip: You can use modern JavaScript features like async/await in your plugin.
 - Use `context.setProgress()` whenever possible to keep the user updated on what's happening. The `.filePath()` method sets its own progress, so you should not do it for that step.
 - The readme should follow the style of [`kap-giphy`](https://github.com/wulkano/kap-giphy).
 - Your plugin must be tested, preferably using [`kap-plugin-test`](https://github.com/SamVerschueren/kap-plugin-test) and [`kap-plugin-mock-context`](https://github.com/samverschueren/kap-plugin-mock-context). [Example](https://github.com/wulkano/kap-giphy/blob/master/test/test.js).
+- If your plugin only supports specific versions of Kap, include a `kapVersion` field in the package.json with a [semver range](https://nodesource.com/blog/semver-a-primer/).
```
sindresorhus: You need to document the `{fileType}` argument.

karaggeorge:
@sindresorhus 

> I think we should disable the toggle switch when the plugin is npm link'd (when it's a symlink).

Is there an easy way we can tell?

> A plugin is correctly marked as incompatible when I use "kapVersion": ">=4.0.0", but if I use "kapVersion": ">=3.0.0-beta.7", it's not. When the latest version is 3.0.0-beta.7.

Isn't that expected? Since `3.0.0-beta.7` is included in `>=3.0.0-beta.7` (satisfies the equality part of it)

> When I export to my plugin using the overridden fileType. The item in the export list should not be clickable. Right now it opens the .mp4 file, which is not wanted, as the user picked GIF and the .mp4 file is an implementation detail.

Should there be an option to overwrite the destination path? So you can provide the path to the final `.gif`? Or just disabled the clicking and drag-n-drop completely for that item?

sindresorhus:
> Is there an easy way we can tell?

Yes, check if the directory is a symlink.

> Isn't that expected? Since 3.0.0-beta.7 is included in >=3.0.0-beta.7 (satisfies the equality part of it)

Copy-paste mistake. Meant to write `3.0.0-beta.6` for the version.

> Should there be an option to overwrite the destination path? So you can provide the path to the final .gif? Or just disabled the clicking and drag-n-drop completely for that item?

For the use-case of the Gifski plugin, it doesn't get the GIF path back, so I would just disable clicking and drop-n-drop. We can add ability to overwrite destination path if the need arises later on.

karaggeorge:
@sindresorhus for the version thing, I tested semver's `satisfies` and it seemed to work. If you tested with the binary of this PR, the version shows up as `3.0.0` which satisfies `3.0.0-beta.x`, but once it's built in the beta the version will actually be `3.0.0-beta.7` and it should work.

Also, I just realized that any version prior to this won't have that check, so the plugins will show up anyway, but won't work

PR #712, opened 2019-09-03
Support ProRes

### LINKED ISSUE #666: Support previewing ProRes
Handle ProRes imports in the Editor.

karaggeorge:
Fixes #666

### DIFF
--- a/main/menus.js
+++ b/main/menus.js
@@ -45,12 +45,12 @@ const openFileItem = {
   click: async () => {
     closeAllCroppers();
 
-    const filePaths = await dialog.showOpenDialogSync({
+    const {canceled, filePaths} = await dialog.showOpenDialog({
       filters: [{name: 'Videos', extensions: supportedVideoExtensions}],
       properties: ['openFile', 'multiSelections']
     });
 
-    if (filePaths) {
+    if (!canceled && filePaths) {
       openFiles(...filePaths);
     }
   }
--- a/main/menus.js
+++ b/main/menus.js
@@ -42,17 +42,17 @@ Workaround:           A workaround for the issue if you've found on. (this will 
 const openFileItem = {
   label: 'Open Video…',
   accelerator: 'Command+O',
-  click: () => {
+  click: async () => {
     closeAllCroppers();
 
-    dialog.showOpenDialogSync({
+    const filePaths = await dialog.showOpenDialogSync({
       filters: [{name: 'Videos', extensions: supportedVideoExtensions}],
-      properties: ['openFile']
-    }, filePaths => {
-      if (filePaths) {
-        openFiles(...filePaths);
-      }
+      properties: ['openFile', 'multiSelections']
     });
+
+    if (filePaths) {
+      openFiles(...filePaths);
+    }
   }
 };
 
--- a/main/utils/encoding.js
+++ b/main/utils/encoding.js
@@ -19,7 +19,7 @@ const getEncoding = async filePath => {
 };
 
 // `ffmpeg -i original.mp4 -vcodec libx264 -crf 27 -preset veryfast -c:a copy output.mp4`
-const convertToH264 = async inputPath => {
+const convertHEVCToH264 = async inputPath => {
   const outputPath = tmp.tmpNameSync({postfix: path.extname(inputPath)});
   track('encoding/converted/hevc');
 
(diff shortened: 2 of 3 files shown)


sindresorhus:
With this, we could now add the `Video codec` preference that we could not add in https://github.com/wulkano/kap/pull/254. @skllcrn ?

The benefit of recording in ProRes would be that it should in theory require a lot less processing power as it doesn't have to compress the video. It can also be useful for people that intend to edit the video after capturing it with Kap for maximum quality.

karaggeorge:
@sindresorhus would the resulting exported videos (trimmed/converted) be in ProRes as well? Or would we switch it to H264 like we have it now? Asking because that might change some of the conversions. `Save Original` would obviously be the ProRes video. 

Also, since ProRes is not compressed at recording time, does that mean that the export would take longer? Or does it trim time off of that as well? (skipping the decoding step)

sindresorhus:
Only “Save Original” would be in ProRes. Export from ProRes to GIF should in theory be faster as ProRes can be decoded faster.

karaggeorge:
So, do we record in ProRes by default? Instead of H264, since it's less processing power, faster export (in theory). Is there any drawbacks to recording in ProRes? 

This way we don't have to create a new setting for it

sindresorhus:
The downside is more disk space. We should probably do some manual benchmarks before making it the default. For example, one downside is that we to convert it before showing it in the editor, so long recordings will make the editor take longer to show.

karaggeorge:
Yeah, even with the fast preset, when I tested it on a big file it takes a couple of seconds. We could show the editor early, with a loader and a message saying "Generating preview" or something like that. I'll run some conversions locally in small/medium/large files, with fast preset or normal (good quality) conversions and send the results

karaggeorge:
Some benchmarking results: https://gist.github.com/karaggeorge/18fe906ea5f1f03c68d758a6e9bf95ff

TL;DR:
Conversions take a while, and the ProRes files are really really big, so imo we should have recording in ProRes under a setting, and definitely come up with a design for the editor while we are generating the preview (6 minute ProRes video took more than a minute to generate)

sindresorhus:
> Conversions take a while, and the ProRes files are really really big, so imo we should have recording in ProRes under a setting

👍 

@skllcrn Are you ok with this?

> and definitely come up with a design for the editor while we are generating the preview

@skllcrn `$ open -a Sketch`

> (6 minute ProRes video took more than a minute to generate)

That is surprising. I imagined it would take much less time. Are we sure we're using the fastest settings with `ffmpeg`? Quality doesn't matter for the preview.

sindresorhus:
Can you try the `ultrafast` preset? https://trac.ffmpeg.org/wiki/Encode/H.264#Preset

karaggeorge:
@sindresorhus 

definitely not the fastest, but the ones we use currently. I'll try `ultrafast`. For `crf` I thought it was 0-51 where 51 was the worst. Should I try 51 instead of 0?

I'll also try the audio and scale options. Scaling it down will also include more complexity, since we use the actual video dimensions for the editor, but we'll have to override them with the not-scaled-down ones

Also, I was still using my computer. Not anything heavy, regular work, but it wasn't the only process running. So it could probably be faster in theory, but a user will probably be doing something else while converting anyway

sindresorhus:
> For crf I thought it was 0-51 where 51 was the worst. Should I try 51 instead of 0?

I misunderstood that option. Yeah, just try. I don't think it will have much impact though.

karaggeorge:
So, I can't use `-c:a copy` because I get an error:
```
[mp4 @ 0x7f91d5801e00] Could not find tag for codec pcm_s16be in stream #1, codec not currently supported in container
Could not write header for output file #0 (incorrect codec parameters ?): Invalid argument
Error initializing output stream 0:0 --
```

But changing `crf` to 51 (worst possible quality) and using `ultrafast` brought a 1 minute video conversion from 10s to 6s, so I guess it improves a bit.

Using the scale down by half brought that down to 4.2s

sindresorhus:
> But changing crf to 51 (worst possible quality) and using ultrafast brought a 1 minute video conversion from 10s to 6s, so I guess it improves a bit.
> Using the scale down by half brought that down to 4.2s

So let's do both of those tricks then.

> We could show the editor early, with a loader and a message saying "Generating preview"

We'll still need this though.

sindresorhus:
> @skllcrn Are you ok with this?

⬆️

skllcrn:
It'd be great to add a settings toggle for this so we could all get a build in our hands and test it out @sindresorhus and @karaggeorge!

sindresorhus:
Closing as this is stale.

PR #713, opened 2019-09-06
Fix progress calculation and add time estimate for conversions

karaggeorge:
Progress calculation was done based on frames completed using an arbitrary fps of 30 and the length of the total video instead of just the trimmed part. That led to sometimes getting a slow progress that would end super fast, or a progress that would go above 100%. 

Refactored the progress to work off of length of video completed vs total length and added time remaining estimation based on that length and the converting speed, both of which are available in `ffmpeg`'s output

### DIFF
--- a/main/export.js
+++ b/main/export.js
@@ -44,7 +44,7 @@ class Export {
       defaultFileName: this.isDefault ? path.basename(this.context.targetFilePath) : this.defaultFileName,
       text: this.text,
       status: this.status,
-      percentage: this.percentage,
+      percentage: this.percentage || 0,
       image: this.image,
       createdAt: this.createdAt,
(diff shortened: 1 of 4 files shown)


main/export.js
```diff
@@ -113,7 +113,7 @@ class Export {
         ...this.exportOptions,
         defaultFileName: fileType ? `${path.parse(this.defaultFileName).name}.${fileType}` : this.defaultFileName,
         inputPath: this.inputPath,
-        onProgress: percentage => this.setProgress('Converting…', percentage)
+        onProgress: (percentage, estimate) => this.setProgress(estimate ? `Converting (${estimate} remaining)` : 'Converting…', percentage)
```
sindresorhus: Can you change this to https://wulkano.slack.com/archives/G20K13PS7/p1567836784038400 ?

PR #717, opened 2019-09-11
Add built-in "Open With" plugin

### LINKED ISSUE #705: Add "Open With" option to the editor
Would be nice if Kap had an "Open With" option in the Editor that would let me open the exported recording in any installed app. That way we could support other apps, even if they don't have a plugin integration. For example, I could export a recording, and then have it opened in Final Cut Pro.

It would look similar to the "Open With" menu in the Finder context menu.

It would be another item here:

Screenshot 2019-08-13 at 22 23 00

And be a submenu that lists apps able to handle the file type. We would have to make a binary module for this, but it's not hard:

```swift
extension NSWorkspace {
	func urlsForAppsThatCanOpen(typeIdentifier: String, role: LSRolesMask = [.viewer, .editor]) -> [URL] {
		let returnValue = LSCopyAllRoleHandlersForContentType(typeIdentifier as CFString, role)?.takeUnretainedValue() as? [String] ?? []
		return returnValue.compactMap { urlForApplication(withBundleIdentifier: $0) }
	}
}
```

And we would need that binary module to also handle opening a specific app: https://developer.apple.com/documentation/appkit/nsworkspace/1528548-open

karaggeorge:
Closes #705 

Also finally fixes https://github.com/wulkano/kap/pull/686#discussion_r301158258

Please test this a lot with exporting with every available option, (un)installing plugins while editor is open and exporting. Basically try to break it in any way. It contains a decent amount of refactor, so I want to make sure it didn't break anything

Built this module for this. Might need some review as well: https://github.com/karaggeorge/mac-open-with


sindresorhus:
macOS 10.15 returns the apps in the correct order, so can you conditionalize https://github.com/karaggeorge/mac-open-with/blob/6bd752b2e10c209043651ab37499430f40745c47/index.js#L13-L17 to only do that on macOS 10.14 and below?

sindresorhus:
You need to filter out Kap and Kap Beta:

Screenshot 2019-09-12 at 11 57 29

main/plugins/open-with-plugin.js
```diff
@@ -0,0 +1,28 @@
...
+const apps = new Map(
+  ['mp4', 'gif', 'apng', 'webm']
+    .map(extension => [
+      extension,
+      getAppsThatOpenExtension.sync(extension).map(app => ({
+        ...app,
+        name: decodeURI(path.parse(app.url).name)
+      }))
+    ])
+    .filter(([_, apps]) => apps.length > 0)
```
sindresorhus: Can you sort `Gifski` to the top? People are likely to want to use it with Kap if they have it installed.

sindresorhus:
You should not show the default in the normal list:

Screenshot 2019-09-12 at 12 06 12

Notice how Chromium is in the list twice.

sindresorhus:
There seems to be no proper sorting of plugins now:

Screenshot 2019-09-12 at 12 10 12
Screenshot 2019-09-12 at 12 10 19

I think "Save to Disk" should be first, then "Copy to Clipboard, then "Open With" and then third-party plugins.

sindresorhus:
> There seems to be no proper sorting of plugins now:

Ah, I forgot, we sort on usage, right?

sindresorhus:
@karaggeorge You need to guard this plugin behind a check. The binary will only work on macOS 10.14.4 and later as the Swift libraries are now bundled in the OS and not in the binary. See: https://github.com/karaggeorge/mac-open-with/pull/1/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9R5

karaggeorge:
Hmm, I'll look into that. Also, I tried to check it but it seems like when the role is `submenu` and not one of `checkbox` or `radio` you can't check it (at least with the Electron API, not sure natively)

karaggeorge:
@sindresorhus should be fixed

sindresorhus:
Looks good :)

sindresorhus:
> Also, I tried to check it but it seems like when the role is submenu and not one of checkbox or radio you can't check it (at least with the Electron API, not sure natively)

You can do it natively, but not very important.

PR #719, opened 2019-09-13
Check if exports are running before quitting Kap

### LINKED ISSUE #715: Confirm quitting if an export is running
During an export, we allow closing the "Export" window and the export continues. But there's no warning when quitting Kap that an export is in progress. I think we should add a prompt. The buttons could be `Quit` and `Show Exports`.

karaggeorge:
Fixes #715

### DIFF
--- a/main/export-list.js
+++ b/main/export-list.js
@@ -1,6 +1,6 @@
 /* eslint-disable array-element-newline */
 'use strict';
-const {dialog, BrowserWindow} = require('electron');
+const {dialog, BrowserWindow, app} = require('electron');
 const fs = require('fs');
 const {dirname} = require('path');
 const {ipcMain: ipc} = require('electron-better-ipc');
@@ -86,6 +86,7 @@ class ExportList {
 
     this.currentExport = this.queue.shift();
     if (this.currentExport.canceled) {
+      delete this.currentExport;
       this._startNext();
       return;
     }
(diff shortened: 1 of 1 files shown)


main/export-list.js
```diff
@@ -261,4 +262,27 @@ const callExportsWindow = (channel, data) => {
...
+      const buttonIndex = dialog.showMessageBoxSync(exportsWindow, {
+        type: 'question',
+        buttons: [
+          'Quit',
+          'Continue'
+        ],
+        defaultId: 1,
+        cancelId: 0,
+        message: 'Do you want to continue exporting?',
+        detail: 'Kap is currently exporting files. If you quit, the export task will be cancelled.'
```
timothyis: Do we go for US English? If so, to be consistent: 
```suggestion
        detail: 'Kap is currently exporting files. If you quit, the export task will be canceled.'
```

sindresorhus: I think `cancelled` is more common in the US too. Even macOS uses `cancelled` for US English, see #715.

main/export-list.js
```diff
@@ -261,4 +262,27 @@ const callExportsWindow = (channel, data) => {
...
+  app.on('before-quit', event => {
+    if (exportList.currentExport) {
+      openExportsWindow();
+      const exportsWindow = getExportsWindow();
+
+      const buttonIndex = dialog.showMessageBoxSync(exportsWindow, {
+        type: 'question',
+        buttons: [
+          'Quit',
+          'Continue'
```
sindresorhus: The buttons are in the wrong order. See the screenshot in #715.

Issue #27, opened 2016-10-06
We need tests


matheuss:
We need some tests:
* [Spectron](https://github.com/electron/spectron) should be used
* [AVA](http://ava.li) should be used

zeke:
https://github.com/electron/electron-api-demos/blob/master/tests/index.js is a pretty good Spectron example to start from.

I tried AVA for a bit but it was quite slow to start up every time because it's doing so much transpilation.

matheuss:
Thanks for the link @zeke – will be needed since I never used Spectron 😅 

Yeah, AVA takes some time to start – but it's awesome 😌 cc @sindresorhus

sindresorhus:
@zeke That should only be the first run. We cache everything. Unless you're using `babel-register`, which circumvents our cache, but I don't see why `babel-register` would be used here. Node.js 6 also improved `require` time significantly, so should help. We're working on a promising solution: https://github.com/avajs/ava/issues/1052.

zeke:
@sindresorhus sweet! I will give AVA another go.

hyalkaf:
I would like to start working on this. Maybe I can start doing tests in chunks and make pull requests for them?

matheuss:
Sounds amazing @hyalkaf 😄 You could write the first ones and then send a `[WIP]` PR and then just keep adding the others 😄

fokusferit:
Any progress here ? 😄 

Otherwise I would try to start on it. Is the decision on AVA now final or Spectron ?

sindresorhus:
@fokusferit No progress. Feel free to take it on. AVA + Spectron. Example here: https://github.com/electron/spectron/blob/master/README.md#with-ava

brodeynewman:
Has any progress been made on this yet? 
This will be my first time contributing to this project.

skllcrn:
That'd be fantastic @brodeynewman! There hasn't been any substantial progress on this, @sindresorhus can comment more on that. You can grab the latest version of Kap off [`master`](https://kap-artifacts.now.sh/master), here's how to get set up: https://github.com/wulkano/kap/blob/master/contributing.md

brodeynewman:
Awesome. I will start on this today then.
Also, do we have a preference on what test libraries / frameworks we use? I usually go with Jest due to Istanbul coverage reporting being built in, and mix Enzyme in for React component testing.

skllcrn:
[...]

PR #721, opened 2019-09-16
added in ava test

dhuang612:
trying from a fresh updated copy of the repo seeing if it can pass continuous integration test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "private": true,
   "main": "main/index.js",
   "scripts": {
-    "test": "xo",
+    "test": "xo && ava --verbose",
     "start": "yarn && electron .",
     "build": "next build renderer && next export renderer",
     "dist": "npm run build && electron-builder",
@@ -74,6 +74,7 @@
     "unstated": "^1.2.0"
   },
   "devDependencies": {
+    "ava": "^2.4.0",
     "babel-eslint": "^10.0.3",
     "electron": "6.0.9",
     "electron-builder": "^21.2.0",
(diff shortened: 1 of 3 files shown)


dhuang612:
still having the same issue of it not running the test at all.
included an env param which didn't resolve anything

dhuang612:
This is pending the circleci update

skllcrn:
This seems like a great starting point @dhuang612, but I'm not sure it fully addresses #27. From what I can tell this adds a test to check if there is a window in addition to dev tools, but that doesn't really tell us much. The tests issue is quite old and would definitively benefit from some more specs.

albinekb:
Since Kap is macOS only, I don't see how a docker image would make tests work? From what I know it's not possible to run the app inside docker? 🤔 The test needs to be ran with macOS?

karaggeorge:
@skllcrn @albinekb @dhuang612 

I think the miss-understanding comes from the issue itself. We need tests, but there's two types of tests. Pure js unit tests and e2e tests.

Spectron is used for e2e tests and you need the app to run in a macOS image, which is not supported by CircleCI right now, and we'd have to set up our own image. However, like @skllcrn said, there's little to no value in writing Spectron tests, as you can't really test functionality past how many windows are open which is not worth the set up.

So, what we are really looking for in adding tests, is adding simple js unit-tests using AVA, to test parts of the app like conversion logic, helper functions and maybe setting up frontend React tests. 

For those, we don't need the whole app to run in an environment, we just need to import those specific js files.

A good start would be setting up tests for our conversion logic in the `main` part of the app

dhuang612:
@karaggeorge 

Thank you, that explains a lot of what I wasn't understanding!

I can start on that and will work to get that completed.
I'll go ahead and close this pr since it's not relevant.

PR #724, opened 2019-09-20
Bug fixes

karaggeorge:
Fixes 3 bugs:

- When reading a plugin's config, if anything goes wrong it would crash and show an empty list. This way at that one will be omitted but at least the rest will show

- When opening the plugin config, it would be an empty window. That was introduced with the `Open With` PR, but fixed with this one

- When installing a plugin, if something went wrong (npm error, or connection), the app wouldn't catch it and not crash, but it remained in the `installing` state with every other plugin disabled. Now it will reset back, with the plugin remaining uninstalled

### DIFF
--- a/main/export-options.js
+++ b/main/export-options.js
@@ -60,7 +60,8 @@ const getExportOptions = () => {
         }
       }
     } catch (error) {
-      console.log(error);
+      const Sentry = require('../utils/sentry');
+      Sentry.captureException(error);
     }
   }
 
--- a/main/export-options.js
+++ b/main/export-options.js
@@ -60,7 +60,7 @@ const getExportOptions = () => {
         }
       }
     } catch (error) {
-      const Sentry = require('../utils/sentry');
+      const Sentry = require('./utils/sentry');
       Sentry.captureException(error);
     }
   }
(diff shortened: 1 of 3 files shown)


main/export-options.js
```diff
@@ -46,17 +46,21 @@ const getExportOptions = () => {
...
+          options.find(option => option.format === format).plugins.push({
+            title: service.title,
+            pluginName: json.name,
+            pluginPath: json.pluginPath,
+            apps: json.name === '_openWith' ? apps.get(format) : undefined
+          });
+        }
       }
+    } catch (error) {
+      console.log(error);
```
sindresorhus: Can you log this to Sentry? So we can catch it and it could help us potentially fix it.

renderer/containers/preferences.js
```diff
@@ -86,6 +86,10 @@ export default class PreferencesContainer extends Container {
         pluginsFromNpm: pluginsFromNpm.filter(p => p.name !== name),
         pluginsInstalled: [plugin, ...pluginsInstalled].sort((a, b) => a.prettyName.localeCompare(b.prettyName))
       });
+    } else {
+      this.setState({
+        pluginBeingInstalled: null
```
sindresorhus: Can we try to use `undefined` whenever possible over `null`? See https://github.com/sindresorhus/meta/issues/7

karaggeorge:
@sindresorhus sorry for the delay on this one, should be ready for another look

PR #742, opened 2019-10-29
Use the `IconMenu` component in the `...` button in the export window

### LINKED ISSUE #559: The `...` button in the export window should use the `IconMenu` component

topalavlad:
This PR fixes #559. I saw that the [previous PR](https://github.com/wulkano/kap/pull/594) was abandoned and closed so I thought I'd take over. 

A few mentions/questions regarding the code I've changed:

- I couldn't find a reason for which `onClick={this.openMenu}` was moved from container to icon in [this PR](https://github.com/wulkano/kap/pull/616/files#diff-bf39c262b039812c1eabc844a553e0d7R24). Unfortunately if the listener is set directly on the svg, we cannot make it fill the parent without making the svg look big
- Since nesting is not supported by styled-jsx, I thought the `title-color` classes would be the best option. Let me know if there's a better one
- Is using `switch-disabled-color` alright or would a custom `disabled-title` color be better?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#559: The `...` button in the export window should use the `IconMenu` component](https://issuehunt.io/repos/65411043/issues/559)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

### DIFF
--- a/renderer/components/exports/export.js
+++ b/renderer/components/exports/export.js
@@ -27,18 +27,23 @@ export default class Export extends React.Component {
   }
 
   openFile = () => {
-    const {filePath, disableOutputActions} = this.props;
-    if (filePath && !disableOutputActions) {
+    const {filePath} = this.props;
+    if (this.isActionable) {
       electron.remote.shell.showItemInFolder(filePath);
     }
   }
 
   onDragStart = event => {
-    const {createdAt, disableOutputActions} = this.props;
+    const {createdAt} = this.props;
     event.preventDefault();
-    if (!disableOutputActions) {
+    if (this.isActionable) {
       electron.ipcRenderer.send('drag-export', createdAt);
     }
+  }
+
+  get isActionable() {
+    const {filePath, disableOutputActions} = this.props;
+    return filePath && !disableOutputActions;
   }
 
   render() {
@@ -55,8 +60,7 @@ export default class Export extends React.Component {
     const cancelable = status === 'waiting' || status === 'processing';
     const fileNameClassName = classNames({
       title: true,
-      'title-color': !cancelable,
-      'title-color-disabled': cancelable
+      disabled: !this.isActionable
     });
     return (
       <div draggable className="export-container" onClick={this.openFile} onDragStart={this.onDragStart}>
@@ -66,8 +70,7 @@ export default class Export extends React.Component {
             {
               cancelable ?
                 <div className="icon" onClick={cancel}>
-                  <CancelIcon className="icon" fill="white"
-                    hoverFill="white" activeFill="white"/>
+                  <CancelIcon fill="white" hoverFill="white" activeFill="white"/>
                 </div> :
                 <IconMenu fillParent icon={MoreIcon} fill="white"
                   hoverFill="white" activeFill="white"
@@ -139,13 +142,10 @@ export default class Export extends React.Component {
             white-space: nowrap;
             overflow: hidden;
             text-overflow: ellipsis;
-          }
-
-          .title-color {
             color: var(--title-color);
           }
 
-          .title-color-disabled {
+          .disabled {
             color: var(--switch-disabled-color);
           }
 
--- a/renderer/components/exports/export.js
+++ b/renderer/components/exports/export.js
@@ -1,6 +1,7 @@
 import electron from 'electron';
 import PropTypes from 'prop-types';
 import React from 'react';
+import classNames from 'classnames';
 
 import IconMenu from '../icon-menu';
 import {CancelIcon, MoreIcon} from '../../vectors';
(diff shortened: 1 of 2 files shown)


renderer/components/exports/export.js
```diff
@@ -75,7 +80,9 @@ export default class Export extends React.Component {
           </div>
         </div>
         <div className="details">
-          <div className="title">{defaultFileName}</div>
+          <div className={'title ' + (status === 'processing' ? 'title-color-disabled' : 'title-color')}>
```
karaggeorge: Can we use classnames for this?

topalavlad: Sorry, not sure I follow. Do you mean remove `title` class and duplicate its properties in `title-color` and `title-color-disabled`?

topalavlad: Oh, you mean using `classnames` attribute instead of `classname`, right?

albinekb: I think this library is what he's talking about: https://www.npmjs.com/package/classnames

renderer/components/exports/export.js
```diff
@@ -52,16 +53,25 @@ export default class Export extends React.Component {
     const {menu} = this.state;
 
     const cancelable = status === 'waiting' || status === 'processing';
-
+    const fileNameClassName = classNames({
+      title: true,
+      'title-color': !cancelable,
+      'title-color-disabled': cancelable
```
karaggeorge: I don't think we need 3 classes, since the two are mutually exclusive. Could be:
```
{
	title: true
	disabled: cancelable
}
```

Then, add the default title color to the `title` class and the disabled color on the `.title.disabled` class

karaggeorge: Also, not sure `cancelable` is the right check here (might be in error state, so cancelable is `false`, but this should still be disabled). I think we need a new variable called `isActionable`. The value should be set to the same thing as line 31:

`const isActionable = filePath && !disableOutputActions`

Might even be worth abstracting it into a getter on the class, so we can check as `this.isActionable`

```
get isActionable() {
	const {filePath, disableOutputActions} = this.props;
	return filePath && !disableOutputActions;
}
```

And then we can use it here, in the `openFile` method and the `onDragStart` as well (which is missing the `filePath` check but it should probably be there)

renderer/components/exports/export.js
```diff
@@ -52,16 +53,25 @@ export default class Export extends React.Component {
...
       <div draggable className="export-container" onClick={this.openFile} onDragStart={this.onDragStart}>
         <div className="thumbnail">
           <div className="overlay"/>
           <div className="icon" onClick={stopPropagation}>
             {
               cancelable ?
-                <CancelIcon fill="white" hoverFill="white" activeFill="white" onClick={cancel}/> :
-                <IconMenu icon={MoreIcon} fill="white" hoverFill="white" activeFill="white" onOpen={menu && menu.popup}/>
+                <div className="icon" onClick={cancel}>
+                  <CancelIcon className="icon" fill="white"
```
karaggeorge:
```suggestion
                  <CancelIcon fill="white"
```

Does this need that class? Since it's on the parent

topalavlad: Not needed. I'll remove it.

renderer/components/exports/export.js
topalavlad: Thanks, this is a much better option. Canceled titles also stay greyed out now.

PR #749, opened 2019-11-06
Support Catalina

### LINKED ISSUE #743: macOS Catalina
I just installed using brew cask install kap or any other way, i keep getting 

Record Error
could not start recording within 5 seconds. 

then i get this

```
Unhandled Promise Rejection
Error: Call `.startRecording()` first
    at Aperture.stopRecording (/Applications/Kap.app/Contents/Resources/app.asar/node_modules/aperture/index.js:141:13)
    at EventEmitter.<anonymous> (file:///Applications/Kap.app/Contents/Resources/app.asar/dist/renderer/js/main.js:207:39)
    at Generator.next (<anonymous>)
    at step (file:///Applications/Kap.app/Contents/Resources/app.asar/dist/renderer/js/main.js:39:191)
    at file:///Applications/Kap.app/Contents/Resources/app.asar/dist/renderer/js/main.js:39:437
    at Promise (<anonymous>)
    at EventEmitter.<anonymous> (file:///Applications/Kap.app/Contents/Resources/app.asar/dist/renderer/js/main.js:39:99)
    at EventEmitter.stopRecording (file:///Applications/Kap.app/Contents/Resources/app.asar/dist/renderer/js/main.js:223:20)
```

karaggeorge:
Fixes #743 

If you are already on Catalina and have granted the right permissions you can't test the initial flow. To do so run this on your terminal to reset your permissions:

`tccutil reset ScreenCapture com.wulkano.kap`

If you also want to test the microphone ones, you can run this:

`tccutil reset Microphone com.wulkano.kap`

Built a small package to handle it: https://github.com/karaggeorge/mac-screen-capture-permissions

So far, this was the only issue I found with running Kap on Catalina. If anything else comes up, we can handle it here as well

### DIFF
--- a/main/common/system-permissions.js
+++ b/main/common/system-permissions.js
@@ -1,8 +1,7 @@
 const {systemPreferences, shell, dialog} = require('electron');
+const {hasScreenCapturePermission, hasPromptedForPermission} = require('mac-screen-capture-permissions');
 
-const getMicrophoneAccess = () => systemPreferences.getMediaAccessStatus('microphone');
-
-const promptSystemPreferences = async ({hasAsked} = {}) => {
+const promptSystemPreferences = opts => async ({hasAsked} = {}) => {
   if (hasAsked) {
     return false;
   }
@@ -11,19 +10,31 @@ const promptSystemPreferences = async ({hasAsked} = {}) => {
     type: 'warning',
     buttons: ['Open System Preferences', 'Cancel'],
     defaultId: 0,
-    message: 'Kap cannot access the microphone.',
-    detail: 'Kap requires microphone access to be able to record audio. You can grant this in the System Preferences. Afterwards, relaunch Kap for the changes to take effect.',
+    message: opts.message,
+    detail: opts.detail,
     cancelId: 1
   });
 
   if (response === 0) {
-    openSystemPreferences();
+    openSystemPreferences(opts.systemPreferencesPath);
   }
 
   return false;
 };
 
-const ensureMicrophonePermissions = async (fallback = promptSystemPreferences) => {
+const openSystemPreferences = path => shell.openExternal(`x-apple.systempreferences:com.apple.preference.security?${path}`);
+
+// Microphone
+
+const getMicrophoneAccess = () => systemPreferences.getMediaAccessStatus('microphone');
+
+const microphoneFallback = promptSystemPreferences({
+  message: 'Kap cannot access the microphone.',
+  detail: 'Kap requires microphone access to be able to record audio. You can grant this in the System Preferences. Afterwards, relaunch Kap for the changes to take effect.',
+  systemPreferencesPath: 'Privacy_Microphone'
+});
+
+const ensureMicrophonePermissions = async (fallback = microphoneFallback) => {
   const access = getMicrophoneAccess();
 
   if (access === 'granted') {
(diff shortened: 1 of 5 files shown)


main/common/system-permissions.js
```diff
@@ -11,19 +10,31 @@ const promptSystemPreferences = async ({hasAsked} = {}) => {
...
   if (response === 0) {
-    openSystemPreferences();
+    openSystemPreferences(opts.systemPreferencesPath);
   }
 
   return false;
 };
 
-const ensureMicrophonePermissions = async (fallback = promptSystemPreferences) => {
+const openSystemPreferences = path => shell.openExternal(`x-apple.systempreferences:com.apple.preference.security?${path}`);
```
sindresorhus: Your `mac-screen-capture-permissions` should ideally provide a method to open the correct preferences pane.

Ideally, `electron-util` would have a utility method to open system preferences, which could be used in your package.

main/common/system-permissions.js
```diff
@@ -1,8 +1,7 @@
 const {systemPreferences, shell, dialog} = require('electron');
+const {hasScreenCapturePermission, hasPromptedForPermission} = require('mac-screen-capture-permissions');
 
-const getMicrophoneAccess = () => systemPreferences.getMediaAccessStatus('microphone');
-
-const promptSystemPreferences = async ({hasAsked} = {}) => {
+const promptSystemPreferences = opts => async ({hasAsked} = {}) => {
```
sindresorhus:
```suggestion
const promptSystemPreferences = options => async ({hasAsked} = {}) => {
```

sindresorhus:
I reset all permissions (`tccutil reset All com.wulkano.kap`) and then ran `npm start` and Kap shows this:

Screenshot 2019-11-09 at 12 55 04

But I think you intended to show the system permission dialog instead? The above dialog is for when the user has already declined the system permission before.

sindresorhus:
Also, instead of asking the user to relaunch Kap, we should quit Kap when the user clicks "Open System Preferences" and instead tell the user to just *launch* Kap (not relaunch). That would prevent the user from forgetting to relaunch Kap.

main/common/system-permissions.js
karaggeorge: I'll open a PR to add it to electron-util, you're right it should live there

karaggeorge:
Surprisingly, the screen recorder permission, unlike the rest, still works even without restarting (even though the system dialog asks you to restart), but yeah I think it's good practice to quit, so they don't have to do it later. I'll add that.

For the dialog thing, there's no way to check wether or not the prompt has been shown. We just know it gets shown the first time, so the package uses a similar logic like `electron-util` does for the first time the permissions was asked for.

When you reset it manually with `tccutil`, the package has no way of knowing, so it assumes you have asked before and shows the dialog anyway. I think that is the better way to handle the edge case

main/common/system-permissions.js
karaggeorge: Opened a PR: https://github.com/sindresorhus/electron-util/pull/23
After that is merged and released, I can make the updates to the package, then release and then I can update the dependency here

We could keep it as is, until that is done, or if you have some time to review that PR, I can update and bump the dependency in this PR

sindresorhus:
@karaggeorge I think you misunderstood me. What I'm saying is that I *never* get the system prompt, even after running `tccutil reset ScreenCapture com.wulkano.kap` and deleting `.has-app-requested-screen-capture-permissions`. I only get our prompt to open system preferences, which is useless, as Kap is not in the list yet, since the system prompt has not been shown.

karaggeorge:
@sindresorhus Ah, I see. I just tested it and it seemed to work for me. I think what happens is that when you're running it locally (`yarn start`) it doesn't actually run with the bundle identifier. It runs as `Electron` or `Iterm`. If you package it and then use the `tccutil` with `com.wulkano.kap`, then I believe it should work. (at least it did when I just tried it with that build)

sindresorhus:
Ah yeah. That's it! Maybe worth adding a tip about that to https://github.com/wulkano/kap/blob/master/maintaining.md ?

For me it was `tccutil reset ScreenCapture com.googlecode.iterm2`.

sindresorhus:
Looks good :)

PR #759, opened 2019-11-22
Use `electron-builder-notarize` package

karaggeorge:


sindresorhus:
I think you forgot to remove `notarize.js`.

karaggeorge:
Still testing it out. Haven't gotten it to work yet

karaggeorge:
Checked that this work, by turning on Signing on PRs. I have turned it off again, so it won't sign/notarize the commit here, but it should work in master. Notarized build from this PR is [here](https://1436-65411043-gh.circle-artifacts.com/0/Users/distiller/project/dist/Kap.dmg)

PR #763, opened 2019-11-26
Updates to microphone permissions and plugins

### LINKED ISSUE #760: Enabling audio recording crashes the app
**macOS version:** 10.14.5
**Kap version:** 3.0.0

#### Steps to reproduce

Install the Open preferences, and enable the record audio option. App crash.

You can also manually edit your `config` at `~/Library/Application Support/Kap Beta/config.json` to set `recordAudio: true`. Opening Kap after saving this updated config crashes the app. (Note - after attempting to enable audio via preferences, `recordAudio` in the config is not set to `true`, so this could be a separate issue?).

#### Current behaviour

Selecting the record audio option in preferences crashes the app. 

#### Expected behaviour

Selecting the record audio option in preferences should work and not crash the app.

#### Workaround

Run Kap by installing the latest master branch (source).

#### Misc info

Not related to this issue specifically, but related to how I found it. When you are able to successfully enable audio recording (via install from source) - capture a recording of your screen (with audio), and attempt to export to WebM, it reproduces: https://github.com/wulkano/kap/issues/681 and https://github.com/wulkano/kap/issues/474

karaggeorge:
Fixes #760 

Not sure what was causing it, but I've added a lot of checks and safeguards around microphone permissions and usage. I was able to reproduce locally and now I can't. I'll also ping people to test the packaged version.

Also tries to fix the issues with the plugins. Turns out if you `npm prune` it removes all of the node_modules, so Kap can't actually require the plugins. [Yarn's docs](https://yarnpkg.com/lang/en/docs/cli/prune/) say that `yarn install` will automatically prune extra dependencies, so I've replaced the prune with that. (also removing npm as a dependency).

### DIFF
--- a/main/preferences.js
+++ b/main/preferences.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const {BrowserWindow, ipcMain} = require('electron');
+const {BrowserWindow} = require('electron');
 const pEvent = require('p-event');
 
 const {ipcMain: ipc} = require('electron-better-ipc');
@@ -52,9 +52,13 @@ const openPrefsWindow = async options => {
   }
 
   ipc.callRenderer(prefsWindow, 'mount');
-  await pEvent(ipcMain, 'preferences-ready');
-  prefsWindow.show();
-  return prefsWindow;
+
+  return new Promise(resolve => {
+    ipc.answerRenderer('preferences-ready', () => {
+      prefsWindow.show();
+      resolve(prefsWindow);
+    });
+  });
 };
 
 const closePrefsWindow = () => {
--- a/main/preferences.js
+++ b/main/preferences.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const {BrowserWindow} = require('electron');
+const {promisify} = require('util');
 const pEvent = require('p-event');
 
 const {ipcMain: ipc} = require('electron-better-ipc');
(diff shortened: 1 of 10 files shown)


karaggeorge:
I still get this issue when I first download/install the packaged version... might be related to https://github.com/electron/electron/issues/20498

karaggeorge:
Update, the issue we were having was this: https://github.com/electron/electron/issues/19307

Adding the extra entry in the entitlements did the trick for me. Waiting for others to test, to make sure it worked

main/preferences.js
```diff
@@ -52,9 +52,13 @@ const openPrefsWindow = async options => {
   }
 
   ipc.callRenderer(prefsWindow, 'mount');
-  await pEvent(ipcMain, 'preferences-ready');
```
sindresorhus: Why not just `await pEvent(ipc.answerRenderer, 'preferences-ready');` ?

karaggeorge: Wait, is `answerRenderer` also an event emitter? Not just a function that accepts channel etc? Only reason I removed the `pEvent` was because it works with event emitters

karaggeorge: I could try `await pEvent(ipc, eventName)` but I'd have to use the util to generate the event name the way it's done in electron-better-ipc

sindresorhus: I was confused. I meant that it could be promisified with `util.promisify`.

karaggeorge: Ah, makes sense: 2146d1f

PR #764, opened 2019-11-27
Ensure dock is showing before showing dialogs

### LINKED ISSUE #758: "Open Video" bugs
1. Click the Kap menu bar icon, then in the floating controls view, click `...` and then select "Open Video". The open panel shows for a second and then disappears.
2. If I drag and drop a video on the Kap menu bar icon, the destination dropdowns in the editor are empty:  Screenshot 2019-11-22 at 14 49 30
3. "Open Video" is missing in the "File" menu in the Kap editor.

karaggeorge:
Fixes #758 

> Click the Kap menu bar icon, then in the floating controls view, click ... and then select "Open Video". The open panel shows for a second and then disappears.

So, this was the main issue. I've noticed this with any electron dialogs. I narrowed it down, to it happening where the dock is not showing and the dialog is not attached to a window (similar to the app not being key). Same went for the permissions dialogs. 

Added a utility to ensure we show the dock before we open the dialog, and then hide it (if it wasn't showing already)

> If I drag and drop a video on the Kap menu bar icon, the destination dropdowns in the editor are empty

This was an issue with yarn fixed in #763  

> "Open Video" is missing in the "File" menu in the Kap editor.

I can't reproduce this 🤔 When I'm in an editor, when I click File, I see the Open Video option:

Screen Shot 2019-11-26 at 11 39 50 PM

Branched off of #763, which fixes most of the weird issues relating to plugins

### DIFF
--- a/main/menus.js
+++ b/main/menus.js
@@ -4,6 +4,7 @@ const os = require('os');
 const {Menu, app, dialog, BrowserWindow} = require('electron');
 const {openNewGitHubIssue, appMenu} = require('electron-util');
 const {ipcMain: ipc} = require('electron-better-ipc');
+const delay = require('delay');
 
 const {supportedVideoExtensions} = require('./common/constants');
 const {ensureDockIsShowing} = require('./utils/dock');
@@ -45,6 +46,8 @@ const openFileItem = {
   accelerator: 'Command+O',
   click: async () => {
     closeAllCroppers();
+
+    await delay(200);
 
     await ensureDockIsShowing(async () => {
       const {canceled, filePaths} = await dialog.showOpenDialog({
--- a/main/menus.js
+++ b/main/menus.js
@@ -6,6 +6,7 @@ const {openNewGitHubIssue, appMenu} = require('electron-util');
 const {ipcMain: ipc} = require('electron-better-ipc');
 
 const {supportedVideoExtensions} = require('./common/constants');
+const {ensureDockIsShowing} = require('./utils/dock');
 const {openPrefsWindow} = require('./preferences');
 const {openExportsWindow} = require('./exports');
 const {openAboutWindow} = require('./about');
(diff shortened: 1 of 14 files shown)


main/menus.js
```diff
@@ -45,14 +46,16 @@ const openFileItem = {
   click: async () => {
     closeAllCroppers();
 
-    const {canceled, filePaths} = await dialog.showOpenDialog({
-      filters: [{name: 'Videos', extensions: supportedVideoExtensions}],
-      properties: ['openFile', 'multiSelections']
-    });
+    ensureDockIsShowing(async () => {
```
sindresorhus:
```suggestion
    await ensureDockIsShowing(async () => {
```

It doesn't matter in this case, but if you add something after this block, it could be a problem.

sindresorhus:
> Click the Kap menu bar icon, then in the floating controls view, click ... and then select "Open Video". The open panel shows for a second and then disappears.

This happens for me.

karaggeorge:
Hmm, I can't reproduce, but I'll try to see if the artifact is any different

karaggeorge:
@sindresorhus 

For the latest artifact in this PR, when I follow the steps, the open window dialog shows behind my last active app (Kap activates, shows the dialog, then deactivates so the dialog goes behind that last active app) but it doesn't completely go away like it used to. I can try making it so we close the croppers after we've shown the dock, so maybe it doesn't lose focus in the meantime. Hopefully that will keep it at the top.

karaggeorge:
Okay, so I was able to reproduce the issue when running in the terminal, but not the packaged app, which was a small tell. I figured that the issue was that `closeAllCroppers` for some reason takes a bit longer to finish (the `window.destroy()` is not async, but I don't think it completes all events before returning. So, after closing all the croppers, we have a `window-all-closed` event that hides the dock. That had a timing issue with showing the dock for the open dialog which was causing the window to disappear sometimes. Similar issue we had with the about window. Used the same solution as we had there, adding a small delay after closing the croppers, which seems to work consistently for me both when running locally and packaged. Also the window now correctly stays on top

sindresorhus:
Can you fix the merge conflict?

sindresorhus:
> I figured that the issue was that closeAllCroppers for some reason takes a bit longer to finish

Yes, everything is a bit slower in dev as everything runs in dev mode, including React and Next.

> I figured that the issue was that closeAllCroppers for some reason takes a bit longer to finish (the window.destroy() is not async, but I don't think it completes all events before returning.

Worth opening an Electron issue?

> adding a small delay after closing the croppers, which seems to work consistently for me

👍 Yup. Works well now.

karaggeorge:
@sindresorhus Should be fixed 👍 

> Worth opening an Electron issue?

I don't know. The docs don't explicitly say that when you call `window.destroy()` the method doesn't return until all `app` events are called and finished. So I guess it might be async by nature (not the method itself, but the logic) which is causing the race issue we had here

sindresorhus:
> I don't know. The docs don't explicitly say that when you call window.destroy() the method doesn't return until all app events are called and finished. So I guess it might be async by nature (not the method itself, but the logic) which is causing the race issue we had here

I would argue that is a bug then, they either need to:
1. Better document how it works.
2. Fix the race condition.
3. Make the method return a promise.

PR #767, opened 2019-11-28
Fix plugin usage history

### LINKED ISSUE #766: Remember last used export option
**Feedback/Feature request**

I just started using the new version of Kap, which is great. I realised one thing that's different than the old version.

It would be great if it remembered the Save setting. I always choose **Save to Disk**, but the new version of Kap doesn't remember that and always shows **Copy to clipboard** instead. Would be great if it remembered what you last used.

Screenshot 2019-11-27 at 20 57 56
**Kap version:**      3.0.0

Thanks!

karaggeorge:
Fix #766 

At some point we renamed the `name` field on the plugins to `pluginName` but forgot to update it here. So it was updating the `undefined` plugin every time.

Stuff like this makes me wonder if it's worth looking into slowly converting this to TypeScript

### DIFF
--- a/renderer/containers/editor.js
+++ b/renderer/containers/editor.js
@@ -207,6 +207,6 @@ export default class EditorContainer extends Container {
     };
 
     ipc.callMain('export', data);
-    ipc.callMain('update-usage', {format, plugin: plugin.name});
+    ipc.callMain('update-usage', {format, plugin: plugin.pluginName});
   }
 }


sindresorhus:
> Stuff like this makes me wonder if it's worth looking into slowly converting this to TypeScript

It would indeed give us more confidence, but it would also make it more difficult to contribute to Kap and [TypeScript has a lot of problems](https://twitter.com/sindresorhus/status/1197849204137185280), especially with subpar external type definition, which is especially true for the Electron and React type definitions.

Issue #624, opened 2019-02-05
Add "Tips" window


sindresorhus:
We should have a tips window where we could document handy things that are not immediately apparent to all users. We could add a menu item to the "Help" menu and to the menu bar item menu. We could also show the Tips window the first time Kap is launched as a welcome window.

Some tips ideas:
- Resize from center in the cropper by pressing <kbd>Option</kbd>.
- Resize with locked aspect ratio in the cropper by pressing <kbd>Shift</kbd>.
- Resize from center with locked aspect ratio in the cropper by pressing <kbd>Shift</kbd>+<kbd>Option</kbd>.
- Press up/down arrow keys in input fields to increase/decrease the value. And the <kbd>Shift</kbd> modifier to increase/decrease by 10.
- Edit and convert an existing video by dragging it to the menu bar icon (I bet very few knows about this one) or clicking "Open video" in the menu bar item menu.
- Extending functionality with plugins.
- Clicking the exported item in the "Export" window, reveals it in Finder.

Any other suggestions?

karaggeorge:
@skllcrn if you can come up with a design and content for this I can start adding it

skllcrn:
Sounds like the contents of that window could be served via a web page? That way it'd be accessible via e.g. help.getkap.co too.

karaggeorge:
Yeah, and then that window would just render the contents of that link

PR #769, opened 2019-11-28
Evaluate math expressions in the editor controls

### LINKED ISSUE #234: Allow configurable scaling so Retina displays don't generate enormous files
I tend to create simple gifs for work to illustrate a bug or UX issue I run into on the software we're developing.  Although I use a retina-display Macbook Pro, the resulting gif doesn't need to be ultra crisp and high-resolution.  It merely needs to get the point across.

I tend to use kap to create a gif, and then immediately open that gif up in Photoshop to remove frames from the beginning and end, then scale it to 50% to save on space.  This will often turn a 5mb gif into a ~1mb gif.  It would be great if I didn't need to use photoshop for this process (large gifs tend to create gigantic – multi-gigabite! – scratch disc files on my hard drive as well).

karaggeorge:
Closes #234

@skllcrn will love this one

Kapture 2019-11-28 at 12 53 48

### DIFF
--- a/renderer/containers/editor.js
+++ b/renderer/containers/editor.js
@@ -67,7 +67,7 @@ export default class EditorContainer extends Container {
           updates.width = val;
         }
 
-        updates.height = Math.round(updates.width / ratio);
+        updates.height = Math.floor(updates.width / ratio);
       } else {
         const min = Math.max(1, Math.ceil(1 / ratio));
 
@@ -83,7 +83,7 @@ export default class EditorContainer extends Container {
           updates.height = val;
         }
 
-        updates.width = Math.round(updates.height * ratio);
+        updates.width = Math.ceil(updates.height * ratio);
       }
     } else if (name === 'width') {
       shake(currentTarget, {className: 'shake-left'});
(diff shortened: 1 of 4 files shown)


skllcrn:
You're something else.

renderer/containers/editor.js
```diff
@@ -37,8 +38,21 @@ export default class EditorContainer extends Container {
...
+      this.setState({[name]: value, lastValid: {width, height, ...lastValid}});
+      return;
+    }
+
+    let parsedValue;
+    try {
+      parsedValue = stringMath(value);
+    } catch {
+      parsedValue = null;
+    }
```
sindresorhus: Let's prefer `undefined` over `null`.

```js
} catch {}
```

sindresorhus:
I think this should be documented somewhere. Maybe in a "Tip" section in the readme for now? At least until we have https://github.com/wulkano/kap/issues/624.

sindresorhus:
Not related to this change, but I think I found a bug in the inputs, if I write 200, tab to the next and then tab back, it changes.

karaggeorge:
> Not related to this change, but I think I found a bug in the inputs, if I write 200, tab to the next and then tab back, it changes.

I've noticed that too, but I don't know what the expected behavior is. It rounds the dimensions when it does the Math, and rounds the other way again. Maybe if I replaced `Math.round` with `Math.floor`? Not sure if that would guarantee it doesn't change backwards

sindresorhus:
https://github.com/wulkano/kap/pull/769#issuecomment-559647107 still needs to be done, but can be done later.

PR #779, opened 2019-12-20
Ability to record keys pressed

### LINKED ISSUE #419: Record pressed keys during recording
I know it can be done in [Mac OS Sierra](https://superuser.com/a/1239253) but it has some limitations like it does not show Alphabet keys + Tab & many more. 

I don't know if its in the scope of this project but it would be a great addition. 

My use case was for issues like [this](https://github.com/Microsoft/vscode/issues/45381).

Checkout the GIF in the above mentioned issue that the keystrokes that are recorded are not proper which are from Mac OS Keyboard as mentioned in the above [SuperUser answer](https://superuser.com/a/1239253)

karaggeorge:
Closes #419

### DIFF
--- a/renderer/components/preferences/categories/general.js
+++ b/renderer/components/preferences/categories/general.js
@@ -50,11 +50,13 @@ class General extends React.Component {
       setAudioInputDeviceId,
       audioDevices,
       recordAudio,
+      keyCast,
       pickKapturesDir,
       setOpenOnStartup,
       cropperShortcut,
       updateShortcut,
       toggleShortcuts,
+      toggleKeyCast,
       category
     } = this.props;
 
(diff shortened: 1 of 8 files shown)


renderer/components/preferences/categories/general.js
```diff
@@ -136,6 +138,14 @@ class General extends React.Component {
         >
           <Switch tabIndex={tabIndex} checked={doNotDisturb} onClick={() => toggleSetting('doNotDisturb')}/>
         </Item>
+        <Item
+          key="keyCast"
+          title="Key Cast"
+          subtitle="Show keys pressed while recording"
```
sindresorhus:
```suggestion
          subtitle="Show pressed keys while recording"
```

sindresorhus:
Do you plan to implement support for showing the key view inside the cropped view in this PR or later?

karaggeorge:
Hmm, I wasn't planning on it, because then that comes with edge cases like what if the window displaying the keys doesn't fit the cropper.

Easiest thing I can think of is replacing the logic for the initial position (center x on display and ~80% down on y axis), with the ability to pass in a CGRect, which will default to the screen's frame. So if you don't pass anything works as it is, if you do it'll use that instead. And if the cropper is too small, the user is in charge of moving it away (so no edge cases handled on the library side)

sindresorhus:
> Hmm, I wasn't planning on it, because then that comes with edge cases like what if the window displaying the keys doesn't fit the cropper.

I don't think that's an edge-case worth caring about.

> Easiest thing I can think of is replacing the logic for the initial position (center x on display and ~80% down on y axis), with the ability to pass in a CGRect, which will default to the screen's frame. So if you don't pass anything works as it is, if you do it'll use that instead. And if the cropper is too small, the user is in charge of moving it away (so no edge cases handled on the library side)

👍

Issue #784, opened 2020-01-13
Mouse clicks are NOT highlighted


xstast24:
**macOS version:** 10.13.6
**Kap version:** 3.0.0 & 3.0.1

#### Steps to reproduce
Install Kap on 10.13, open Preferences or capture a clip.
#### Current behaviour
Mouse click highlighting (_highlightClicks_) does not work - clicks are not displayed in the video.
'Show cursor' option is not even displayed in _Preferences_ (cmd+,). However, it is present on macOS 10.15.2, so this bug seems to be OS-specific.
#### Expected behaviour
Option to show cursor should be present in _Preferences_. Mouse clicks should be visible in captured videos.
#### Workaround
Don't know any...  Tried configuring 'highlightClicks' option manually in _~/Library/Preferences/com.wulkano.kap.plist_, but it didn't work.

mootari:
Looks like the setting is only available in 10.14 and greater:
https://github.com/wulkano/kap/blob/bcb9781792b576e281b36a4b185f41199f1d2ca8/renderer/components/preferences/categories/general.js#L25-L29

https://github.com/wulkano/kap/blob/bcb9781792b576e281b36a4b185f41199f1d2ca8/renderer/components/preferences/categories/general.js#L97-L105

The change was introduced in #643.

pettomartino:
Is this just allow from `10.13` ? Can I open a PR for this ?

pettomartino:
![Kapture 2020-01-24 at 13 46 37](https://user-images.githubusercontent.com/101433/73086771-035fd600-3eb0-11ea-8963-60858de97a1c.gif)
![image](https://user-images.githubusercontent.com/101433/73086812-15da0f80-3eb0-11ea-9f5a-948cc01d6d4b.png)
![Kapture 2020-01-24 at 13 47 55](https://user-images.githubusercontent.com/101433/73086841-25595880-3eb0-11ea-9792-bc82904b191f.gif)

I changed that and worked fine. If you guys think it's fine I will open a PR

tonisives:
Clicks are not highlighted for me in 10.15.6.

I checked all the privacy settings and Kap is enabled there.

Highlight clicks checkbox is enabled. I also tried switching it off and on again.

<img width="477" alt="Screen Shot 2020-08-10 at 19 13 35" src="https://user-images.githubusercontent.com/1083534/89781599-dae2b500-db3d-11ea-9c3d-958efe85e8e0.png">

sprazzeus:
[...]

Issue #477, opened 2018-06-11
Hide cursor not working on High Sierra 10.13.5


equinusocio:
**macOS Version:** 10.13.5
**Kap Version:** 2.2.0

#### Steps to reproduce
Uncheck the **show mouse cursor** toggle

#### Current behaviour
Mouse cursor is recorded

#### Expected behaviour
Mouse cursor should be hidden from records

albinekb:
Can't reproduce on `10.13.4 (17E202)`

equinusocio:
Here a gif recorded with kap
![kapture 2018-06-11 at 17 54 49](https://user-images.githubusercontent.com/10454741/41242739-a1dde7dc-6da0-11e8-9609-0adc403ef48e.gif)

albinekb:
I'm updating to 10.13.5 now and will see if I can reproduce and check what's up here 👍

albinekb:
Must be a bug with 10.13.5 then? 🤔 
https://github.com/wulkano/aperture/blob/master/swift/aperture/Recorder.swift#L34-L43
These are the lines that disables the cursor, using Apple's swift API.

equinusocio:
https://developer.apple.com/documentation/avfoundation/avcapturescreeninput/1385601-capturescursor

Maybe this link will help you. I don't know if they changed something
https://developer.apple.com/documentation/avfoundation/avcapturescreeninput/1385601-capturescursor

albinekb:
![image](https://user-images.githubusercontent.com/5027156/41243291-08b6cf40-6da2-11e8-97a8-1a5dff03b498.png)
Yeah, we're sending false to it, so maybe a bug that got introduced as part of Mojave screen capture built in? 🤔 Don't know. Need to dig deeper.

equinusocio:
Damn. So i need to wait Apple to fix this (even if they will know about it :D)

albinekb:
I've filed a bug report with apple... Don't think there's anymore we could do here :/

equinusocio:
@albinekb There is a public link?

sindresorhus:
@albinekb Make sure you mirror it on https://openradar.appspot.com and paste a link here ;)

albinekb:
https://openradar.appspot.com/radar?id=4935854782939136

equinusocio:
Thank you guys
-- 
Mattia Astorino

Il giorno 11 giugno 2018 @ 18:31:12, Albin Ekblom (notifications@github.com)
ha scritto:

[...]

PR #786, opened 2020-01-24
Allow configure highlight click settings from macOS high sierra

pettomartino:
Addressing Issue https://github.com/wulkano/kap/issues/784

As I mentioned in that issue I've tested this on my machine which runs macOS high sierra

image

Kapture 2020-01-24 at 13 46 37
image
Kapture 2020-01-24 at 13 47 55

### DIFF
--- a/renderer/components/preferences/categories/general.js
+++ b/renderer/components/preferences/categories/general.js
@@ -24,7 +24,7 @@ class General extends React.Component {
 
   componentDidMount() {
     this.setState({
-      showCursorSupported: electron.remote.require('./utils/macos-version').isGreaterThanOrEqualTo('10.14')
+      showCursorSupported: electron.remote.require('./utils/macos-version').isGreaterThanOrEqualTo('10.13')
     });
   }
 


karaggeorge:
@sindresorhus do you remember why we had that check there in the first place? I looked at the commit and it looks like it was part of a bigger PR refactoring the settings, so I don't know what the purpose was exactly. 

From the API page, I don't see any OS version requirements: https://developer.apple.com/documentation/avfoundation/avcapturescreeninput/1385722-capturesmouseclicks

Should we remove that check completely?

sindresorhus:
It was introduced in https://github.com/wulkano/kap/commit/92178c548eaa709bffd0b88e22685b4a80260ad3. See https://github.com/wulkano/kap/issues/477.

// @albinekb

albinekb:
That check was added there because:
>Apple broke the cursor hide/show API in high sierra, so we hide the option there to avoid confusion.

I could not get it to work correctly, I even filed a radar for it see: https://feedbackassistant.apple.com/feedback/5538640

karaggeorge:
@albinekb sounds like there might have been other circumstances that caused it not to work, since @pettomartino  has it working on High Sierra. I think it's okay to remove the check until we encounter the issue again and then we can do some more research then. Thoughts?

pettomartino:
The issue described seem to be related to "show cursor" instead of "highlight click". However, I have tested it as well, and seem to be working too:

image

pettomartino:
Sorry closed by mistake

pettomartino:
I also tested with trackpad and external mouse and both work the same way.
I can't reproduce issues in high sierra.
I got an old MBP 2011 that won't update further than high sierra, so I'm stuck in this version, more people might have same issue, so if you guys are ok with that I think that might be beneficial.

PR #756, opened 2019-11-21
Use yarn for plugin installation

karaggeorge:
Fixes #753 

Uses yarn for plugin installation. Unfortunately yarn [doesn't support prune](https://github.com/yarnpkg/yarn/issues/696), so I kept the npm binary as well.

I tested locally can didn't get any errors. There wasn't specific STR, it just happened sometimes, so I'll do some more testing to make sure the issue is resolved

PR #787, opened 2020-01-28
Update plugins docs and add a link to them in the main readme

karaggeorge:
Adresses https://github.com/wulkano/kap/pull/756#issuecomment-578382551

Might want to consider adding a link to the plugin docs on the website as well


sindresorhus:
> Might want to consider adding a link to the plugin docs on the website as well

Yes. We should definitely have a link from the website.

PR #791, opened 2020-02-05
Fix plugin config text wrap

karaggeorge:
Ensures that the text can wrap and doesn't create a horizontal scrollbar


sindresorhus:
Can you fix the Yarn conflict?

karaggeorge:
Yup, was already on it, should be good after build is done

Issue #785, opened 2020-01-13
More powerful plugin API proposal


karaggeorge:
# Plugins API

Looking for feedback on couple of thoughts around ways to improve our plugin API

## Reasoning

There's a lot of requested features for Kap and I know we try to not add new settings/preferences, but it gets hard to manage all of that. I think a good solution is for these features to be plugins. The current plugin API is a bit limiting (only exporting). We should try to expand it a bit more and allow for a bigger variety of plugins, that can take the load off of Kap. This will also help people only have the functionality they want.

I've been thinking about a few of these for a while, but I wanted to put it in writing, so I can keep track and get other people's thoughts on the matter before I start implementing parts

### Exports

This is the only type of plugin we support at the moment. I think the API we currently have is good, but we might experiment with more options, like:
- Allowing plugins to add/modify ffmpeg arguments (use cases: slow motion, speed up)
- Allow plugins to not use the default ffmpeg process we provide but handle the export completely on their own (more complex ffmpeg conversions)
- Allow plugins to provide error messages (we let them provide loading text, should be able to have them show a meaningful error when the export fails)
- Maybe somehow allow export plugins to be connected? As in, Recording -> Reverse plugin -> Slow mo plugin -> Export to GIF
    - For this it might be worth having two types of plugins, one being editing and one being export (editing being mp4 -> mp4, exporting being mp4 -> target)

### Recording

The way I'm picturing this is with some hooks (maybe an event emitter) that allows plugins to run code before the recording starts, and after it's done (maybe separate for error/success). This would allow current built-in features like `Hide Desktop Icons`/`DND`/`Key Cast`/`Custom Cursor` to be plugins.

### Deep Linking

For any type of plugin, I think deep linking is something that might be useful for OAuth with external services. We could have the format of the default url be something like `kap://plugin/{pluginName}/some/path` and the app is responsible of taking that and sending `/some/path` to the right plugin.

Implementation-wise, I'm not sure what the best way to do it. OAuth is generally an async process, but the plugin will be probably opening a link to the login page and waiting for a response. So maybe a `waitForDeepLink()` function that just waits for that deeplink and resolves when it receives it? Or maybe have them define a general deep link callback, which then can take care of writing whatever data it needs (token) to the store async from when the login happened? (either way we can document it)

### Settings

[...]

Issue #637, opened 2019-02-23
Handle `password` field type for plugin configuration


ochorocho:
It would be great to be able to store a password in a secure way in the plugins configuration.

Example:
```
password: {
      title: 'Password',
      type: 'password',
      required: true
},
```

This consists of 2 components:

* Add an `ajv` definition for type `password`
* Use `bcrypt` to store the encrpyted password

Related to wulkano/kap/pull/623

karaggeorge:
I think I'm missing something. Wouldn't the point of saving the password in the config be to use it at some point in the plugin runtime? How would we retrieve the password if we hash it with bcrypt before saving? We would have to save it as plaintext.

Does the service you are trying to build a plugin for support tokens? I think that would probably be the best way of going about it, rather than saving the user's password.

ochorocho:
@karaggeorge seems like a token can be used: https://docs.nextcloud.com/server/13/developer_manual/api/OC/Authentication/Token/IToken.html

The service i try to integrate ist called "Nextcloud"

karaggeorge:
@ochorocho I would suggest using that instead, as we can't really save a user's password other than plaintext

sindresorhus:
Supporting a password field might be a bad idea as it gives the illusion of security, but in reality, the password is just stored in plain text on disk.

Instead, I think we should update the plugin guide about recommending using tokens instead.

skllcrn:
To me it sounds like we should force tokens @sindresorhus for security reasons @sindresorhus, but I'm not sure how or if that would work. Like @karaggeorge mentioned:

>Does the service you are trying to build a plugin for support tokens?

 I'll update the docs to encourage token usage and close this issue unless anyone has additions @ochorocho.

sindresorhus:
> To me it sounds like we should force tokens for security reasons @sindresorhus

I don't see how we would enforce that. Not providing a `password` type and documenting the recommendation is the most we can do, I think.

karaggeorge:
Yeah, we can add a note that basically says the plugins' config is saved as a plaintext file, so saving passwords is not particularly safe, but at the end of the day the plugins are made by other users, so if a user wants to install one and enter their password that's up to them

ochorocho:
@skllcrn ok, thanks. I''m using now nextclouds login flow and its working how i wanted it to work :-)

PR #795, opened 2020-02-14
Add deep linking support for plugins

### LINKED ISSUE #697: Add Deep Linking (Plugin OAuth)
Feature Request: deep linking

It would be nice to have deep linking for Kap, mainly for oauth purposes, or anything else we can think of. My main use-case would be for plugin oauth. Some services like Dropbox don't accept apps if the oauth happens in a "web view". They need it to happen in the user's browser. A way to get around that without having a server running, is to set the callback url to be a deep link to Kap, which can then receive the code. After that we'll need some kind of API for Kap to forward that code to the right plugin (and add that to the plugin docs).

If not able to deep link directly (some services require a https url), we might set up a dummy Now deployment that explains that it will open the app and deep link from there (similar to how Slack does it)

karaggeorge:
Closes #697 

Initial implementation of adding deep linking to Kap, as discussed in #785 

Will be updating `kap-dropbox` to test this PR's build


karaggeorge:
The kap-dropbox plugin has been updated to use the deep linking functionality and it worked as expected.

https://github.com/karaggeorge/kap-dropbox

I've undeprecated it, but set minimum version to 3.0.2 since it won't work with the current released version. Once this is released it can be used in Kap

sindresorhus:
@jsejcksn I think this will simplify authentication for you as you could probably use: https://developers.gfycat.com/api/#browser-based-authentication

main/utils/deep-linking.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+const {URL} = require('url');
```
sindresorhus:
```suggestion
```

URL is a global now.

main/utils/deep-linking.js
```diff
@@ -0,0 +1,39 @@
...
+const handlePluginsDeepLink = path => {
+  const [plugin, ...rest] = path.split('/');
+
+  if (pluginPromises.has(plugin)) {
+    pluginPromises.get(plugin)(rest.join('/'));
+    pluginPromises.delete(plugin);
+    return;
+  }
+
+  console.log(`Received link for plugin ${plugin} but there was no registered listener.`);
```
sindresorhus:
```suggestion
  console.error(`Received link for plugin ${plugin} but there was no registered listener.`);
```

?

main/utils/deep-linking.js
```diff
@@ -0,0 +1,39 @@
...
+};
+
+const addPluginPromise = (plugin, resolveFunc) => {
```
sindresorhus:
```suggestion
const addPluginPromise = (plugin, resolveFunction) => {
```

jsejcksn:
@sindresorhus Thanks. It looks like the gfycat browser auth flow still requires utilizing a registered API key (`client_id`). Does Kap provide the default API keys for plugins?

main/share-service-context.js
```diff
@@ -109,6 +110,12 @@ class ShareServiceContext {
 
     this.config.openInEditor();
   }
+
+  waitForDeepLink() {
```
sindresorhus:
```suggestion
  async waitForDeepLink() {
```

For readability, I prefer to mark any function that returns a promise with `async`.

sindresorhus:
> I've undeprecated it, but set minimum version to 3.0.2

Nitpick, but this is a new feature, so the Kap version would be 3.1.0.

sindresorhus:
> It looks like the gfycat browser auth flow still requires utilizing a registered API key (client_id). Does Kap provide the default API keys for plugins?

I'm not sure. I guess we could register a `client_id` or you could.

jsejcksn:
> I'm not sure. I guess we could register a `client_id` or you could.

I _can_ register one, but I think it's better for user trust if the Kap project owner is the one managing the default API keys. I suppose that would need to be a part of the process of integrating each new service as a plugin.

karaggeorge:
I don't think it's a bad thing for the plugin owner to register the API keys. The whole point about having support for plugins is that the community can create them independent from Kap's team. Similar to how if there was an issue with the plugin, the issue would open in that repo and not this one (not controlled by the Kap team). We love to help with developing and debugging the plugins as much as possible, but I don't know if we have a central locations that we would manage API keys

main/utils/deep-linking.js
```diff
@@ -0,0 +1,38 @@
...
+]);
+
+const handleDeepLink = url => {
+  const {host, pathname} = new URL(url);
+
+  if (routes.has(host)) {
+    return routes.get(host)(pathname.slice(1));
+  }
+
+  console.log(`Route not recognized: ${host} (${url}).`);
```
sindresorhus:
```suggestion
  console.error(`Route not recognized: ${host} (${url}).`);
```

main/utils/deep-linking.js
```diff
@@ -0,0 +1,38 @@
...
+const handlePluginsDeepLink = path => {
+  const [plugin, ...rest] = path.split('/');
+
+  if (pluginPromises.has(plugin)) {
+    pluginPromises.get(plugin)(rest.join('/'));
+    pluginPromises.delete(plugin);
+    return;
+  }
+
+  console.error(`Received link for plugin ${plugin} but there was no registered listener.`);
```
sindresorhus:
```suggestion
  console.error(`Received link for plugin “${plugin}” but there was no registered listener.`);
```

karaggeorge:
@sindresorhus this should be good after it's built

jsejcksn:
> We love to help with developing and debugging the plugins as much as possible, but I don't know if we have a central location that we would manage API keys

So if a plugin author wants to contribute a plugin that requires API keys, but does not want to be associated to unknown user activity by providing an API key with the plugin, should the author simply leave the fields required and empty?

sindresorhus:
> I don't think it's a bad thing for the plugin owner to register the API keys. The whole point about having support for plugins is that the community can create them independent from Kap's team. Similar to how if there was an issue with the plugin, the issue would open in that repo and not this one (not controlled by the Kap team). We love to help with developing and debugging the plugins as much as possible, but I don't know if we have a central location that we would manage API keys

I agree, but we should definitely document that fact in the plugins docs.

karaggeorge:
> So if a plugin author wants to contribute a plugin that requires API keys, but does not want to be associated to unknown user activity by providing an API key with the plugin, should the author simply leave the fields required and empty?

I think we are confusing two different things here. If the user needs an API key to make the upload, then there's two ways to obtain that (usually):
- They generate and find an API key themselves. In which case, you can add the API key as a required and empty field with instructions on where to get it. The user will see a message saying the plugin requires configuration.
- There's an OAuth flow of some sort, in which case when the plugin runs, you can check if you have the key, and if not trigger the browser flow. On return, save that key in the config without the user ever seeing or having to enter it.

For the latter, most OAuth APIs required a registered "App". This app has a client and secret key used for the code <-> token exchange. Those client/secret keys should be maintained by the author of the plugin.

The API keys (or tokens) should only be maintained in the app locally (either by the user or the plugin), but not by the author or the Kap team.

For an example of the first flow you can look at [kap-giphy](https://github.com/wulkano/kap-giphy), and for an example of the second flow check [kap-dropbox](https://github.com/karaggeorge/kap-dropbox)

I hope this makes sense 😄

sindresorhus:
@karaggeorge Almost forgot, this will need some docs additions too.

karaggeorge:
@sindresorhus Added 😄

jsejcksn:
> If you want I can create a new repo, and try to explore the browser api of gfycat, and how it would work, so I can get a much more detailed flow

The gfycat browser auth uses the second flow described above, which requires registering an "app" with them to obtain a `client_id` that it uses in the auth requests.

Alternatively, a user can generate an `id` and `secret` (which is how the plugin works now).

I do not want to have any connections with the use of the plugin beyond the code that I write for it. I simply want to contribute the functionality of the plugin because I like Kap.

karaggeorge:
> which is how the plugin works now

@jsejcksn I think that's fine. Especially since the deep linking flow is not released yet. If you don't mind, I can open a PR with the OAuth flow after we release the deep linking feature, and I can generate the app keys myself

jsejcksn:
> If you don't mind, I can open a PR with the OAuth flow after we release the deep linking feature, and I can generate the app keys myself

I'd like that. I anticipate that it will be an improved experience in two ways:

 - the user won't have to generate keys, and
 - the current auth flow doesn't associate uploads to the user's account unless a username/password are provided that match the account which generated the keys (and I don't think that's a good idea #637)

docs/plugins.md
```diff
@@ -100,6 +100,7 @@ The `action` function is where you implement the behavior of your service. The f
 - `.setProgress(text, percentage)`: Update progress information in the Kap export window. Use this whenever you have long-running jobs, like uploading. The `percentage` should be a number between `0` and `1`.
 - `.openConfigFile()`: Open the plugin config file in the user’s editor.
 - `.cancel()`: Indicate that the plugin operation canceled for some reason. [Example.](https://github.com/wulkano/kap/blob/efc32d12f381615c9fcfc41065d9c2ee200e8975/app/src/main/save-file-service.js#L28-L31) This closes the Kap export window. If the cancelation was not the result of a user gesture, use `.notify()` to inform the user why it was canceled.
+- `.waitForDeepLink()`: Returns a Promise that resolves when a deep link for this plugin is opened. The link should be of the format `kap://plugins/{pluginName}/{rest}`, where `pluginName` is the npm module name and `rest` is the string the Promise will resolve with.
```
sindresorhus:
```suggestion
- `.waitForDeepLink()`: Returns a `Promise` that resolves when a deep link for this plugin is opened. The link should be in the format `kap://plugins/{pluginName}/{rest}`, where `pluginName` is the npm package name and `rest` is the string the `Promise` will resolve with.
```

sindresorhus: I don't think it's clear from a plugin author perspective what this method is for specifically. We should say that it can be useful for OAuth flows, which requires opening the browser, authenticating, and then have the website open Kap. This is where it becomes useful. And we could also link to kap-dropbox as an example.

karaggeorge: Just added a whole OAuth dedicated section with examples

sindresorhus:
Can you fix the Yarn conflict?

sindresorhus:
🕺✨

PR #796, opened 2020-02-14
Display error when a plugin throws

karaggeorge:
While reworking kap-dropbox I noticed that if a plugin throws a javascript error at load time (I had forgotten to mark a function as async and used await), Kap silently fails to load all other plugins as well. The app still works, but you can only export using the default services.

This makes sure that if a plugin fails, we just ignore it instead of causing other plugins to hide as well.

### DIFF
--- a/main/export-options.js
+++ b/main/export-options.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const {dialog} = require('electron');
 const Store = require('electron-store');
 const {ipcMain: ipc} = require('electron-better-ipc');
 
@@ -8,6 +7,7 @@ const plugins = require('./common/plugins');
 const {converters} = require('./convert');
 const {setOptions, getEditors} = require('./editor');
 const {apps} = require('./plugins/open-with-plugin');
+const {showError} = require('./utils/errors');
 
 const exportUsageHistory = new Store({
   name: 'export-usage-history',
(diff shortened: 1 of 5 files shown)


sindresorhus:
Is the user informed in some way which plugin failed and the error message it failed with?

karaggeorge:
No, not if the plugin fails while it's being loaded. I mean this type of error probably should be caught by the developer before they publish it (assuming they tested it), since it happens at `require('kap-plugin')` time. I'm working on a PR to show the errors if they happen at export time. Should I add some kind of error dialog if it fails at require time too?

sindresorhus:
> Should I add some kind of error dialog if it fails at require time too?

I think that would be dev-friendly, yes.

karaggeorge:
Would a simple notification saying `Plugin giphy failed to load` be sufficient you think? Or show a full dialog with the error stack

sindresorhus:
> Or show a full dialog with the error stack. I could console.log the error if running in dev mode

This. It should be a loud error.

karaggeorge:
Screen Shot 2020-02-14 at 2 23 59 PM

main/export-options.js
```diff
@@ -60,6 +62,7 @@ const getExportOptions = () => {
         }
       }
     } catch (error) {
+      dialog.showErrorBox(`Something went wrong while loading ${json.pluginName}`, error.stack);
```
sindresorhus:
```suggestion
      dialog.showErrorBox(`Something went wrong while loading “${json.pluginName}”`, error.stack || error || 'Unknown error');
```

It could fail with a native C error if the `.stack` is not defined, which could happen. I experienced that with Caprine.

karaggeorge: Nice little utility. I abstracted the method and used it 👍

sindresorhus:
Purrfect 🐈

PR #798, opened 2020-02-14
Plugin error messages

karaggeorge:
Make sure to show error messages when a plugin export fails. Shows the message next to the status, truncated if it's too long. On hover there's a tooltip with the entire message

Screen Shot 2020-02-14 at 12 41 02 PM


sindresorhus:
How did you simulate an error?

karaggeorge:
I edited `kap-giphy`'s `index.js` file in my `node_modules` and simply did `throw new Error('message')`

sindresorhus:
It looks a bit messy to show just part of the message. How about instead we just show a small info `(i)` icon after "Export Failed" and if the user hover over "Export Failed" or the icon, we show the full message?

// @skllcrn

skllcrn:
That could work, but I honestly find seeing at least part of the message without having to hover anything to be better. The error should trigger a notification with the error message too.

PR #801, opened 2020-02-18
Use native about window

karaggeorge:
Switch to use the native about window implementation﻿


main/menus.js
```diff
@@ -75,7 +74,12 @@ const sendFeedbackItem = {
 
 const aboutItem = {
   label: `About ${app.name}`,
-  click: openAboutWindow
+  click: () => {
+    closeAllCroppers();
+    ensureDockIsShowing(() => {
+      showAboutWindow({copyright: 'Copyright © Wulkano'});
```
sindresorhus: It's better to set the options upfront with https://www.electronjs.org/docs/api/app#appsetaboutpaneloptionsoptions at launch. You also don't need the `electron-util` method. It's only useful for cross-platform support.

sindresorhus:
I launch Kap, the cropper opens, in the cropper, I click the "About" menu item, and the About window opens behind Chrome for some reason. If Chrome is not open, same happens with Finder.

karaggeorge:
Hmm, it worked for me. In general I noticed weird behavior with the native about window when there's no other windows open. Similar to the "Open file" when the dock is not showing. I tried using the `ensureDockIsShowing` we used for that, and it seemed to work for me (as in it doesn't show in the background). Not sure if there's anything else I can do to make it show up front 🤔

PR #802, opened 2020-02-18
Remember FPS preference across editors

karaggeorge:
Save the fps used in an export in a similar way as we do for plugin usage and use that when the editor first loads. If your preference is let's say 45, but you recorded with 30 fps (or opened a video that has 30fps) the minimum of those values will be used﻿

### DIFF
--- a/main/export-options.js
+++ b/main/export-options.js
@@ -19,6 +19,13 @@ const exportUsageHistory = new Store({
   }
 });
 
+const fpsUsageHistory = new Store({
+  name: 'fps-usage-history',
+  defaults: {
+    fps: 60
+  }
+});
+
 const prettifyFormat = format => {
   const formats = new Map([
     ['apng', 'APNG'],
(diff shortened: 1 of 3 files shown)


renderer/containers/editor.js
```diff
@@ -20,7 +20,17 @@ export default class EditorContainer extends Container {
     const src = `file://${filePath}`;
     this.finishLoading = resolve;
 
-    this.setState({src, filePath, originalFilePath, fps, originalFps: fps, wasMuted: false, isNewRecording});
+    this.setState({
+      src,
+      filePath,
+      originalFilePath,
+      // TODO: Fix this ESLint violation
+      // eslint-disable-next-line react/no-access-state-in-setstate
```
sindresorhus: `setState` supports a callback which is meant for modifying the current state.

main/export-options.js
```diff
@@ -19,6 +19,13 @@ const exportUsageHistory = new Store({
   }
 });
 
+const fpsUsageHistory = new Store({
+  name: 'fps-usage-history',
+  defaults: {
```
sindresorhus: I would recommend using the `schema` option for new things as it validates the type: https://github.com/sindresorhus/electron-store#schema

renderer/containers/editor.js
karaggeorge: unstated version 1 didn't support this. Version 2.x supports it but when I try to upgrade the cropper breaks. Moving the state changes to being async in 2.x, means that there's a ton of changes needed in the cropper logic, with a lot of `await`s for `setState` to make sure the timing works for resizing/moving. I have to convert the entire container to the new format, but it'll take a while and should probably be in a new PR

sindresorhus: Ah ok. No point in upgrading to unstated 2 though, as there is: https://github.com/jamiebuilds/unstated-next We should stay with unstated 1 for now and we can look at a hook-based replacement at some point when the state hook ecosystem is more mature.

karaggeorge: Sounds good. Might look into converting the React code to use hooks at some point. =

Also, we could look into converting at least the renderer code in TypeScript, since Next.js supports it out of the box with 0 config. Not sure if that makes more or less friendly for contributors

sindresorhus: Slowly converting to hooks would be super nice. I think that would help lower the threshold for contribution as hooks are more easily approachable.

sindresorhus: Yeah, TS would be nice, but I would wait at least a year. The current React type definition for TS is pretty subpar, and TS is annoying to deal with and way too many bugs.

Issue #774, opened 2019-12-05
Memorable file names


sospedra:
### Feature request

Gret job with Kap folks. Really good. I come to explain a situation I found myself quite often.
I take multiple short videos to show what I did. Usually these files are posted alongisde a PR. The thing is that at the moment I have to put them in the PR description is quite difficult to remember which gif correspond to what feature.

![image](https://user-images.githubusercontent.com/3116899/70246644-5488fe80-1778-11ea-9444-300f9bd68f75.png)
_Kap file explorer screenshot_

I think that memorable names instead of time hashes will help with this issue. I found this other report: https://github.com/wulkano/kap/issues/695. Depending on how that is implemented my proposal could fit.

I'd love to help with this if you think is something you wanna have in Kap. Let me know if so.

Again, thanks for the hard work 👏

skllcrn:
Do you think being able to edit filenames in the exports list could be a good solution for this too?

sospedra:
I'm no expert in UX. It'll work for me tho I can see that the solution might be suboptimal.
Honestly, I feel like the other issue I linked is worth a shot. But it's also way more complex.
I have the same problem with other date-based naming conventions. Like mobile or macos screenshots. Maybe it's just my brain 🧠😜

blandonfrank:
I think that making the filenames editable in the export list would work well. A simple click on to rename would work. Currently I usually just go into the export folder and edit the names there.

sindresorhus:
I like the idea. We could use something like https://github.com/andreasonny83/unique-names-generator to generate fun names like `Awesome Donkey`, `Mesmerizing Sea Turtle`, etc.

blandonfrank:
> I like the idea. We could use something like https://github.com/andreasonny83/unique-names-generator to generate fun names like `Awesome Donkey`, `Mesmerizing Sea Turtle`, etc.

I don't know if "Awesome Donkey" would help him remember what he was recording. But then again, maybe it would lol

skllcrn:
I think the ideal solution here is to easily let you change the filename prior to export, some apps let you edit the filename from the titlebar, like Sketch: 
![image](https://user-images.githubusercontent.com/8440026/75118212-73f93e80-5678-11ea-9773-e1c290f5d294.png)

karaggeorge:
Can we close this in favor of: https://github.com/karaggeorge/kap-recording-name ?

PR #804, opened 2020-02-20
Support recording and edit plugins

### LINKED ISSUE #453: Record in slow-motion / Adjust playback speed
Noticed [this tweet](https://twitter.com/dan_abramov/status/987802086355210240) and thought of the general use-case of Kap. Recording behaviour to report somewhere. Would be amazing to record in slow-motion, even if it's faux slow-motion just with increasing the playback time of the recording so that each frame displays longer.

### LINKED ISSUE #550: Edit-service plugin
I want to write a plugin for reverse the capture. But it seems there is no API for using the built-in FFmpeg module.

Use story for the `kap-reverse`:

1. Use Kap to record a movie segment
2. Then use `kap-reverse` to get a funny GIF or MP4 file

Also, I think it will be useful for others.

karaggeorge:
This PR refactors the plugin system, along with the UI as discussed with @skllcrn (discover tab is now Browse, and contains all plugins)

There are also 4 recording plugins created:
-  https://github.com/karaggeorge/kap-hide-desktop-icons
-  https://github.com/karaggeorge/kap-do-not-disturb
-  https://github.com/karaggeorge/kap-key-cast
-  https://github.com/karaggeorge/kap-soundflower

And 3 edit plugins: (closes #550)
- https://github.com/karaggeorge/kap-progress-bar
- https://github.com/karaggeorge/kap-reverse
- https://github.com/karaggeorge/kap-playback-speed (closes #453)

Version is set to 3.2.0, so we can test the build with the released plugins. Will need to turn back to 3.1.0 before merging

Other changes include:
- `didConfigChange` hook for all plugins
- Improvements around tracking and error handling
- Not trimming size and time in ffmpeg command when the user doesn't change it in the editor
- Cleaning up Plugin and ServiceContext implementations

This is ready to be tested


main/common/aperture.js
```diff
@@ -18,34 +16,41 @@ const {hasMicrophoneAccess} = require('./system-permissions');
...
-    hideDesktopIcons,
-    doNotDisturb
-  } = lastUsedSettings;
+const callPlugins = async method => Promise.all(recordingPlugins.map(async ({plugin, service}) => {
+  if (service[method] && typeof service[method] === 'function') {
+    try {
+      await service[method]({
+        options: apertureOpts,
+        state: serviceState.get(service.title),
+        config: plugin.config
```
karaggeorge: This will be replaced with a context similar to the share service, they will share a bunch of common fields, so I'll do an abstraction

main/common/aperture.js
```diff
@@ -1,8 +1,6 @@
 'use strict';
 
 const {dialog} = require('electron');
-const desktopIcons = require('hide-desktop-icons');
-const dnd = require('@sindresorhus/do-not-disturb');
```
karaggeorge: Need to remove these from the dependencies and also from the settings Store

main/common/aperture.js
```diff
@@ -163,6 +165,8 @@ const stopRecording = async () => {
     return;
   }
 
+  await callPlugins('willStopRecording');
```
karaggeorge: Thinking of removing this as it might cause recordings to continue longer than the user clicked. If not removing at least not `await`ing

main/menus.js
```diff
@@ -175,9 +185,20 @@ const applicationMenuTemplate = [
...
+const refreshRecordPluginItems = services => {
+  pluginsItems = services.map(service => ({
+    label: service.title,
+    type: 'checkbox',
+    checked: service.isEnabled,
+    click: service.toggleEnbaled
+  }));
+
+  cogMenu = Menu.buildFromTemplate(getCogMenuTemplate());
+};
```
karaggeorge: This is a workaround, since I can't grab the item itself and change submenu. I have to keep recreating the menu itself

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "kap",
   "productName": "Kap",
-  "version": "3.1.0",
+  "version": "3.2.0",
```
karaggeorge: TODO: change this back to 3.1.0 after it's been tested

karaggeorge: Maybe consider allowing Kap in dev mode to ignore version restrictions

main/common/aperture.js
```diff
@@ -18,34 +16,41 @@ const {hasMicrophoneAccess} = require('./system-permissions');
...
+const callPlugins = async method => Promise.all(recordingPlugins.map(async ({plugin, service}) => {
+  if (service[method] && typeof service[method] === 'function') {
+    try {
+      await service[method]({
+        options: apertureOpts,
+        state: serviceState.get(service.title),
+        config: plugin.config
+      });
+    } catch (error) {
+      console.error(error);
```
sindresorhus: I think this should have an error dialog.

main/common/aperture.js
sindresorhus: Agreed, should be removed.

main/common/plugins.js
```diff
@@ -6,27 +8,70 @@ const execa = require('execa');
...
+
+    if (service.willEnable) {
+      try {
+        const canEnable = await service.willEnable();
+
+        if (canEnable) {
+          recordPluginServiceState.set(service.title, true);
+        }
+      } catch (error) {
+        showError(error, {title: `Something went wrong while enabling “${service.title}”`});
```
sindresorhus: Can you make a `showErrorAndReport()` utility method that shows the error and also reports to sentry?

sindresorhus:
```suggestion
        showError(error, {title: `Something went wrong while enabling “${service.title}”`, reportToSentry: true});
```

renderer/components/config/tab.js
```diff
@@ -8,11 +8,21 @@ import Switch from '../preferences/item/switch';
 import {OpenOnGithubIcon, OpenConfigIcon} from '../../vectors';
 
 const ConfigInput = ({name, type, schema, value, onChange, hasErrors}) => {
-  if (type === 'string') {
+  if (type === 'string' || type === 'number') {
     const className = hasErrors ? 'has-errors' : '';
+    const handleChange = e => {
```
sindresorhus:
```suggestion
    const handleChange = event => {
```

renderer/components/config/tab.js
```diff
@@ -8,11 +8,21 @@ import Switch from '../preferences/item/switch';
...
+      const value = type === 'string' ? e.target.value : parseFloat(e.target.value);
```
sindresorhus:
```suggestion
      const value = type === 'string' ? e.currentTarget.value : Number.parseFloat(e.target.value);
```

main/menus.js
```diff
@@ -91,11 +91,21 @@ const preferencesItem = {
...
+  submenu: pluginsItems,
+  visible: pluginsItems.length > 0
+});
+
+const getCogMenuTemplate = () => [
   aboutItem,
   {
     type: 'separator'
   },
+  getPluginsItem(),
```
sindresorhus: This should be moved to after `sendFeedbackItem` and with a separator between.

main/plugin.js
```diff
@@ -1,37 +1,104 @@
 'use strict';
 
-const {app, shell} = require('electron');
 const path = require('path');
 const fs = require('fs');
+const electron = require('electron');
+const {satisfies} = require('semver');
```
sindresorhus:
```suggestion
const {satisfies} = require('semver');
```

I don't really like destructuring like this. It makes it very hard to see the meaning at the callsite. "Satisfies what?" It's not clear it's related to semver then.

sindresorhus:
I quickly tried it out. Seems to work well 👌

main/service-context.js
```diff
@@ -123,4 +89,61 @@ class ShareServiceContext {
...
+  }
+}
+
+class RecordServiceContext extends ServiceContext {
+  constructor(options) {
+    super(options);
+
+    this.apertureOptions = options.apertureOptions;
+    this.state = options.state;
+    this.config = options.config;
```
sindresorhus: This is already set in `ServiceContext`

sindresorhus:
The white line (the longest one) a bit too strong:

Screenshot 2020-02-22 at 14 36 12

sindresorhus:
I wonder if we can create a custom Ajv type for hex color and use `<input type="color">` for rendering it.

And `Hex` => `hex`.

Screenshot 2020-02-22 at 14 39 50

sindresorhus:
Too much line height on the description here:

Screenshot 2020-02-22 at 14 41 16

sindresorhus:
I've noticed that it's not obvious after installing a plugin that it has config. I usually install a plugin and then go to record, and I forget to check whether it has any settings. Maybe after installing a plugin, we should open the config window directly if any, so the user has a chance to set the initial config.

docs/plugins.md
```diff
@@ -84,27 +86,217 @@ const giphy = {
...
+
+The `config` and `configDescription` properties are optional.
+
+Example:
+
+```js
+const action = async context => {
+	// Do something
+
+	context.notify('Notify about something');
```
sindresorhus: This and some other examples are using tab-indentation. You should be consistent and use 2-space indentation for all examples.

docs/plugins.md
```diff
@@ -84,27 +86,217 @@ const giphy = {
...
+};
+
+const config = {
+  percent: {
+    title: 'Slow Down Percentage',
+	type: 'number',
+	maximum: 1,
+	minimum: 0,
```
sindresorhus: Mixed indentation

docs/plugins.md
```diff
@@ -84,27 +86,217 @@ const giphy = {
...
+- `.inputPath`: The path to the input trimmed `mp4` file.
+- `.outputPath`: The path where the resulting `mp4` file should be by the end of the action.
+- `.exportOptions`: An object containing info about the recording (note that the input video has already been resized and trimmed):
+	- `.width`: Width of the input file.
+	- `.height`: Height of the input file.
+	- `.format`: The selected format in which the video will be converted to later on.
+	- `.fps`: The selected FPS that will be used for the final conversion.
+	- `.duration`: Duration of the trimmed input file.
+	- `.isMuted`: Whether the video is muted or not.
+	- `.loop`: Whether the resulting GIF or APNG file will be looped or not.
```
sindresorhus: Should use 2-space

sindresorhus:
Instead of hiding recording plugins in the `...` menu. How about we add a small triangle arrow to the right-side of the record button, that when clicked, opens a menu?

Screenshot 2020-02-22 at 15 19 18

main/common/plugins.js
```diff
@@ -120,6 +120,15 @@ class Plugins {
       await this._yarnInstall();
 
       const plugin = new InstalledPlugin(name);
+
+      if (plugin.plugin.didInstall && typeof plugin.plugin.didInstall === 'function') {
+        try {
+          plugin.plugin.didInstall(plugin.config);
```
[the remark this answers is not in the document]
karaggeorge: That makes sense, will update 👍

karaggeorge:
> I wonder if we can create a custom Ajv type for hex color 

I was actually thinking about that. If we can create custom types, we can use modules like a color picker for color, and something like a file location, where it can be a native file dialog (like when picking ~/Kaptures directory)

> Maybe after installing a plugin, we should open the config window directly if any,

We currently show a notification that can open the config window only if the plugin has config that is not valid (required field is empty etc), since in this case you can't use it until you configure. Should we show a similar notification with different text in any case a plugin has a config?


Three other thoughts I had yesterday:
1. would it be worth having plugin note their desired type in the package.json, so if we have a ton of plugins, and search implemented, we could do github-like filters like `is:record` or `is:share` to filter down?

2. for recording plugins, if they need some more control, should we allow them to provide a submenu they can attach to the menu item under the `…` -> Plugins -> Title? 

3. Should we give the recording hooks more control? Like ability to prevent defaults? So the `didStopRecording` or a new `willOpenEditor` can `preventDefault` and stop the editor from opening. Not just this, but functionalities like that. This might allow for the "Quick mode" to be a plugin (skip editor, trigger an export directly)

sindresorhus:
> Should we show a similar notification with different text in any case a plugin has a config?

:+1: Although, one downside with notifications is for people like me that have notifications paused (DnD).

> would it be worth having plugin note their desired type in the package.json, so if we have a ton of plugins, and search implemented, we could do github-like filters like is:record or is:share to filter down?

I think it's a bit early to think about filters and search, but maybe it would be useful to mark packages with what they contain? We discussed this earlier. How about just predefined keywords? `kap-plugin-share`, `kap-plugin-edit`, `kap-plugin-record`?

> for recording plugins, if they need some more control, should we allow them to provide a submenu they can attach to the menu item under the … -> Plugins -> Title?

What kind of control? Example?

> Should we give the recording hooks more control? Like ability to prevent defaults? So the didStopRecording or a new willOpenEditor can preventDefault and stop the editor from opening. Not just this, but functionalities like that. This might allow for the "Quick mode" to be a plugin (skip editor, trigger an export directly)

Do you have a specific use-case? I don't think we should add just this "just in case". For example, if we allow preventing the editor, we also need to provide a way to manually open it.

karaggeorge:
> What kind of control? Example?
> Do you have a specific use-case? I don't think we should add just this "just in case". For example, if we allow preventing the editor, we also need to provide a way to manually open it.

For both, what I had in mind was the Quick Record plugin, where in the context menu, you would be able to not just enable/disable, but also open submenu which would allow you to pick format and export method, then you record, editor never shows, but the recording is sent to the format/export method you selected directly.

So, in order for something like this to be possible, we would need some extra control (I guess this could just be settings in the plugin config, but it would need access to stuff like other installed export plugins) and then a way to prevent the editor from opening, and triggering an export right away.

Not sure if this is the right approach, but I'm trying to think if this Quick Record could be a plugin somehow

karaggeorge:
Another thought on more flexibility for the plugin API. Should we allow recording plugins to control options? For example, on `willStartRecording` we pass in the `apertureOptions` object. Should we allow plugins to edit those before we pass them to aperture? 

Example use-case: a plugin that triggers something like [Soundflower](https://github.com/mattingalls/Soundflower) to record system audio, and then sets the input id of the `apertureOptions`. 

Also, for the `didStopRecording` should we allow the hook to change the default file name? 

Example use-case: a plugin where when you start recording, it captures the contents of your clipboard, and then at the end uses that to name your recording. Since most people know what they are about to record (for a PR or something), they can copy the PR title, proceed to trigger a recording and the resulting editor would have that title set already

sindresorhus:
> Not sure if this is the right approach, but I'm trying to think if this Quick Record could be a plugin somehow

Would be nice to make that possible, but I think we should tackle that later on after we've had some experience with the plugins.

> Example use-case: a plugin that triggers something like Soundflower to record system audio, and then sets the input id of the apertureOptions.

👍 

> Also, for the didStopRecording should we allow the hook to change the default file name?

This could potentially solve https://github.com/wulkano/kap/issues/774

sindresorhus:
This is absolutely amazing! I'm super excited about the possibilities. 🎉

PR #806, opened 2020-02-21
Show time recorded in the tray

### LINKED ISSUE #805: Display time recorded while capturing the screen
> It's better if I can see the time how long I have recorded when recording screen

karaggeorge:
Fixes #805

Examples of different cases (seconds/minutes/hours):

Screen Shot 2020-02-21 at 4 13 05 PM
Screen Shot 2020-02-21 at 4 15 01 PM
Screen Shot 2020-02-21 at 4 15 36 PM

### DIFF
--- a/main/tray.js
+++ b/main/tray.js
@@ -2,6 +2,7 @@
 
 const {Tray} = require('electron');
 const path = require('path');
+const parseMilliseconds = require('parse-ms');
 
 const {openCropperWindow} = require('./cropper');
 const {cogMenu} = require('./menus');
@@ -10,6 +11,7 @@ const openFiles = require('./utils/open-files');
 
 let tray = null;
 let trayAnimation = null;
+let trayTimerTimeout = null;
 
 const openContextMenu = () => {
   tray.popUpContextMenu(cogMenu);
@@ -37,6 +39,9 @@ const resetTray = () => {
     clearTimeout(trayAnimation);
   }
 
+  clearTimeout(trayTimerTimeout);
+  tray.setTitle('');
+
   tray.removeAllListeners('click');
   tray.setImage(path.join(__dirname, '..', 'static', 'menubarDefaultTemplate.png'));
   tray.on('click', openCropperWindow);
@@ -45,6 +50,7 @@ const resetTray = () => {
 
 const setRecordingTray = stopRecording => {
   animateIcon();
+  showTimeRecorded();
   tray.once('click', stopRecording);
 };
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
I like it, **but**, it's very distracting that the time moves constantly, since it's not using a monospaced font... In native, we could use `NSFont.monospacedDigitSystemFont` to make numbers-only monospaced. Can you open an issue on Electron about being able to set that font for the menu bar text? Could be an option like `tray.monospacedDigits = true`. Or maybe it could be more flexible. `try.font = {size: 10, weight: 'bold', monospacedDigits: true};`. I can image it would be useful in some cases to be able to also control the size and weight.

sindresorhus:
I also noticed that the timer starts before the menu bar icon animation finishes. They should be in sync, so either of them are currently wrong now. Not sure which.

sindresorhus:
When the hour is showing, it looks weird that there's more spacing between the icon and time. I think the hour should also be zero-padded.

main/tray.js
```diff
@@ -10,6 +11,7 @@ const openFiles = require('./utils/open-files');
 
 let tray = null;
 let trayAnimation = null;
+let trayTimerTimeout = null;
```
sindresorhus:
```suggestion
let trayTimerTimeout;
```

main/tray.js
```diff
@@ -70,6 +76,29 @@ const animateIcon = () => new Promise(resolve => {
   next();
 });
 
+const padNumber = (num, char = '0') => `${char}${num}`.slice(-2);
```
sindresorhus:
```suggestion
const padNumber = (number, character = '0') => `${character}${number}`.slice(-2);
```

main/tray.js
```diff
@@ -70,6 +76,29 @@ const animateIcon = () => new Promise(resolve => {
...
+  let trayTimer = 0;
+  const interval = 1000;
+
+  const tick = () => {
+    const {hours, minutes, seconds} = parseMilliseconds(trayTimer);
+    if (hours) {
+      tray.setTitle(` ${padNumber(hours, ' ')}:${padNumber(minutes)}:${padNumber(seconds)}`);
+    } else {
+      tray.setTitle(` ${padNumber(minutes)}:${padNumber(seconds)}`);
+    }
```
sindresorhus:
```js
const {hours, minutes, seconds} = parseMilliseconds(trayTimer);
let duration = `${padNumber(minutes)}:${padNumber(seconds)}`;

if (hours) {
  duration = `${padNumber(hours, ' ')}:${duration}`;
}

tray.setTitle(` ${duration}`);
```

?

sindresorhus:
@skllcrn Thoughts on this (design-wise)?

karaggeorge:
@sindresorhus I'll open an electron issue. Until then, is it worth making a swift module to use the native API? Or just wait for them to implement?

Also, in regards to the timing, both the animation and the counter start when the recording starts (aperture resolves), so I guess they are both right, if we say that the animation *starts* when the recording *starts*. Not sure if we can make it end when the recording starts, since we can't know the aperture timing

sindresorhus:
> Until then, is it worth making a swift module to use the native API? Or just wait for them to implement?

Definitely. I don't think they'll implement that soon.

> Not sure if we can make it end when the recording starts, since we can't know the aperture timing

If we can change the animation so that we can control it and end it at any time, we could start it when the user clicks the red record button, then just repeat the bounce animation until we get notification from Aperture that it started.

skllcrn:
This is looking good. It's a real bummer that we can't easily monospace the numbers, please link the Electron issue here so we can upvote! Are waiting for Aperture to start the recording, and then using the actual length of the capture to display the time?

karaggeorge:
Electron issue opened: https://github.com/electron/electron/issues/22363

Will start working on a small module to help us do this in the meantime

karaggeorge:
Closing in favor of https://github.com/karaggeorge/kap-recording-time

sindresorhus:
@karaggeorge Nice! Can you add a screenshot/GIF to its readme?

PR #814, opened 2020-03-25
Make sure pause is not called before play is finished

karaggeorge:
Fix this sentry issue:
https://sentry.io/organizations/wulkano-l0/issues/1230019159/?project=255536&query=&sort=freq&statsPeriod=14d

That has been triggered 698k times 

This is the error that is causing it: 
https://developers.google.com/web/updates/2017/06/play-request-was-interrupted

When basically `pause` is called before the `play` promise has resolved

Added a bunch of checks to make sure that doesn't happen any more

### DIFF
--- a/renderer/components/editor/video.js
+++ b/renderer/components/editor/video.js
@@ -28,17 +28,18 @@ class Video extends React.Component {
   }
 
   contextMenu = () => {
+    const {play, pause} = this.props;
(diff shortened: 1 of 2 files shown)


karaggeorge:
Might need to make another release for this if it keeps happening, otherwise it will flood our Sentry events

PR #827, opened 2020-04-16
Fix editor options bug

### LINKED ISSUE #824: Unable to export anything after upgrade to 3.2.0
macOS version: 10.13.6 
Kap version: 3.2.0

Upgrading to v3.2.0 removed all export/destination options after making a screen recording:

Screen Shot 2020-04-14 at 2 58 50 PM

I reverted back to 3.1.0 and it works again. Running High Sierra on a MBP, nothing unusual/notable about my setup as far as I can tell.

karaggeorge:
This PR attempts to debug the editor bug where the options don't appear. After we pinpoint the issue it'll contain the fix

Okay, few, figured it out after hours of super slow experimenting. My old macbook can barely run, so imagine running a VM on it, and trying to run visual applications. It was hell. But the bug was a sneaky one. So, it turns out it's not because they are on 10.13 per se, but it did mean they were the only ones able to see it. So, the bug would happen for everyone that only had 1 plugin for a format. But since we have the built-in ones, we have save to disk, copy to clipboard and open with, so 2 for mp4 webm (since no clipboard) and 3 for gif, apng. So, no way to get 1 plugin, so we never see the bug. Well, if you are below 10.14.4, the package open-with doesn't work unless you have the Swift runtime libs thing, but we built it well in a way that it doesn't break, it just doesn't add that plugin, which brings those users (if they have no plugins) to 1 plugin for mp4/webm (only save to disk), which then causes the bug. The bug itself was 2 lines: https://github.com/wulkano/kap/pull/827/files#diff-206b3e1cacb7c4112db05c82a975cdf2R133-R138 But because of the explanation above we just could not trigger it unless we were on 10.13 or lower. Few. Anyway, bug is (hopefully) fixed, but figured I'd share my journey

Fixes #824

### DIFF
--- a/main/common/plugins.js
+++ b/main/common/plugins.js
@@ -255,12 +255,11 @@ class Plugins {
       pluginPath: './plugins/save-file-plugin',
       isCompatible: true,
       name: '_saveToDisk'
+    }, {
+      pluginPath: './plugins/open-with-plugin',
+      isCompatible: true,
+      name: '_openWith'
     }];
-    // , {
-    //   pluginPath: './plugins/open-with-plugin',
-    //   isCompatible: true,
-    //   name: '_openWith'
-    // }
   }
 
   async getFromNpm() {
--- a/main/common/plugins.js
+++ b/main/common/plugins.js
@@ -255,11 +255,13 @@ class Plugins {
       pluginPath: './plugins/save-file-plugin',
       isCompatible: true,
       name: '_saveToDisk'
-    }, {
-      pluginPath: './plugins/open-with-plugin',
-      isCompatible: true,
-      name: '_openWith'
-    }];
+    }
+    // , {
+    //   pluginPath: './plugins/open-with-plugin',
+    //   isCompatible: true,
+    //   name: '_openWith'
+    // }
+    ];
   }
 
   async getFromNpm() {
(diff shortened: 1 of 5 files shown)


sindresorhus:
This a good situation to use a draft PR ;)

(You can switch the PR to draft even after creation now)

karaggeorge:
I thought about it but wasn't sure if a draft PR would still get a build and an artifact

sindresorhus:
Haha. That's sounds like a very annoying process. I would honestly just have dropped support for macOS 10.13 in that case if it was my own app.

rdunk:
> Haha. That's sounds like a very annoying process. I would honestly just have dropped support for macOS 10.13 in that case if it was my own app.

@karaggeorge Just want to chip in and say thanks for continuing to support 10.13. Would also have understood if it had been dropped, but it is appreciated.

karaggeorge:
@rdunk No problem. At that point I just needed to know what the bug was caused by for my own sanity as well 😅 . Just out of curiosity though, is there a reason you are still on 10.13 and haven't moved on to one of the later OSes?

rdunk:
Partly laziness, partly "if it ain't broke...".

I heard some (mostly anecdotal) accounts of issues on Mojave/Catalina, and none of the new features were appealing. I never have any problems with my macbook—asidee from the buttteerrfly keeeeyboooaard—so I just haven't got round to it yet. I probably should, though. Maybe this should be the impetus.

PR #828, opened 2020-04-16
Attempt to fix the audio enabled error

karaggeorge:
Only thing I could see causing the audio to not enable if the permission is there, is the promise which fetches the audio devices failing for some reason. So I added some error handling around it to verify.

Also removed some old settings like DnD and Hide Desktop Icons which are now replaced by the plugins

### DIFF
--- a/renderer/containers/preferences.js
+++ b/renderer/containers/preferences.js
@@ -160,7 +160,6 @@ export default class PreferencesContainer extends Container {
           await this.getAudioDevices();
         } catch (error) {
           this.showError(error, {reportToSentry: true});
-          console.log(error);
         }
       }
 
--- a/renderer/containers/preferences.js
+++ b/renderer/containers/preferences.js
@@ -19,6 +19,7 @@ export default class PreferencesContainer extends Container {
     this.systemPermissions = this.remote.require('./common/system-permissions');
     this.plugins = this.remote.require('./common/plugins');
     this.track = this.remote.require('./common/analytics').track;
+    this.showError = this.remote.require('./utils/errors').showError;
 
     const pluginsInstalled = this.plugins.getInstalled().sort((a, b) => a.prettyName.localeCompare(b.prettyName));
 
@@ -158,6 +159,7 @@ export default class PreferencesContainer extends Container {
         try {
           await this.getAudioDevices();
         } catch (error) {
+          this.showError(error, {reportToSentry: true});
           console.log(error);
         }
       }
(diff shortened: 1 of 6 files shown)


renderer/containers/preferences.js
```diff
@@ -155,7 +155,11 @@ export default class PreferencesContainer extends Container {
 
     if (!newValue || await this.systemPermissions.ensureMicrophonePermissions()) {
       if (newValue) {
-        await this.getAudioDevices();
+        try {
```
sindresorhus, line 158: Isn't this moot as `getAudioDevices` cannot fail? Since we already wrap it in try/catch at https://github.com/wulkano/kap/pull/828/files#diff-c8ec885dc71d0b6d7444b52b062ebfa2R214

karaggeorge: This is calling `getAudioDevices` of this container class, which calls the main one and then does some computations. I added that as an additional fail check to make sure whatever goes in that method doesn't stop the toggle for being enabled

sindresorhus: Okey. I think this one too should call `showError(error, {reportToSentry: true});` instead of just logging, so the user can better report to us any possible issues.

karaggeorge: Done 👍 Might be worth looking into using electron-unhandled to make sure we capture everything. Should help with people reporting as well

renderer/containers/preferences.js
```diff
@@ -155,7 +156,12 @@ export default class PreferencesContainer extends Container {
...
+          await this.getAudioDevices();
+        } catch (error) {
+          this.showError(error, {reportToSentry: true});
+          console.log(error);
```
sindresorhus, line 163:
```suggestion
```

karaggeorge:
@sindresorhus I've pushed some more updates fixing the audio devices error. I opened an issue at aperture-node: https://github.com/wulkano/aperture-node/issues/10

But I figured until that can be investigated and fixed, we can use `macos-audio-devices` to get the same list

sindresorhus:
From https://github.com/karaggeorge/macos-audio-devices:

> macOS 10.13 or earlier needs to download the Swift runtime support libraries.

This needs to be documented in the Kap readme now.

karaggeorge:
Do you think we should use the aperture one and only use the other one if it fails? It look like it only fails sometime. Not sure what's causing it but we could have it there in case it does. That way if you are running older than 10.13 you might at least have a chance to get the audio devices without the support libraries

sindresorhus:
> Do you think we should use the aperture one and only use the other one if it fails? It look like it only fails sometime. Not sure what's causing it but we could have it there in case it does. That way if you are running older than 10.13 you might at least have a chance to get the audio devices without the support libraries

Yes, let's do that for now. Also add a comment that it will not be needed when we fix the Aperture issue.

karaggeorge:
@sindresorhus should be good 👍 I'll try to take a look at the Aperture issue during the week

PR #837, opened 2020-04-28
Improve error handling

karaggeorge:
`showErrorBox` is a sync method, which not only blocks, but also doesn't let you interact with the app until the dialog is dismissed. This created a really bad bug when a recording plugin failed in the `willStartRecording` phase, since it would show a dialog with the error, which would be under the cropper, so you couldn't close the dialog but also couldn't dismiss the cropper since the app forces you to interact with the error box first.

Ideally, the error box should be async, and not blocking, so I switched to using `showMessageBox` with type `error`, along with some other optimizations from `electron-unhandled` (including a button to copy the error). 

However, that method blocks main process as well, which means it's not really async either. I've opened an issue for it: https://github.com/electron/electron/issues/23319. There used to be a way to make it async with a callback, but when it was split to sync/async methods they both became blocking. 

This is still better than `showErrorBox` since you can at least interact with the dialog, so it still fixes the bug described above, but for it to work perfectly we would need the above issue resolved.

This PR also adds a call to `cleanup` in case aperture fails to start recording, which will ensure the croppers and tray get reset, taking care of another rare bug.

### DIFF
--- a/main/common/aperture.js
+++ b/main/common/aperture.js
@@ -135,8 +135,6 @@ const startRecording = async options => {
 
   await callPlugins('willStartRecording');
 
-  console.log('Here');
-
   try {
     await aperture.startRecording(apertureOptions);
   } catch (error) {
--- a/main/common/aperture.js
+++ b/main/common/aperture.js
@@ -50,7 +50,7 @@ const callPlugins = async method => Promise.all(recordingPlugins.map(async ({plu
         })
       );
     } catch (error) {
-      showError(error, {title: `Something went wrong while using “${plugin.name}”`});
+      showError(error, {title: `Something went wrong while using the plugin “${plugin.prettyName}”`});
     }
   }
 }));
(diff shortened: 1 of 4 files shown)


main/common/aperture.js
```diff
@@ -50,7 +50,7 @@ const callPlugins = async method => Promise.all(recordingPlugins.map(async ({plu
         })
       );
     } catch (error) {
-      showError(error);
+      showError(error, {title: `Something went wrong while using “${plugin.name}”`});
```
sindresorhus, line 53: Is `plugin.name` `kap-foo` or just `foo`. If just `foo`, I think this should be:
```suggestion
      showError(error, {title: `Something went wrong while using the plugin “${plugin.name}”`});
```

karaggeorge: It's `kap-foo` `plugin.prettyName` is `foo`. I can use either

sindresorhus: Use `plugin.prettyName`

PR #843, opened 2020-04-29
Improve the order of the menu items

sindresorhus:
#### Before

Screenshot 2020-04-29 at 23 28 34

#### After

Screenshot 2020-04-29 at 23 27 30

---

I feel like this is a better order. It's also makes it the same as the `Kap` menu in the editor and also makes it easier to reach "Preferences…" in the cropper:

Screenshot 2020-04-29 at 23 27 41

### DIFF
--- a/main/menus.js
+++ b/main/menus.js
@@ -109,7 +109,7 @@ const getCogMenuTemplate = () => [
   {
     type: 'separator'
   },
-  sendFeedbackItem,
+  preferencesItem,
   {
     type: 'separator'
   },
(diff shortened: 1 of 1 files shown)


karaggeorge:
We have two more items coming for that menu. One is `Clear recordings` or something similar that deletes our temp files, and then a `Enable audio recording` which is a radio shortcut to the preferences toggle. Where do you think those should be in that list?

sindresorhus:
Maybe after "Export History"?

skllcrn:
Clearing recordings may not need to be in the list, but rather grouped alongside options to reset preferences and plugin configurations near the bottom of the Preferences window itself?

PR #844, opened 2020-05-01
Fix input devices logic

karaggeorge:
While adding the workaround for the devices bug, I forgot to return the devices array in the good case scenario.

This fixes this and also uses the workaround at `startRecording` time


karaggeorge:
If the aperture `getDevices` works, you get no inputs. If it doesn't, the fallback return them correctly. I forgot the return statement in the good case scenario

PR #859, opened 2020-06-10
Add Microphone option in cog menu

### LINKED ISSUE #855: Add quick link to "record audio" in the dropdown
image

I constantly have to switch to recording videos with and without audio.

In Quicktime this is an option in the "more dropdown menu" to toggle audio recording on/off.

Kap should have this too!

karaggeorge:
Fixes #855 

Mimics the behavior of the native screen recording util:

CleanShot 2020-06-10 at 10 34 56@2x

Not sure if that's the correct placement for the menu item.

Also, stays in sync with the open cropper to update the ripple effect of the record button, and took that as a chance to rewrite that component to use hooks.

### DIFF
--- a/main/common/constants.js
+++ b/main/common/constants.js
@@ -4,6 +4,5 @@ const supportedVideoExtensions = ['mp4', 'mov', 'm4v'];
 
 module.exports = {
   supportedVideoExtensions,
-  defaultInputDevice: {id: 'SYSTEM_DEFAULT', name: 'System Default'},
-  builtInMicrophoneId: 'BuiltInMicrophoneDevice'
+  defaultInputDeviceId: 'SYSTEM_DEFAULT'
 };
--- a/main/common/constants.js
+++ b/main/common/constants.js
@@ -1,9 +1,9 @@
 'use strict';
 
 const supportedVideoExtensions = ['mp4', 'mov', 'm4v'];
-const defaultInputDevice = {id: 'SYSTEM_DEFAULT', name: 'System Default'};
 
 module.exports = {
   supportedVideoExtensions,
-  defaultInputDevice
+  defaultInputDevice: {id: 'SYSTEM_DEFAULT', name: 'System Default'},
+  builtInMicrophoneId: 'BuiltInMicrophoneDevice'
 };
--- a/main/common/constants.js
+++ b/main/common/constants.js
@@ -1,5 +1,9 @@
 'use strict';
 
 const supportedVideoExtensions = ['mp4', 'mov', 'm4v'];
+const defaultInputDevice = {id: 'SYSTEM_DEFAULT', name: 'System default'};
 
-module.exports = {supportedVideoExtensions};
+module.exports = {
+  supportedVideoExtensions,
+  defaultInputDevice
+};
--- a/main/common/constants.js
+++ b/main/common/constants.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const supportedVideoExtensions = ['mp4', 'mov', 'm4v'];
-const defaultInputDevice = {id: 'SYSTEM_DEFAULT', name: 'System default'};
+const defaultInputDevice = {id: 'SYSTEM_DEFAULT', name: 'System Default'};
 
 module.exports = {
   supportedVideoExtensions,
--- a/main/common/plugins.js
+++ b/main/common/plugins.js
@@ -66,7 +66,7 @@ class Plugins {
         plugin => plugin.recordServices.map(service => ({
           ...service,
           isEnabled: recordPluginServiceState.get(service.title) || false,
-          toggleEnbaled: () => this.enableService(service)
+          toggleEnabled: () => this.enableService(service)
         }))
       )
     );
--- a/main/utils/devices.js
+++ b/main/utils/devices.js
@@ -1,7 +1,7 @@
 'use strict';
 
-const {getInputDevices} = require('macos-audio-devices');
-const {audioDevices} = require('aperture');
+const audioDevices = require('macos-audio-devices');
+const aperture = require('aperture');
 
 const {showError} = require('./errors');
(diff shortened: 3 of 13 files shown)


karaggeorge:
> How are the microphone devices sorted? I think the internal MacBook microphone should be right after None.

Right now they are not. They are just displayed the way they come from the API. Should the only sorting be bringing the internal microphone to the top? Or have any other rules as well

main/common/plugins.js
```diff
@@ -66,7 +66,7 @@ class Plugins {
         plugin => plugin.recordServices.map(service => ({
           ...service,
           isEnabled: recordPluginServiceState.get(service.title) || false,
-          toggleEnbaled: () => this.enableService(service)
+          toggleEnabled: () => this.enableService(service)
```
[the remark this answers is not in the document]
karaggeorge: This is the handler to toggle the enabled or not status. The field above it is the one containing wether it is enabled or not. So isn't this fine as the action verb?

sindresorhus:
> Right now they are not. They are just displayed the way they come from the API. Should the only sorting be bringing the internal microphone to the top? Or have any other rules as well

For now, yes.

karaggeorge:
Is there a way to figure out which one it is other than by name? On my machine it shows as `MacBook Pro Microphone`, but I think I've seen it as `Built-in Microphone` or something like that as well

sindresorhus:
Maybe https://developer.apple.com/documentation/coreaudio/kaudiohardwarepropertydefaultinputdevice ?

karaggeorge:
That is the default device the user has selected as their input device in the settings. I use it here as well: https://github.com/karaggeorge/macos-audio-devices/blob/master/Sources/audio-devices/AudioDevices.swift#L109

So if the user has their headphones connected and set as preferred, it'd be set to that.

I can use that, especially since it's already exposed from `macos-audio-devices` which we already have as a dependency, it just won't always be the built-in microphone

We could also at some point have an option that says use the system default device automatically, so you wouldn't have to switch between built-in and let's say AirPods if you use both frequently. The system remembers your preferences so when you connect your device it switches automatically, but for Kap you would have to manually go and change it.

cc @skllcrn

sindresorhus:
> I can use that, especially since it's already exposed from macos-audio-devices which we already have as a dependency, it just won't always be the built-in microphone

👍  That sounds even better than just prioritizing the built-in.

> Maybe options could be:
> None
System default
...specific input devices

That would be the optimal solution, yes.

karaggeorge:
Should the option just say `System default` or should it be `System default (${current default name})`. I feel like the latter has more info but it might be a bit lengthy

sindresorhus:
> System default (${current default name})

👍 , but `default` => `Default`.

karaggeorge:
@sindresorhus this is ready for another review. I added a default device that is available in the preferences and the new menu list. Selecting it sets the `SYSTEM_DEFAULT` id in the settings, and the device id is pulled the moment before aperture starts recording. I also set that as the new default if the selected audio device is no longer connected when the app starts or a recording starts.

This way, if you had a mic connected, then start the app, you'll most likely get the built-in microphone instead of getting a random input that comes first in the list (like Soundflower for me)

Finally, I did find a way to sort the built-in mic to the top. Even though the name might change betwenn macos versions, I noticed that the uid is actually `BuiltInMicrophoneDevice` for both my macos versions, so if I find a device with that id I bring it to the top. Hopefully that works for all versions

main/utils/devices.js
```diff
@@ -18,16 +18,28 @@ const getAudioDevices = async () => {
       const Sentry = require('./sentry');
       Sentry.captureException(new Error(`devices is not an array: ${JSON.stringify(devices)}`));
 
-      return (await audioDevices.getInputDevices()).map(device => ({id: device.uid, name: device.name}));
+      return (await audioDevices.getInputDevices()).map(device => ({id: device.uid, name: device.name})).sort(devicesSort);
```
sindresorhus, line 22: Shouldn't this be changed in `macos-audio-devices` instead?

main/common/constants.js
```diff
@@ -1,9 +1,9 @@
 'use strict';
 
 const supportedVideoExtensions = ['mp4', 'mov', 'm4v'];
-const defaultInputDevice = {id: 'SYSTEM_DEFAULT', name: 'System Default'};
 
 module.exports = {
   supportedVideoExtensions,
-  defaultInputDevice
+  defaultInputDevice: {id: 'SYSTEM_DEFAULT', name: 'System Default'},
+  builtInMicrophoneId: 'BuiltInMicrophoneDevice'
```
sindresorhus, line 8: I think `macos-audio-devices` should expose this. And maybe also a `isBuiltIn` property on the device info.

main/common/plugins.js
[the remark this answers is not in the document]
sindresorhus: Yeah, sorry, my mistake.

sindresorhus:
In "System Default", I don't see the current default in parens as discussed.

main/common/constants.js
karaggeorge: Well, I'm not sure if that is actually right. I've just noticed it in my machines. I don't know if there's a way to know for sure what device is built in

main/utils/devices.js
karaggeorge: Hmm, but the package in general, doesn't really have a use-case for sorting them. I feel like that's a Kap-specific requirement

karaggeorge:
@sindresorhus Updated this to sort built-in devices to the top and include the current system default in parens. I also tested it in both cropper and preferences along with recording and it seems to be working great.

Updated screenshots:
Screen Shot 2020-06-25 at 3 52 51 PM
Screen Shot 2020-06-25 at 3 56 00 PM

sindresorhus:
Yup. Works well 👍

PR #872, opened 2020-06-25
Allow plugins to specify `macosVersions` they support

### LINKED ISSUE #841: Let plugins define the macOS version they support
We currently support `"kapVersion": ">=3.2.0" in package.json.

I'm now thinking that we should namespace our custom package.json properties:

```json
{
	"kap": {
		"version": ">=3.2.0",
		"macosVersion": ">=10.15.4"
	}
}

karaggeorge:
Fixes #841

Creates the `kap` namespace for package.json and adds `macosVersion` as an option.

I kept backwards compatibility with `kapVersion` with a deprecated message. We can remove once we know that all plugins have migrated


sindresorhus:
Don't forget to update your own packages: https://github.com/search?l=JSON&q=%22kapVersion%22&type=Code

PR #873, opened 2020-06-26
Only start editor playback on load if the editor is focused

### LINKED ISSUE #853: Don't start auto-playing after stopping recording
Not much more to say besides the issue title – it'd be nice to have this option. I'd rather manually press play to begin previewing/watching from the start of my recording (especially if it involves audio) instead of it starting automatically.

karaggeorge:
Fixes #853 

This is pretty hard to achieve 😂  I was only able to test it with a very long recording that takes a split second to load, so I can blur between the editor opening and the video starting.

### DIFF
--- a/renderer/containers/video.js
+++ b/renderer/containers/video.js
@@ -1,4 +1,5 @@
 import {Container} from 'unstated';
+import electron from 'electron';
 
 export default class VideoContainer extends Container {
   state = {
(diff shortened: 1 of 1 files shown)


frozenpandaman:
Really cool solution, will test it out later – thank you @karaggeorge!

PR #874, opened 2020-06-26
Persist format-specific fps options

### LINKED ISSUE #861: Persist FPS on a format basis
We persist fps across editors and launch, but we only remember one value. Most people don't usually export with the same fps for every format.

For example, I usually export mp4 with 60fps, and gif with 15fps. So, even though we persist fps, I still have to change it quite often.

I think we should remember the last used fps for each format, same as we do for plugin usage etc

karaggeorge:
Fixes #861

### DIFF
--- a/main/export-options.js
+++ b/main/export-options.js
@@ -22,7 +22,22 @@ const exportUsageHistory = new Store({
 const fpsUsageHistory = new Store({
   name: 'fps-usage-history',
   schema: {
-    fps: {
+    apng: {
+      type: 'number',
+      minimum: 0,
+      default: 60
+    },
+    webm: {
+      type: 'number',
+      minimum: 0,
+      default: 60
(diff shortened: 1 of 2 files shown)


sindresorhus:
In the editor, I have "GIF" active, I set FPS to 10, then switch to MP4 and then back to GIF again and it's no longer 10. I think it should be persisted there too.

karaggeorge:
I had it persisting every time you actually export with that value, thinking that in normal use you wouldn't be switching back and forth between formats. Should I persist every time the fps value changes for each format?

sindresorhus:
> Should I persist every time the fps value changes for each format?

It's probably an edge-case, but that's what I would expect as a user.

karaggeorge:
@sindresorhus this is ready for another look 😄

PR #877, opened 2020-07-02
Allow plugin control via deep links

karaggeorge:
Adds ability to install/configure plugins via the following deep links:

`kap://install-plugin/{plugin name}`

Pops up preferences on the discover tab scrolls to the plugin, and triggers an installation. If it's already installed, it just scrolls to it.

`kap://configure-plugin/{plugin name}`

Pops up preferences, scrolls to the plugin and opens the config window

Also fixed some small issues, including a circular dependency I introduced a few PRs ago which was causing the built app not to run

To test this build, after you've downloaded and installed it, click here:

[Install](https://getkapco-git-add-deep-link-assets.wulkano1.vercel.app/deep-link/install/kap-recording-time)

And then to configure:

kap://configure-plugin/kap-record-time

EDIT:
seems like GH doesn't support links with custom schemes, so copy-paste the links underneath for now. We can probably create a small lamda to re-route to the deep link

~These badges are some examples I generated with shield.io, but we can make our own custom images, host them under `https://getkap.co/buttons/{install|configure}` so we can add buttons like that to the plugin readmes~

https://github.com/wulkano/getkap.co/pull/36

### DIFF
--- a/main/utils/errors.js
+++ b/main/utils/errors.js
@@ -10,6 +10,7 @@ const showError = async (error, {title, reportToSentry} = {}) => {
 
   console.error(error);
   if (reportToSentry) {
+    // Imported here to avoid circular dependency
     const Sentry = require('./sentry');
     Sentry.captureException(ensuredError);
   }
--- a/main/utils/errors.js
+++ b/main/utils/errors.js
@@ -2,7 +2,6 @@
 
 const {dialog, clipboard} = require('electron');
 const ensureError = require('ensure-error');
-const Sentry = require('./sentry');
 const cleanStack = require('clean-stack');
 
 const showError = async (error, {title, reportToSentry} = {}) => {
@@ -11,6 +10,7 @@ const showError = async (error, {title, reportToSentry} = {}) => {
 
   console.error(error);
   if (reportToSentry) {
+    const Sentry = require('./sentry');
     Sentry.captureException(ensuredError);
   }
 
--- a/renderer/containers/preferences.js
+++ b/renderer/containers/preferences.js
@@ -33,11 +33,6 @@ export default class PreferencesContainer extends Container {
       isMounted: true
     });
 
-    if (this.state.target && pluginsInstalled.some(plugin => plugin.name === this.state.target.name)) {
-      this.openTarget(this.state.target);
-      this.setState({target: undefined});
-    }
-
     if (this.settings.store.recordAudio) {
       this.getAudioDevices();
     }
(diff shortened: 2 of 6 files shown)


main/utils/errors.js
```diff
@@ -11,6 +10,7 @@ const showError = async (error, {title, reportToSentry} = {}) => {
 
   console.error(error);
   if (reportToSentry) {
+    const Sentry = require('./sentry');
```
karaggeorge, line 13: This was moved here to avoid a circular dependency

sindresorhus:
> Pops up preferences on the discover tab scrolls to the plugin, and triggers an installation.

Is it safe to trigger to installation automatically? I'm honestly not sure. Just wanted to bring it up.

main/utils/errors.js
sindresorhus: Can you add a code comment about that so it's not accidentally moved back in the future?

renderer/containers/preferences.js
```diff
@@ -60,7 +60,45 @@ export default class PreferencesContainer extends Container {
     this.setState(updates);
   }
 
-  setNavigation = ({category, tab}) => this.setState({category, tab})
+  scrollIntoView = (tabId, pluginId) => {
```
sindresorhus, line 63: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView ?

sindresorhus:
> seems like GH doesn't support links with custom schemes, so copy-paste the links underneath for now. We can probably create a small lamda to re-route to the deep link

That's a good workaround, yes. Or a Vercel function.

karaggeorge:
> Is it safe to trigger to installation automatically? I'm honestly not sure. Just wanted to bring it up.

I had the same thought initially but I don’t see how it would not be. Another option I thought of would be to open, scroll into view and have a yes/cancel modal saying ‘would you like to install x’

So you’re sure that the plugin linked is what you think you’re installing

karaggeorge:
It now shows a dialog like this to confirm before installing:

Screen Shot 2020-07-03 at 9 11 20 AM

renderer/containers/preferences.js
```diff
@@ -60,7 +60,62 @@ export default class PreferencesContainer extends Container {
...
+
+        const buttonIndex = this.remote.dialog.showMessageBoxSync(this.remote.getCurrentWindow(), {
+          type: 'question',
+          buttons: [
+            'Install',
+            'Cancel'
+          ],
+          defaultId: 0,
+          cancelId: 1,
+          message: `Do you want to install ${target.name}?`
```
sindresorhus, line 92:
```suggestion
          message: `Do you want to install the “${target.name}” plugin?`
```

sindresorhus:
Tried it out and works great. I really like how it even scroll down and shows the plugin.

PR #878, opened 2020-07-03
Unit tests the conversion

### LINKED ISSUE #393: Unit test the video conversion and plugin system
They are quite separate components and it should be possible to test them individually.

This is not the same as #27, which is about integration tests.

I'll set up some testing with [AVA](https://github.com/avajs/ava) when I have a chance.

karaggeorge:
Fixes #393

Adds unit tests of the conversion process using AVA. It actually runs ffmpeg with different options and checks the meta data of the resulting files to ensure our methods have the desired results.

It already helped me find two bugs, one with the experimental `vorbis` audio encoder we use for webm which would break if the audio wasn't dual channel, and one with a case where the speed is slow enough to be parsed to `0`.

I'll soon add some tests (probably separate PR) for the editing plugins which take effect in the same file, and then the rest of the plugins, even though I'm not sure how we can test recording and export plugins other than making sure the methods are being called in the right order.

This will run on CircleCI automatically, but we think it's heavy (ffmpeg) we could run these in GH actions if we have free minutes. When we add more tests, we can also split them in non-ffmpeg unit tests, ffmpeg unit tests and integration tests that we can have running in separate commands.

```
~/workspace/kap tests*
❯ time ava

  ✔ mp4: retina with sound (10.9s)
  ✔ mp4: retina without sound (2s)
  ✔ mp4: non-retina (2.8s)
  ✔ webm: retina with sound (14.2s)
  ✔ webm: retina without sound (4.1s)
  ✔ webm: non-retina (7.7s)
  ✔ apng: retina (21.3s)
  ✔ apng: non-retina (3.4s)
  ✔ gif: retina (4s)
  ✔ gif: non-retina (4.3s)
  ─

  10 tests passed
ava  161.83s user 5.54s system 220% cpu 1:15.97 total
```

Didn't set up coverage, but I did run it with `nyc` and the file is pretty covered other than the cases where the processes crash or are cancelled, which we can cover later if needed
```
------------|---------|----------|---------|---------|---------------------------------------------
File        | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
------------|---------|----------|---------|---------|---------------------------------------------
All files   |   70.49 |    45.45 |   54.55 |      70 |
 convert.js |   70.49 |    45.45 |   54.55 |      70 | 35-39,81-85,103,118,143,206,238,245,254-320
------------|---------|----------|---------|---------|---------------------------------------------
```


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     macos:
-      xcode: '11.4.0'
+      xcode: '12.0.0'
```
karaggeorge, line 5: Had to bump because of:

```
error ava@3.9.0: The engine "node" is incompatible with this module. Expected version ">=10.18.0 <11 || >=12.14.0 <12.17.0 || >=12.17.0 <13 || >=14.0.0". Got "13.13.0"
 
error Found incompatible module.
```

If we can't bump this for some reason I can set it up with GH actions

test/convert.js
```diff
@@ -0,0 +1,294 @@
...
+mockImport('./common/analytics', 'analytics');
+mockImport('./service-context', 'service-context');
+
+const {convertTo} = require('../main/convert');
+
+test.afterEach.always(t => {
+  if (t.context.outputPath && fs.existsSync(t.context.outputPath)) {
+    fs.unlink(t.context.outputPath, () => {});
+  }
+});
```
sindresorhus, line 28:
```suggestion
test.afterEach.always(async t => {
  if (t.context.outputPath && fs.existsSync(t.context.outputPath)) {
    await fs.promises.unlink(t.context.outputPath);
  }
});
```

Or maybe just use the sync function, since it's unlikely to have much overhead and sync is usually faster.

sindresorhus:
This is great!

Maybe we should mention in the contribution guidelines that new features should come with tests whenever possible?

karaggeorge:
> Maybe we should mention in the contribution guidelines that new features should come with tests whenever possible?

I would say let's wait for the next PR. Right now the only tests you can really run are for the conversion system. I plan on trying to add some frontend pure react tests, and maybe some integration ones. I think once we have the setup and a few tests for each, it'll be easier for people to add more

PR #879, opened 2020-07-04
Better error handling

karaggeorge:
Creates a new dialog window (much more flexible and truly async compared to system dialog)

It might seem a bit more complex than it needs to be for this PR, but I needed the extra functionality for another PR coming soon.

Uses that to create better error handling. Logic goes as follows:

For each error,
- If it's related to a plugin, show a dialog, allow copying and if we know the repo of the plugin, offer a button that opens an issue with the details of the error on that plugin's repo.
- If it's not related:
  - If we're offline or analytics are not allowed, show dialog with ability to copy
  - Otherwise:
    - Send to sentry, grab the eventId
    - Use the eventId through a serveless function to find the related Sentry issue (takes about 8 seconds). 
    - Check if we have a GH issue with that Sentry issue linked to it
    - If we do, offer a button to open that issue
    - If we don't, offer a button to create it with a pre-filled body along with the Sentry-related ids that will allow the issue to be tracked in the future

For right now this is the serverless function: https://github.com/wulkano/kap-sentry-tracker

It's linked to a test repo I made for now. Also, it checks for 3 different types of Sentry issue id, and it also looks only at issues with the sentry label. I did that to limit the queries to the GH api, but I think when users create an issue they can't automatically assign a label, so I might need to remove that, or we might need a webhook that checks if a new PR is opened and it has those sentry ids, it labels it automatically.

Also, added system colors and accent color as css vars that we can use in the UI from now on


main/utils/errors.js
```diff
@@ -1,33 +1,161 @@
...
+          });
+        } else {
+          updateUi({
+            buttons: [
+              ...mainButtons,
+              {
+                label: 'Open Issue',
+                action: () => openNewGitHubIssue({
+                  user: 'karaggeorge',
+                  repo: 'kap-test-playground',
```
karaggeorge, line 128: Used for testing, these will change

renderer/components/dialog/icon.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+export default () => {
+  return (
+    <div>
+      <img src="https://getkap.co/static/favicon/kap.svg"/>
```
karaggeorge, line 6: This needs to change to the app icon. Wonder if there's a method to get the current app icon (stable/beta etc)

karaggeorge:
I tried using this to get the Caution icon but I it's super low quality, so when I size it up it's looks pixelated: https://www.electronjs.org/docs/api/native-image#nativeimagecreatefromnamedimageimagename-hslshift-macos

renderer/components/dialog/icon.js
sindresorhus: Try the string `NSApplicationIcon` for https://www.electronjs.org/docs/api/native-image#nativeimagecreatefromnamedimageimagename-hslshift-macos

sindresorhus:
> I tried using this to get the Caution icon but I it's super low quality, so when I size it up it's looks pixelated: electronjs.org/docs/api/native-image#nativeimagecreatefromnamedimageimagename-hslshift-macos

Did you try the `scaleFactor` parameter? https://www.electronjs.org/docs/api/native-image#imagetopngoptions

main/dialog.js
```diff
@@ -0,0 +1,92 @@
...
+    useContentSize: true,
+    webPreferences: {
+      nodeIntegration: true
+    }
+  });
+
+  loadRoute(dialogWindow, 'dialog');
+
+  let buttons;
+  let actionTaken;
```
sindresorhus, line 29: `isActionTaken`

main/utils/errors.js
```diff
@@ -1,34 +1,161 @@
...
+
+const MAX_RETRIES = 20;
+
+const getSentryIssue = async (eventId, tries = 0) => {
+  if (tries > MAX_RETRIES) {
+    return;
+  }
+
+  try {
+    const {body} = await got.get(`https://kap-sentry-tracker.vercel.app/api/event/${eventId}`, {json: true});
```
sindresorhus, line 22: For future readers of this code, I would add a short comment about what this endpoint does.

main/utils/errors.js
```diff
@@ -1,34 +1,161 @@
...
+
+    if (body.pending) {
+      await delay(1000);
+      return getSentryIssue(eventId, tries + 1);
+    }
+
+    return body;
+  } catch {
+
```
sindresorhus, line 31: Handle this?

sindresorhus:
I get this in the terminal:

> (node:281) electron: 'alternate-selected-control-text' is deprecated as an input to getColor.  Use 'selected-content-background' instead.

sindresorhus:
I tried this by throwing an error here: https://github.com/sindresorhus/kap-gifski/blob/b1fde8e3e565ed8e5e47f6ec78084fff03bb1096/index.js#L28

Just `throw new Error('Unicorn');`.

But the error dialog never shows. All I'm seeing is the export being stuck:

Screenshot 2020-07-05 at 05 37 52

main/utils/errors.js
```diff
@@ -1,34 +1,161 @@
...
+  const buttons = [...mainButtons];
+
+  if (isOnline && settings.get('allowAnalytics')) {
+    const eventId = Sentry.captureException(ensuredError);
+    const sentryIssuePromise = getSentryIssue(eventId);
+
+    message = 'Reporting this issue will help us track it better and resolve it faster';
+
+    buttons.push({
+      label: 'Collect info and report',
```
sindresorhus, line 100:
```suggestion
      label: 'Collect Info and Report',
```

main/utils/errors.js
```diff
@@ -1,34 +1,161 @@
...
+  const Sentry = require('./sentry');
+
+  let message;
+  const buttons = [...mainButtons];
+
+  if (isOnline && settings.get('allowAnalytics')) {
+    const eventId = Sentry.captureException(ensuredError);
+    const sentryIssuePromise = getSentryIssue(eventId);
+
+    message = 'Reporting this issue will help us track it better and resolve it faster';
```
sindresorhus, line 97:
```suggestion
    message = 'Reporting this issue will help us track it better and resolve it faster.';
```

main/utils/errors.js
```diff
@@ -1,34 +1,161 @@
...
+
+    buttons.push({
+      label: 'Collect info and report',
+      activeLabel: 'Collecting info…',
```
sindresorhus, line 101:
```suggestion
      activeLabel: 'Collecting Info…',
```

main/utils/errors.js
```diff
@@ -1,34 +1,161 @@
...
+      action: async (_, updateUi) => {
+        const issue = await sentryIssuePromise;
 
-  if (response === 1) {
-    clipboard.writeText(`${errorTitle}\n${cleanStack(ensuredError.stack)}`);
+        if (!issue || issue.error) {
+          updateUi({
+            message: 'Something went wrong while collecting the information',
```
sindresorhus, line 107:
```suggestion
            message: 'Something went wrong while collecting the information.',
```

main/utils/errors.js
```diff
@@ -1,34 +1,161 @@
...
+  const title = customTitle || ensuredError.name;
+  const detail = cleanStack(ensuredError.stack, {pretty: true});
 
-  console.error(error);
-  if (reportToSentry) {
-    // Imported here to avoid circular dependency
-    const Sentry = require('./sentry');
-    Sentry.captureException(ensuredError);
+  const mainButtons = [
+    'OK',
```
sindresorhus, line 57:
```suggestion
    'Don't Report',
```

sindresorhus: Or

```suggestion
    'Don't Report 😭',
```

🤣

main/dialog.js
```diff
@@ -0,0 +1,92 @@
...
+const {ipcMain: ipc} = require('electron-better-ipc');
+
+const loadRoute = require('./utils/routes');
+
+const showDialog = options => new Promise(resolve => {
+  const dialogWindow = new BrowserWindow({
+    width: 1,
+    height: 1,
+    resizable: false,
+    minimizable: false,
```
sindresorhus, line 13:
```suggestion
    minimizable: false,
    maximizable: false,
```

sindresorhus:
The dialog becomes a bit wide. It should have a max width. I think we should also do something about the stack trace. It's too long. Maybe slice off `/Users/sindresorhus/Library/Application Support/Kap/plugins/node_modules/`?

Screenshot 2020-07-05 at 05 43 30

sindresorhus:
I tried to collect and report the error, but I just got:

> Something went wrong while collecting the information

renderer/pages/dialog.js
```diff
@@ -0,0 +1,123 @@
...
+        html {
+          font-size: 62.5%;
+        }
+
+        html,
+        body {
+          margin: 0;
+          width: 100vw;
+          height: 100vh;
+          font-family: -apple-system, BlinkMacSystemFont, 'Helvetica Neue', sans-serif;
```
sindresorhus, line 80:
```suggestion
          font-family: system-ui, 'Helvetica Neue', sans-serif;
```

main/utils/errors.js
sindresorhus: We need to mention above what kind of information it will collect. Otherwise, I doubt anyone will want to click it.

karaggeorge: How do you think it should be worded? All we're doing is pretty much waiting for Sentry to categorize the event under an Issue and then looking for matching GH issues

renderer/components/dialog/icon.js
karaggeorge: Screen Shot 2020-07-05 at 10 10 33 PM

This is what I get while in dev and built.

Also, this is the best quality I've been able to get using `scaleFactor`

karaggeorge:
@sindresorhus I think this is ready for another look

renderer/components/dialog/icon.js
sindresorhus: > Also, this is the best quality I've been able to get using scaleFactor

You should open an Electron issue about that. Most likely a bug.

sindresorhus:
When I put the following:

```js
Promise.resolve().then(() => {
		setTimeout(() => {
			throw new Error('Unicorn');
		}, 1000);
	});
```

here:
https://github.com/sindresorhus/kap-gifski/blob/b1fde8e3e565ed8e5e47f6ec78084fff03bb1096/index.js#L9

The whole export window just freezes and I get a beachball.

karaggeorge:
I've been looking into those steps and I consistently get the app not responding. I haven't been able to locate where it's happening other than it has something to do with the export process. I might have to rewrite some of that logic, as it's a bit complex with multiple parallel promises. If that fixes it I might open it as a new PR as it'll be pretty disconnected from this

karaggeorge:
Never mind. Switching to a new branch didn't break the app any more. It ended up being a one-letter typo which would throw an error in:

```
process.on('uncaughtException', error => {
    showError(error, {title: 'Unhandled Error'});
  });
```

Which would then probably get caught by the same handler and cycle through.

This error the way you mentioned it doesn't break the conversion for me. I'm not sure if that's expected but I think it is since the promise is created but not awaited or anything, so it's just another unhandled error.

For me after this last commit, I get the error dialog, but the export completes successfully

sindresorhus:
Yes, works now.

sindresorhus:
When clicking the "Collect info and report" button, the width of the button should not change when it changes to "Collecting info…".

sindresorhus:
Submitting never works for me. Is it supposed to work while running with `npm start`?

Screenshot 2020-07-06 at 22 35 50

sindresorhus:
> I think we should also do something about the stack trace. It's too long. Maybe slice off /Users/sindresorhus/Library/Application Support/Kap/plugins/node_modules/?

Any thoughts on this?

sindresorhus:
Also notice in the screenshot here https://github.com/wulkano/Kap/pull/879#issuecomment-654277624 that there's too much space below the buttons.

karaggeorge:
> I think we should also do something about the stack trace. It's too long. Maybe slice off /Users/sindresorhus/Library/Application Support/Kap/plugins/node_modules/?

> Any thoughts on this?

I thought this would address that: https://github.com/wulkano/Kap/pull/879/files#diff-1ff302b55f76f9d95eefef19b4f508fbR41-R42

> When clicking the "Collect info and report" button, the width of the button should not change when it changes to "Collecting info…".

How do you think I should handle that, since the length of the text changes? I could make all buttons a set width but I don't think that's the right way either.

> Submitting never works for me. Is it supposed to work while running with npm start?

It should 🤔  I added and error handle for when the "collecting info" fails. Did no other window pop up? 

> Also notice in the screenshot here #879 (comment) that there's too much space below the buttons.

Yup, I'll fix that

sindresorhus:
> How do you think I should handle that, since the length of the text changes? I could make all buttons a set width but I don't think that's the right way either.

It's not a big thing. It's just a little bit jarring for users when UI moves around like that. I would just at load set the current inferred width as a hard-coded width.

sindresorhus:
> It should 🤔 I added and error handle for when the "collecting info" fails. Did no other window pop up?

Nope. Nothing else happened.

karaggeorge:
> It's not a big thing. It's just a little bit jarring for users when UI moves around like that. I would just at load set the current inferred width as a hard-coded width.

I think it'll work for this case but we'll have to keep an eye on it, since the `activeLabel` can be something completely different from the original, so it might be way wider.

> Nope. Nothing else happened.

Just remembered we disable sentry in development so the issue will never be found. I had disabled it while I was testing. I need to add a check if it's development to skip that whole check, since it's wasted calls to the endpoint.

To test it I would either build it (or download the artifact) or just comment out the `is.development` check in `sentry.js`

sindresorhus:
> To test it I would either build it (or download the artifact) or just comment out the is.development check in sentry.js

Works perfectly now.

sindresorhus:
> I think it'll work for this case but we'll have to keep an eye on it, since the activeLabel can be something completely different from the original, so it might be way wider.

Feel free to defer that to later. Was just an observation as a user.

main/utils/errors.js
```diff
@@ -1,34 +1,172 @@
...
+  const pluginsPath = path.join(app.getPath('userData'), 'plugins', 'node_modules');
+  return cleanStack(error.stack, {pretty: true, basePath: isPluginError && pluginsPath});
+};
+
+const getIssueBody = (title, errorStack, sentryTemplate) => `
+${sentryTemplate}<!--
+Thank you for helping us test Kap. Your feedback helps us make Kap better for everyone!
+-->
+
+**macOS version:**    ${os.release()} (darwin)
```
sindresorhus, line 50: Super minor, but we could use https://github.com/sindresorhus/macos-release to use a more user-friendly version here. Not Darwin.

main/utils/errors.js
```diff
@@ -1,34 +1,172 @@
...
+**Kap version:**      ${app.getVersion()}
+\`\`\`
```
sindresorhus, line 52:
```suggestion

\`\`\`
```

karaggeorge:
@sindresorhus This should be good to check one more time. I've also updated the repo info in the code, and I've updated the endpoint to point to Kap.

Also, I added a webhook to the repo, for when an issue is created with the template above, it'll auto label it with Sentry, since we the label query param of open issue request only works if you are part of the org

sindresorhus:
Works great! 👍

sindresorhus:
Totally unrelated, but maybe we should set up this in Sentry?

Screenshot 2020-07-10 at 15 15 17

PR #880, opened 2020-07-10
Fix select config bug

karaggeorge:
Fixes a small bug where if the current value is not in the available options, it'll show an ugly react error. This way, it'll default to the first option.

This happens semi-often in a new plugin I'm working on. Only reason I came across it

Also it adds permissions and string for accessing the camera, also needed for the same plugin. Followed the same generic message type as apps like iTerm use.


karaggeorge:
@sindresorhus This is pretty small. Can we get this in as well? It'll allow me to test the whole camera permissions issue with the signed/notarized app

PR #892, opened 2020-07-10
Add sentry releases

karaggeorge:
Adds sentry releases. This should help us get better view into what commit fixed/caused issues.

Releases will be in the format of `kap@3.2.0`

Required env vars have been added to CircleCI, but I haven't tested it. I might run it a few times to make sure it works

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,14 +13,14 @@ jobs:
           path: dist/Kap.dmg
   sentry-release:
     docker:
-      - image: circleci/node:4.8.2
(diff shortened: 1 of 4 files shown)


karaggeorge:
From what I understand as long as the release we pass to Sentry.init and the one we use in the build match, Sentry will be able to associate any issues with that along with the commits it gets from the build step. I haven't set this up before, so if something doesn't work we can fix in future releases

sindresorhus:
Looks good code-wise. I haven't tested this though.

PR #899, opened 2020-07-21
Recover recordings after unexpected exit/crash

### LINKED ISSUE #762: Temp files not deleted when video preview is closed
**macOS version:**
10.15

**Kap version:**
3.0.0

#### Steps to reproduce

1. Record a video
2. save it or not
3. close the preview window
4. open your temp folder (/private/var/folders/yy/xxxxxxxxxxxx/T/)

#### Current behaviour
The video is still on the temp folder (root above) as well as on subfolders with the exported format.

#### Expected behaviour
Remove the temp files when the video is discarded. We're talking about videos that could easily grow to Gb, and mostly Mac users don't restart their computer often enough to trigger the cleaning process

#### Workaround
Access the temp folder and delete the files manually

### LINKED ISSUE #871: Changing temporary save location?
**macOS version:** 10.14.6
**Kap version:** 3.2.2, installed via Homebrew cask

It'd be great if Kap were customizable to be able to save files in a location other than `/private/var/folders`. I often record a bunch of GIFs/videos at once and want to save/export them all when done, but my computer has accidentally crashed in the middle of this and I've lost all of my progress (since that folder is cleared on computer restart) – so it'd be nice to be able to change the location where "temporary" (not yet exported or permanently saved) mp4s are stored.

karaggeorge:
Inspired by https://github.com/wulkano/aperture-node/pull/13#issuecomment-653593766

Instead of allowing users to save files in a custom directory, this adds better handling of our temp files. It allows us to track all our recordings, so the user can remove them if they want.

Also it tracks an active recording, so if Kap crashes, we can recover it (hopefully playable or fixable) and direct the user to the file. Also allows for plugins to clean up after an unexpected crash.

Fixes #871 (or the need for it anyway)
Fixes #762


test/mocks/electron.js
```diff
@@ -1,10 +1,23 @@
 'use strict';
+const sinon = require('sinon');
+const tempy = require('tempy');
+const path = require('path');
 
+const temporaryDir = tempy.directory();
+
+process.env.TZ = 'America/New York';
```
sindresorhus, line 8: This time zone identifier is invalid.

```suggestion
process.env.TZ = 'America/New_York';
```

docs/plugins.md
```diff
@@ -240,6 +240,7 @@ The record service is a plain object defining some metadata and hooks:
 - `didStartRecording`: Function that is called after the recording starts. [Read more below.](#hooks)
 - `didStopRecording`: Function that is called after the recording stops. [Read more below.](#hooks)
 - `willEnable`: Function that is called when the user enables the service. [Read more below.](#hooks)
+- `cleanUp`: Function that is called if Kap exits unexpectedly, without the `didStopRecording` hook being called. This hook will only receive the `persistedState` object from the `state` passed to the rest of the hooks. Use this to clean up any effects introduced when the recording started and don't automatically clear out once Kap stops.
```
sindresorhus, line 243:
```suggestion
- `cleanUp`: Function that is called if Kap exited unexpectedly last time it was run (for example, if it crashed), without the `didStopRecording` hook being called. This hook will only receive the `persistedState` object from the `state` passed to the rest of the hooks. Use this to clean up any effects introduced when the recording started and don't automatically clear out once Kap stops.
```

Maybe also provide a practical example of what it can be used for?

docs/plugins.md
```diff
@@ -286,7 +287,8 @@ You can use this to check if you have enough permissions for the service to work
 
 The hook functions receive a `context` argument with some metadata and utility methods.
 
-- `.state`: A plain empty object that will be shared and passed to all hooks in the same recording process. It can be useful to persist data between the different hooks.
+- `.state`: An object that will be shared and passed to all hooks in the same recording process. It can be useful to persist data between the different hooks.
+    - `state.persistedState`: An object under `state` which should only contain serializable fields. It will be passed to the `cleanUp` hook if Kap doesn't shut down correctly. Use this to store fields necessary to clean up remaining effects.
```
sindresorhus, line 291:
```suggestion
    - `state.persistedState`: An object under `state` which should only contain serializable fields. It will be passed to the `cleanUp` hook if Kap didn't shut down correctly last time it was run. Use this to store fields necessary to clean up remaining effects.
```

main/common/aperture.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const createAperture = require('aperture');
+const exitHook = require('async-exit-hook');
```
sindresorhus, line 4: Not sure we want to use this. It's unmaintained and has some issues.

main/common/aperture.js
```diff
@@ -35,6 +37,16 @@ const setRecordingName = name => {
   recordingName = name;
 };
 
+const serializeEditPluginState = () => {
+  return recordingPlugins.reduce((acc, {plugin, service}) => ({
```
sindresorhus, line 41: A `for-of` loop would be much more readable than using `reduce`. IMHO.

main/common/aperture.js
```diff
@@ -200,11 +220,21 @@ const stopRecording = async () => {
...
+    // Also since the editor never opens, we show the dialog on next start so the user can still retrieve it
+    if (!isAppExiting) {
+      stopCurrentRecording(recordingName);
+    }
   }
 };
 
+exitHook(callback => {
+  stopRecording(true).then(callback);
+});
```
sindresorhus, line 236:
```suggestion
exitHook(async callback => {
  await stopRecording(true);
  callback()
});
```

main/recording-history.js
```diff
@@ -0,0 +1,304 @@
...
+    date: recording.date
+  });
+
+  return showDialog({
+    title: 'Kap didn\'t shut down correctly',
+    detail: 'Looks like Kap crashed during a recording. We were able to locate the file and it appears to be playable.',
+    buttons: [
+      'Close',
+      {
+        label: 'Open in Finder',
```
sindresorhus, line 162:
```suggestion
        label: 'Show in Finder',
```

main/recording-history.js
```diff
@@ -0,0 +1,304 @@
...
+    cancelId: 0,
+    defaultId: 2,
+    buttons: [
+      'Close',
+      {
+        label: 'Copy Error',
+        action: () => clipboard.writeText(error)
+      },
+      {
+        label: 'Open in Finder',
```
sindresorhus, line 211:
```suggestion
        label: 'Show in Finder',
```

main/recording-history.js
```diff
@@ -0,0 +1,304 @@
...
+  const options = {
+    title: 'Kap didn\'t shut down correctly',
+    detail: `Looks like Kap crashed during a recording. We were able to locate the file. Unfortunately it appears to be corrupt.\n\n${error}`,
+    cancelId: 0,
+    defaultId: 2,
+    buttons: [
+      'Close',
+      {
+        label: 'Copy Error',
+        action: () => clipboard.writeText(error)
```
sindresorhus, line 208: It's IMHO more readable to not use inline arrow functions when the function body doesn't return anything.

```suggestion
        action: () => {
        	clipboard.writeText(error);
        }
```

sindresorhus: I assume `error` is the error message, not the error object, so I would name it clearer: `errorMessage`.

main/recording-history.js
```diff
@@ -0,0 +1,304 @@
...
+              date: new Date().toISOString()
+            });
+
+            return updateUi({
+              message: 'The recording was repaired successfully',
+              defaultId: 2,
+              buttons: [
+                'Close',
+                {
+                  label: 'Open in Finder',
```
sindresorhus, line 243:
```suggestion
                  label: 'Show in Finder',
```

main/recording-history.js
```diff
@@ -0,0 +1,304 @@
...
+                  action: () => shell.showItemInFolder(outputPath)
+                },
+                {
+                  label: 'Open in Editor',
```
sindresorhus, line 247:
```suggestion
                  label: 'Show in Editor',
```

test/mocks/dialog.js
```diff
@@ -0,0 +1,60 @@
...
+      resolve(index);
+    }
+  } else {
+    resolve(index);
+  }
+};
+
+const getCurrentState = () => {
+  return dialogState;
+};
```
sindresorhus, line 48:
```suggestion
const getCurrentState = () => dialogState;
```

test/mocks/electron-store.js
```diff
@@ -0,0 +1,25 @@
+const sinon = require('sinon');
+
+const mocks = {};
+
+const getMock = sinon.fake((key, defaultValue) => {
+  return mocks[key] || defaultValue;
+});
```
sindresorhus, line 7:
```suggestion
const getMock = sinon.fake((key, defaultValue) => mocks[key] || defaultValue);
```

test/recording-history.js
```diff
@@ -0,0 +1,298 @@
...
+  const missingPath = tempy.file({extension: 'mp4'});
+
+  fs.writeFileSync(existingPath, 'data');
+  t.context.paths = [existingPath];
+
+  recordingHistory.set('recordings', [{filePath: existingPath}, {filePath: missingPath}]);
+  // RecordingHistory.mockGet('recordings', [{filePath: existingPath}, {filePath: missingPath}]);
+
+  t.deepEqual(getPastRecordings(), [{filePath: existingPath}]);
+  // T.true(recordingHistory.set.calledOnceWith('recordings', [{filePath: existingPath}]));
```
sindresorhus, line 68: In case you didn't intend to leave this here.

test/recording-history.js
```diff
@@ -0,0 +1,298 @@
...
+
+  plugins.getRecordingPlugins = sinon.fake.returns([fakePlugin]);
+
+  recordingHistory.set('activeRecording', {
+    filePath: incomplete,
+    name: 'Incomplete',
+    date: new Date().toISOString(),
+    apertureOptions: {},
+    plugins: {
+      'kap-fake-plugin': {'Fake Service': {some: 'state'}}
```
sindresorhus, line 95:
```suggestion
      'kap-fake-plugin': {
      	'Fake Service': {
      		some: 'state'
      	}
      }
```

sindresorhus:
Wow. This is very comprehensive 🙌🏻

main/common/aperture.js
karaggeorge: Any alternatives? I was just trying to find a package like that that allows async code execution

test/recording-history.js
karaggeorge: Oops, yeah that is old. I started by mocking electron-store, but I figured why not just use it in the testing

main/common/aperture.js
sindresorhus: Not that I'm aware.

karaggeorge: We could ask that `cleanUp` be a sync method, and call that instead when we get an exit signal and then we can use https://github.com/sindresorhus/exit-hook?

karaggeorge: Or we can just ignore plugins, and we can set the only action when we get a signal be stopping aperture (which is async, so not sure how good that'll go) and then let the whole recover recording logic handle cleanup on next start

I have noticed that when you kill it, Aperture closes the file anyway, and it doesn't really get corrupt. The only time it really messes up is if you force kill or the computer crashes.

sindresorhus: > and then let the whole recover recording logic handle cleanup on next start

👍🏻

karaggeorge: So, since Aperture handles closing the recording file, and we're not calling plugins, should I remove this entire exit hook logic? The recover logic at next startup I think covers everything, so I don't think there's need to listen for a kill signal

main/recording-history.js
```diff
@@ -0,0 +1,314 @@
...
+    } catch {
+
+    }
+  }
+}];
+
+const handleCorruptRecording = async (recording, error) => {
+  const options = {
+    title: 'Kap didn\'t shut down correctly',
+    detail: `Looks like Kap crashed during a recording. We were able to locate the file. Unfortunately it appears to be corrupt.\n\n${error}`,
```
sindresorhus, line 201:
```suggestion
    detail: `Looks like Kap crashed during a recording. We were able to locate the file. Unfortunately, it appears to be corrupt.\n\n${error}`,
```

main/recording-history.js
```diff
@@ -0,0 +1,314 @@
...
+        action: () => {
+          shell.showItemInFolder(recording.filePath);
+        }
+      }
+    ]
+  };
+
+  const applicableErrors = knownErrors.filter(({test}) => test(error));
+
+  if (applicableErrors.length > 0) {
```
sindresorhus, line 223: Early-return here instead?

main/recording-history.js
```diff
@@ -0,0 +1,314 @@
...
+      return outputPath;
+    } catch {
+
+    }
+  }
+}];
+
+const handleCorruptRecording = async (recording, error) => {
+  const options = {
+    title: 'Kap didn\'t shut down correctly',
```
sindresorhus, line 200: `title` is not shown on macOS. Use `message` instead. And end it with a dot.

Applies in multiple places.

main/recording-history.js
```diff
@@ -0,0 +1,314 @@
...
+
+          if (outputPath) {
+            addRecording({
+              filePath: outputPath,
+              name: recording.name,
+              date: new Date().toISOString()
+            });
+
+            return updateUi({
+              message: 'The recording was repaired successfully',
```
sindresorhus, line 242:
```suggestion
              message: 'The recording was successfully repaired.',
```

main/recording-history.js
```diff
@@ -0,0 +1,314 @@
...
+                  label: 'Show in Editor',
+                  action: () => openEditorWindow(outputPath, {recordingName: recording.name})
+                }
+              ]
+            });
+          }
+        }
+
+        return updateUi({
+          message: 'We were unable to repair the recording',
```
sindresorhus, line 262:
```suggestion
          message: 'Kap was unable to repair the recording.',
```

main/recording-history.js
```diff
@@ -0,0 +1,314 @@
...
+    if (Sentry.isSentryEnabled) {
+      // Collect info about possible unknown errors, to see if we can implement fixes using ffmpeg
+      Sentry.captureException(new Error(`Corrupt recording: ${error}`));
+    }
+  }
+
+  return showDialog(options);
+};
+
+const checkForActiveRecording = async () => {
```
sindresorhus, line 293: Can you add a comment on what it returns? Because the prefix is `check` and not `has`, it's not clear what it would return.

test/recording-history.js
```diff
@@ -0,0 +1,300 @@
...
+    {recordingName: 'Incomplete'}
+  ]);
+
+  t.true(await checkPromise);
+
+  t.false(recordingHistory.has('activeRecording'));
+  t.true(fakeService.cleanUp.calledOnceWith({some: 'state'}));
+  t.deepEqual(recordingHistory.get('recordings'), [
+    {filePath: incomplete, name: 'Incomplete', date: new Date().toISOString()}
+  ]);
```
sindresorhus, line 121: While this is more verbose, it's also IMHO more readable, as each assertion value are on separate lines.

```suggestion
  t.deepEqual(
    recordingHistory.get('recordings'),
    [
      {
        filePath: incomplete, name: 'Incomplete', date: new Date().toISOString()
      }
    ]
  );
```

test/recording-history.js
```diff
@@ -0,0 +1,300 @@
...
+
+  t.true(shell.showItemInFolder.calledWithExactly(corrupt));
+
+  t.true(await checkPromise);
+
+  t.false(recordingHistory.has('activeRecording'));
+  t.is(recordingHistory.get('recordings').length, 0);
+});
+
+test('checkForActiveRecording with unknown corrupt recording', async t => {
```
sindresorhus, line 158:
```suggestion
test('`checkForActiveRecording()` with unknown corrupt recording', async t => {
```

Just to make it clear it's a function. Applies in multiple places.

main/menus.js
```diff
@@ -201,6 +201,17 @@ const appMenuTemplate = [
       {
         type: 'separator'
       },
+      {
+        label: 'Clean temporary files',
```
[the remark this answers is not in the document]
karaggeorge: From discussion on #762 , sounds like some people want to clean up old recordings before the system clean-up runs

main/recording-history.js
karaggeorge: Returns true if there was a recording, and false otherwise. I'll rename to `has`. The return type is only used on start-up to control whether we show the cropper or not

main/menus.js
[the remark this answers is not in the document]
sindresorhus: Can't we just remove the files when Kap exits normally?

main/common/aperture.js
sindresorhus: Yes, I think that should be fine.

main/menus.js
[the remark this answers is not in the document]
karaggeorge: Hmm, I guess we could. I was thinking about this as if you don't have an issue with space, we could show these recordings on the exports window. Not at the moment, but when the editor window is reworked to also show the export, the exports window could become history of sorts, and it would persist across relaunch as long as the temp files were still there.

That would be the same as now (we don't clean the temp files) but we would just keep track of them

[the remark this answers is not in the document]
sindresorhus: For what purpose though? If the user is done with it, they'll want it deleted. If they're not, they'll rather want it in a safe place that will not be cleaned up at a random time.

main/recording-history.js
karaggeorge: This is our custom dialog, so `title` is the title, `detail` is the body, and `message` is an optional extra message towards the bottom.

I'll add the dot

main/menus.js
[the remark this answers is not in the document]
karaggeorge: Updated

main/common/aperture.js
karaggeorge: Updated

sindresorhus:
Can you fix the conflict?

karaggeorge:
Resolved 👍

frozenpandaman:
So awesome to see, thank you!

PR #900, opened 2020-07-21
Fix time label when time is more than an hour

### LINKED ISSUE #893: Video time displayed incorrectly after 60 minutes
**macOS version:** 10.14.6
**Kap version:** 3.2.2, installed via Homebrew cask

Videos recorded that are over an hour in length roll around to zero instead of counting up from 60 (in minutes), or displaying with an hours digit, e.g. this recorded video shows it's 4:17.83 in length (i.e. four minutes), but it should show 64:17.83 or 1:04:17.83. The time displayed on hover while seeking via the time bar is correct, however.

image

karaggeorge:
Fixes #893

### DIFF
--- a/renderer/utils/format-time.js
+++ b/renderer/utils/format-time.js
@@ -2,7 +2,7 @@ import moment from 'moment';
 
 const formatTime = (time, options) => {
   options = {
-    showMilliseconds: true,
+    showMilliseconds: false,
     ...options
   };
 
@@ -14,7 +14,7 @@ const formatTime = (time, options) => {
 };
 
 const format = (time, {showMilliseconds} = {}) => {
-  const formatString = `${time >= 60 * 60 ? 'h:m' : ''}m:ss${showMilliseconds ? '.SS' : ''}`;
+  const formatString = `${time >= 60 * 60 ? 'hh:m' : ''}m:ss${showMilliseconds ? '.SS' : ''}`;
 
   return moment().startOf('day').millisecond(time * 1000).format(formatString);
 };
--- a/renderer/utils/format-time.js
+++ b/renderer/utils/format-time.js
@@ -6,19 +6,17 @@ const formatTime = (time, options) => {
     ...options
   };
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Why are we showing milliseconds there?

renderer/utils/format-time.js
```diff
@@ -6,19 +6,17 @@ const formatTime = (time, options) => {
...
 
-  return `${moment()
-    .startOf('day')
-    .milliseconds(time * 1000)
-    .format(format)}${durationFormatted}`;
+  return `${format(time, options)}${durationFormatted}`;
+};
+
+const format = (time, {showMilliseconds} = {}) => {
+  const formatString = `${time >= 60 * 60 ? 'h:m' : ''}m:ss${showMilliseconds ? '.SS' : ''}`;
```
sindresorhus, line 17: Not related to this PR, but I think we should show it as
```suggestion
  const formatString = `${time >= 60 * 60 ? 'hh:mm' : ''}mm:ss${showMilliseconds ? '.SS' : ''}`;
```

This is how most video players show duration.

karaggeorge: I think you meant this:

```suggestion
  const formatString = `${time >= 60 * 60 ? 'h:' : ''}mm:ss${showMilliseconds ? '.SS' : ''}`;
```

right? Otherwise if it's more than an hour the format becomes `hh:mmmm:ss`

karaggeorge:
I'm not sure about the milliseconds. I've never found it useful, but it was like that since Kap 2.x, so I just kept the logic during the rewrite

renderer/utils/format-time.js
sindresorhus: I meant:

```suggestion
  const formatString = `${time >= 60 * 60 ? 'hh:m' : ''}m:ss${showMilliseconds ? '.SS' : ''}`;
```

sindresorhus:
Yeah, let's try removing it and see if anyone complains.

karaggeorge:
Removed milliseconds and changed the format 👍

PR #902, opened 2020-07-22
Add AV1 support

### LINKED ISSUE #221: AV1 support
[Issuehunt badges](https://issuehunt.io/r/wulkano/kap/issues/221)

Is support for the [AV1 codec](https://en.wikipedia.org/wiki/AOMedia_Video_1) (in WebM container) on the roadmap? Is it even possible since Kap uses [Aperture](https://github.com/wulkano/aperture) (and thus AVFoundation)?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $100.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($100.00)

### Submitted pull Requests
- [#902 Add AV1 support](https://issuehunt.io/r/wulkano/kap/pull/902)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/wulkano/kap/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

karaggeorge:
Closes #221 

Also added tests for the recent lossy gif addition to the convert file

Was able to get a pretty good set of options. Conversions take a little longer than the rest of the formats but not unreasonable if you reduce the size/fps from the original. But, even so, if someone uses AV1 it'd be because they understand it takes longer to encode, but it's better for multiple subsequent plays

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#221: AV1 support](https://issuehunt.io/repos/65411043/issues/221)
---
</details>

### DIFF
--- a/main/common/constants.js
+++ b/main/common/constants.js
@@ -6,9 +6,7 @@ const formatExtensions = new Map([
   ['av1', 'mp4']
 ]);
 
-const getFormatExtension = format => {
-  return formatExtensions.get(format) || format;
-};
+const getFormatExtension = format => formatExtensions.get(format) || format;
 
 module.exports = {
   supportedVideoExtensions,
--- a/main/common/constants.js
+++ b/main/common/constants.js
@@ -2,7 +2,16 @@
 
 const supportedVideoExtensions = ['mp4', 'mov', 'm4v'];
 
+const formatExtensions = new Map([
+  ['av1', 'mp4']
+]);
+
+const getFormatExtension = format => {
+  return formatExtensions.get(format) || format;
+};
+
 module.exports = {
   supportedVideoExtensions,
+  getFormatExtension,
   defaultInputDeviceId: 'SYSTEM_DEFAULT'
 };
--- a/main/convert.js
+++ b/main/convert.js
@@ -23,6 +23,8 @@ const speedRegex = /speed=\s*(-?\d+(,\d+)*(\.\d+(e\d+)?)?)/gm;
 // https://trac.ffmpeg.org/ticket/309
 const makeEven = n => 2 * Math.round(n / 2);
 
+const areDimensionsEvent = ({width, height}) => width % 2 === 0 && height % 2 === 0;
+
 const getRunFunction = (shouldTrack, mode = 'convert') => (outputPath, options, args) => {
   const modes = new Map([
     ['convert', ffmpegPath],
@@ -65,6 +67,7 @@ const getRunFunction = (shouldTrack, mode = 'convert') => (outputPath, options, 
         // Wait 2 second in the conversion for the speed to be stable
         if (processedMs > 2 * 1000) {
           const msRemaining = (durationMs - processedMs) / speed;
+
           options.onProgress(progress, prettyMs(Math.max(msRemaining, 1000), {compact: true}));
         } else {
           options.onProgress(progress);
@@ -141,7 +144,7 @@ const convertToMp4 = PCancelable.fn(async (options, onCancel) => {
     '-i', options.inputPath,
     '-r', options.fps,
     ...(
-      options.shouldCrop ? [
+      options.shouldCrop || !areDimensionsEvent(options) ? [
         '-s', `${makeEven(options.width)}x${makeEven(options.height)}`,
         '-ss', options.startTime,
         '-to', options.endTime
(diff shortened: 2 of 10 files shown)


main/convert.js
```diff
@@ -185,6 +188,39 @@ const convertToWebm = PCancelable.fn(async (options, onCancel) => {
...
+      ] : []
+    ),
+    '-c:v', 'libaom-av1',
+    '-c:a', 'libopus',
+    '-crf', '34',
+    '-b:v', '0',
+    '-strict', 'experimental',
+    '-cpu-used', '4',
+    '-row-mt', '1',
+    '-tiles', '2x2',
```
karaggeorge, line 219: These are not necessary but they do help a bit with how long it takes for me. If for some reason they don't work, we can remove later.

main/plugins/open-with-plugin.js
```diff
@@ -1,17 +1,18 @@
...
   const filePath = await context.filePath();
   openFileWithApp(filePath, context.appUrl);
 };
 
 const apps = new Map(
-  ['mp4', 'gif', 'apng', 'webm']
-    .map(extension => [
-      extension,
-      getAppsThatOpenExtension.sync(extension)
+  ['mp4', 'gif', 'apng', 'webm', 'av1']
```
karaggeorge, line 12: This pretty much adds the same apps as `mp4` since they both have an `.mp4` extension (and that's how we determine which apps open what file).

Some apps don't support it, but there's no way to know. Gifski has a pretty good error message, VLC supports it, and even Kap does (it opens in the preview just fine)

test/convert.js
```diff
@@ -293,3 +295,107 @@ test('gif: non-retina', async t => {
...
+    height: 286,
+    startTime: 10,
+    endTime: 15,
+    shouldCrop: true
+  });
+
+  t.true(
+    fs.statSync(regular).size >
+    fs.statSync(lossy).size
+  );
```
karaggeorge, line 325: Not sure how else to test the lossy feature, other than comparing the file sizes

main/convert.js
sindresorhus: Can you add a short code comment about what it does?

main/common/constants.js
```diff
@@ -2,7 +2,16 @@
 
 const supportedVideoExtensions = ['mp4', 'mov', 'm4v'];
 
+const formatExtensions = new Map([
+  ['av1', 'mp4']
+]);
+
+const getFormatExtension = format => {
+  return formatExtensions.get(format) || format;
+};
```
sindresorhus, line 11:
```suggestion
const getFormatExtension = format => (formatExtensions.get(format) || format);
```

main/convert.js
```diff
@@ -23,6 +23,8 @@ const speedRegex = /speed=\s*(-?\d+(,\d+)*(\.\d+(e\d+)?)?)/gm;
 // https://trac.ffmpeg.org/ticket/309
 const makeEven = n => 2 * Math.round(n / 2);
 
+const areDimensionsEvent = ({width, height}) => width % 2 === 0 && height % 2 === 0;
```
sindresorhus, line 26:
```suggestion
const areDimensionsEven = ({width, height}) => width % 2 === 0 && height % 2 === 0;
```

?

main/export-options.js
```diff
@@ -50,7 +56,8 @@ const prettifyFormat = format => {
     ['apng', 'APNG'],
     ['gif', 'GIF'],
     ['mp4', 'MP4'],
-    ['webm', 'WebM']
+    ['webm', 'WebM'],
+    ['av1', 'AV1']
   ]);
```
sindresorhus, line 61:
```js
const formats = new Map([
  ['apng', 'APNG'],
  ['gif', 'GIF'],
  ['mp4', 'MP4 (H264/HEVC)'],
  ['av1', 'MP4 (AV1)'],
  ['webm', 'WebM']
]);
```

(changed order and titles)

?

test/convert.js
sindresorhus: That's the only way.

test/mocks/settings.js
```diff
@@ -1,5 +1,17 @@
 'use strict';
+const sinon = require('sinon');
+
+const mocks = {};
+
+const mockGet = sinon.fake((key, defaultValue) => {
+  return mocks[key] || defaultValue;
+});
```
sindresorhus, line 8:
```suggestion
const mockGet = sinon.fake((key, defaultValue) => mocks[key] || defaultValue);
```

main/export-options.js
karaggeorge: Wouldn't hevc and h264 be separate options? (when we add hevc)

sindresorhus: Yes, but it's not added yet.

sindresorhus:
The dropdown doesn't look very good when closed:

Screen Shot 2020-07-23 at 20 19 40

Maybe we should make it wider?

PR #904, opened 2020-07-23
Rethrowing an original error if the error message is not matched to the video text regEx

OleksandrPodoliako:
Added rethrowing an error if  there's no match.

### DIFF
--- a/main/utils/encoding.js
+++ b/main/utils/encoding.js
@@ -14,7 +14,11 @@ const getEncoding = async filePath => {
   try {
     await execa(ffmpegPath, ['-i', filePath]);
   } catch (error) {
-    return /.*: Video: (.*?) \(.*/.exec(error.stderr)[1];
+    if (/.*: Video: (.*?) \(.*/.exec(error.stderr)[1]) {
+      throw error;
+    } else {
+      return /.*: Video: (.*?) \(.*/.exec(error.stderr)[1];
+    }
   }
 };
 
--- a/main/utils/encoding.js
+++ b/main/utils/encoding.js
@@ -6,7 +6,7 @@ const ffmpeg = require('@ffmpeg-installer/ffmpeg');
 const util = require('electron-util');
 const execa = require('execa');
 
-const { track } = require('../common/analytics');
+const {track} = require('../common/analytics');
 
 const ffmpegPath = util.fixPathForAsarUnpack(ffmpeg.path);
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for contributing. The PR needs a better title and I think you could [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) up the code a bit.

OleksandrPodoliako:
@sindresorhus, no problem. I've made changes.

main/utils/encoding.js
```diff
@@ -14,7 +14,12 @@ const getEncoding = async filePath => {
   try {
     await execa(ffmpegPath, ['-i', filePath]);
   } catch (error) {
-    return /.*: Video: (.*?) \(.*/.exec(error.stderr)[1];
+    const errorMessageMatchedText = /.*: Video: (.*?) \(.*/.exec(error.stderr);
+    if (errorMessageMatchedText) {
+      throw error;
+    } else {
```
sindresorhus, line 20: The `else` here is not needed.

main/utils/encoding.js
```diff
@@ -14,7 +14,12 @@ const getEncoding = async filePath => {
   try {
     await execa(ffmpegPath, ['-i', filePath]);
   } catch (error) {
-    return /.*: Video: (.*?) \(.*/.exec(error.stderr)[1];
+    const errorMessageMatchedText = /.*: Video: (.*?) \(.*/.exec(error.stderr);
+    if (errorMessageMatchedText) {
+      throw error;
+    }
+
+  return errorMessageMatchedText[1];
```
sindresorhus, line 22: Indentation

main/utils/encoding.js
```diff
@@ -14,7 +14,12 @@ const getEncoding = async filePath => {
   try {
     await execa(ffmpegPath, ['-i', filePath]);
   } catch (error) {
-    return /.*: Video: (.*?) \(.*/.exec(error.stderr)[1];
+    const errorMessageMatchedText = /.*: Video: (.*?) \(.*/.exec(error.stderr);
+    if (errorMessageMatchedText) {
```
sindresorhus, line 18: Shouldn't this be negated?

main/utils/encoding.js
OleksandrPodoliako: I was wondering if you can give me a hint how to set up code formatting  automatically in VSC?

main/utils/encoding.js
OleksandrPodoliako: Correct

main/utils/encoding.js
```diff
@@ -6,21 +6,26 @@ const ffmpeg = require('@ffmpeg-installer/ffmpeg');
 const util = require('electron-util');
 const execa = require('execa');
 
-const {track} = require('../common/analytics');
+const { track } = require('../common/analytics');
```
sindresorhus, line 9: Don't do unrelated changes.

sindresorhus:
Does this fix an issue? If so, it would be great if you could link to it. Also, make sure your changes actually fixes the issue by running the PR and manually confirming.

main/utils/encoding.js
OleksandrPodoliako: I got it

OleksandrPodoliako:
I've checked, now it works properly.

PR #915, opened 2020-08-19
Actually return devices

### LINKED ISSUE #913: Input devices crash on 10.13
**Sentry Issue:** [KAP-7C9](https://sentry.io/organizations/wulkano-l0/issues/1769949136/)

**macOS version:**    High Sierra (10.13)
**Kap version:**      3.3.0

```
Unhandled Promise Rejection

TypeError: Cannot read property 'some' of undefined
    at /Applications/Kap.app/Contents/Resources/app.asar/main/common/settings.js:126:18
```

### LINKED ISSUE #914: Cannot read property 'some' of undefined
**macOS version:** 10.13.6
**Kap version:** v3.3.0

#### Steps to reproduce
Was a fresh install on a mac running OSX 10.13.6
Once OSX security allowed the app to run as an unsigned developer, start-up would fail with the following dialogue. 
Screen Shot 2020-08-19 at 20 33 21

#### Current behaviour
Kap will not start, error dialogue box is broken and can not be interacted with (e.g. the buttons can not be pressed)

#### Expected behaviour
Kap to start up

#### Workaround
No workaround found. 

Couldn't really spot how `getAudioDevices()` isn't returning an array. 
https://github.com/wulkano/Kap/blob/35f18894b3c77fe710acf4507d45e920498877cf/main/utils/devices.js#L9

Installed https://github.com/karaggeorge/macos-audio-devices locally and got the following;
```
apple$ audio-devices
dyld: Library not loaded: @rpath/libswiftCore.dylib
  Referenced from: /usr/local/bin/audio-devices
  Reason: image not found
Abort trap: 6
```

karaggeorge:
Fixes #913 
Fixes #914 

When we use aperture.devices fallback, I forgot to actually return the devices array

Also fixes the issue for when the user is on 10.13, and doesn't have swift libs installed, where we can't get the system default device (aperture doesn't return that). So it hides the name from the menus, and if it's selected at the time of recording, we just pick the first input in the list as the "default".

### DIFF
--- a/main/utils/devices.js
+++ b/main/utils/devices.js
@@ -30,11 +30,18 @@ const getAudioDevices = async () => {
       return 0;
     }).map(device => ({id: device.uid, name: device.name}));
   } catch (error) {
-    const devices = await aperture.audioDevices();
+    try {
+      const devices = await aperture.audioDevices();
 
-    if (!Array.isArray(devices)) {
-      const Sentry = require('./sentry');
-      Sentry.captureException(new Error(`devices is not an array: ${JSON.stringify(devices)}`));
+      if (!Array.isArray(devices)) {
+        const Sentry = require('./sentry');
+        Sentry.captureException(new Error(`devices is not an array: ${JSON.stringify(devices)}`));
+        showError(error);
+        return [];
+      }
+
+      return devices;
+    } catch (error) {
       showError(error);
       return [];
     }
--- a/main/utils/devices.js
+++ b/main/utils/devices.js
@@ -49,11 +49,15 @@ const getAudioDevices = async () => {
 };
 
 const getDefaultInputDevice = () => {
-  const device = audioDevices.getDefaultInputDevice.sync();
-  return {
-    id: device.uid,
-    name: device.name
-  };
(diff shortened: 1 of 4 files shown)


main/utils/devices.js
```diff
@@ -30,23 +30,34 @@ const getAudioDevices = async () => {
...
+        const Sentry = require('./sentry');
+        Sentry.captureException(new Error(`devices is not an array: ${JSON.stringify(devices)}`));
+        showError(error);
+        return [];
+      }
 
-    if (!Array.isArray(devices)) {
-      const Sentry = require('./sentry');
-      Sentry.captureException(new Error(`devices is not an array: ${JSON.stringify(devices)}`));
+      return devices;
```
karaggeorge, line 43: This is what caused the crash, the rest is just additional checks and fallbacks for devices

sindresorhus:
I cannot test the actual change as I don't have macOS 10.13, but the code looks ok.

karaggeorge:
I threw errors in the code where the 10.13 issue would be to test it, and it worked 👍

PR #929, opened 2020-09-09
Fix infinite error loop

### LINKED ISSUE #926: Swarm of error windows overnight
I've left my macbook overnight and came to this:

**Sentry Issue:** [KAP-7P0](https://sentry.io/organizations/wulkano-l0/issues/1850041104/)

**macOS version:**    Catalina (10.15)
**Kap version:**      3.3.1

```
RequestError

RequestError: getaddrinfo ENOTFOUND kap-sentry-tracker.vercel.app
    at ClientRequest.<anonymous> (/Applications/Kap.app/Contents/Resources/app.asar/node_modules/got/source/request-as-event-emitter.js:178:14)
    at ClientRequest.origin.emit (/Applications/Kap.app/Contents/Resources/app.asar/node_modules/@szmarczak/http-timer/source/index.js:37:11)
```

I think my wifi broke over the night or something, which would explain the `RequestError` maybe?
Those window alerts are from Kap (I should have taken a screenshot of one up close, sorry...).

image

karaggeorge:
Fixes #926 

When there's a problem with the service that tracks sentry issues, the user can just see the message we add to the dialog `Something went wrong while collecting the information.`

Right now we would try to show an error dialog for it, which would call the service again, creating an endless cycle of errors

### DIFF
--- a/main/utils/errors.js
+++ b/main/utils/errors.js
@@ -33,6 +33,7 @@ const getSentryIssue = async (eventId, tries = 0) => {
 
     return body;
   } catch (error) {
+    // We are not using `showError` again here to avoid an infinite error cycle
     console.log(error);
   }
 };
--- a/main/utils/errors.js
+++ b/main/utils/errors.js
@@ -33,7 +33,7 @@ const getSentryIssue = async (eventId, tries = 0) => {
 
     return body;
   } catch (error) {
-    showError(error);
+    console.log(error);
   }
 };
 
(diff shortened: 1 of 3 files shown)


main/utils/errors.js
```diff
@@ -33,7 +33,7 @@ const getSentryIssue = async (eventId, tries = 0) => {
 
     return body;
   } catch (error) {
-    showError(error);
```
sindresorhus, line 36: Add a code comment about this, so we don't accidentally change it back in the future.

PR #930, opened 2020-09-09
Add ability to ignore certain errors

karaggeorge:
Fixes #927 

Added ability to ignore certain errors based on regex. Also added regex for the one in the issue above

PR #932, opened 2020-09-09
Ignore auto-updater errors and some additional based on regex

karaggeorge:
Addresses the comments here:
- https://github.com/wulkano/Kap/pull/930#issuecomment-689664621
- https://github.com/wulkano/Kap/pull/930#issuecomment-689664836

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -71,7 +71,7 @@ const checkForUpdates = () => {
 
   setInterval(checkForUpdates, toMilliseconds({hours: 1}));
 
-  checkForUpdates()
+  checkForUpdates();
 };
 
 // Prepare the renderer once the app is ready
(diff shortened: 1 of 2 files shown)


main/index.js
```diff
@@ -59,15 +59,19 @@ const checkForUpdates = () => {
     return false;
   }
 
+  const checkForUpdates = async () => {
+    try {
+      await autoUpdater.checkForUpdates();
+    } catch {}
```
albinekb, line 65: Should there be a log here or something to catch it in developer or `Console.app`, not sure?

albinekb (approved):

main/index.js
karaggeorge: Well the updater doesn't run in developer mode. Do you think there's value in logging it to Console.app? Usually these are the errors like connection timed out etc

albinekb: I think this is fine, I test the updater every version manually anyways 👍

PR #989, opened 2021-03-13
Upgrade dependencies

sindresorhus:
Mostly just linting fixes and fixing some deprecation notices.


renderer/hooks/editor/use-conversion-id.tsx
```diff
@@ -23,6 +23,8 @@ export const ConversionIdContextProvider = (props: PropsWithChildren<Record<stri
     setConversionId(id || savedConversionId);
   };
 
+  // TODO: Fix this.
+  // eslint-disable-next-line react/jsx-no-constructed-context-values
```
karaggeorge, line 27: This actually might be the reason for some random re-renders I was seeing. I'll push a commit to fix it with a useCallback

karaggeorge (approved):
Looks good. Can you check my last commit, and then if that's all good we can merge

Issue #976, opened 2021-02-04
[Feature Request] Add an undo / redo stack for cropper.


steveruizok:
### Problem

One of the vanishingly few negative user experiences I've had with Kap is when I'm recording multiple captures of the same part of my screen, but then accidentally reside the cropped area between recordings. It's often impossible to manually restore the same cropped bounds.

### Solution

To fix this, we could **implement an undo / redo stack for the cropper**, allowing a user to restore previous cropper settings. 

### Implementation

This feature could be implemented in the cropper container, following a standard undo/redo implementation pattern.

- The container's state would keep an array of settings (the "stack") and a number (the "stack pointer") indicating the index of the current settings. 
- Each `updateSettings` call would update the stack by slicing it up to the pointer, pushing the new settings, and then updating the pointer. 
- Standard keyboard shortcuts would be used to move the pointer up and down, with each change restoring the settings found in the stack at the pointer's index.

This change would also replace the logic for restoring the previous (or `original`) cropper bounds after leaving full screen. We could move `isFullscreen` into the settings definition (stored in the stack), allowing a user to undo or redo back into or out of full screen mode.

PR #1004, opened 2021-04-23
Added undo / redo stack for cropper

lescobar723:
I added a stack to keep track of each cropper state whenever the cropper is adjusted or edited. The user can now press the 'z' key for undo or 'y' key for redo and the cropper will be adjusted accordingly. 

This added feature is in response to #976

### DIFF
--- a/renderer/containers/cropper.js
+++ b/renderer/containers/cropper.js
@@ -67,7 +67,7 @@ export default class CropperContainer extends Container {
       // Added this to keep track of cropper history to add undo button
       cropperHistory: [],
       cropperPointer: -1,
-      undone: false,
+      isUndone: false,
       recordAudio: this.settings.get('recordAudio'),
       audioInputDeviceId: this.settings.getSelectedInputDeviceId()
     };
@@ -127,7 +127,7 @@ export default class CropperContainer extends Container {
 
   updateSettings = updates => {
     // If already undone, if moved again then erase everything after current pointer
-    if (this.state.undone) {
+    if (this.state.isUndone) {
       let newCropperHistory = this.state.cropperHistory;
       const CropperHistoryLength = this.state.cropperHistory.length;
       if (this.state.cropperPointer > -1) {
@@ -137,8 +137,10 @@ export default class CropperContainer extends Container {
         newCropperHistory = [];
       }
 
-      this.setState({cropperHistory: newCropperHistory});
-      this.setState({undone: false});
+      this.setState({
+        cropperHistory: newCropperHistory,
+        isUndone: false
+      });
     }
 
     const {x, y, width, height, ratio, displayId} = this.state;
@@ -152,8 +154,10 @@ export default class CropperContainer extends Container {
       displayId
     });
     // Added in: keep track of cropper
-    this.setState({cropperHistory: [...this.state.cropperHistory, this.state]});
-    this.setState({cropperPointer: this.state.cropperPointer + 1});
+    this.setState({
+      cropperHistory: [...this.state.cropperHistory, this.state],
+      cropperPointer: this.state.cropperPointer + 1
+    });
 
     this.setState(updates);
   };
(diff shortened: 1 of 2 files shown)


sindresorhus:
It should use Shift+Command+Z to redo. That's the convention on macOS.

renderer/containers/cropper.js
```diff
@@ -334,6 +356,50 @@ export default class CropperContainer extends Container {
...
+
+  undo = () => {
+    if (this.state.cropperPointer > 0) {
+      const originalCropperHistory = this.state.cropperHistory;
+      const currPointer = this.state.cropperPointer - 1;
+      const newState = this.state.cropperHistory[currPointer];
+      this.setState(newState);
+      this.setState({undone: true});
+      this.setState({cropperHistory: originalCropperHistory});
+      this.setState({cropperPointer: currPointer});
```
sindresorhus, line 389: I think it would be better to use a single setState call.

renderer/containers/cropper.js
```diff
@@ -334,6 +356,50 @@ export default class CropperContainer extends Container {
...
+        width,
+        height,
+        ratio,
+        displayId
+      });
+    }
+  }
+
+  undo = () => {
+    if (this.state.cropperPointer > 0) {
```
sindresorhus, line 382: Use early-return.

renderer/containers/cropper.js
```diff
@@ -334,6 +356,50 @@ export default class CropperContainer extends Container {
     }
   };
 
+  redo = () => {
+    if (this.state.undone && this.state.cropperPointer < this.state.cropperHistory.length - 1) {
+      const originalCropperHistory = this.state.cropperHistory;
+      const currPointer = this.state.cropperPointer + 1;
```
sindresorhus, line 362: Don't use abbreviations.

renderer/containers/cropper.js
```diff
@@ -64,6 +64,10 @@ export default class CropperContainer extends Container {
       isActive: false,
       isReady: false,
       ratio: [1, 1],
+      // Added this to keep track of cropper history to add undo button
+      cropperHistory: [],
+      cropperPointer: -1,
+      undone: false,
```
sindresorhus, line 70: `isUndone`

sindresorhus (changes requested):

sindresorhus:
Would have been nice if the all the undo-related logic would go into a `UndoManager` class.

skllcrn:
Thank you! Great first commit. I'm curious about you reasoning for going with <kbd>Z<kbd> and <kbd>Y<kbd> over <kbd>Command<kbd> + <kbd>Z<kbd> to undo, <kbd>Command<kbd> + <kbd>Shift<kbd> + <kbd>Z<kbd> to redo?

lescobar723:
Hello, thank you for the response! I will try to make the new changes, but I am new to using Electron and React so it will take me a bit but I am up for the challenge! I used the 'z' and 'y' keys for undo and redo because I noticed in the function handleKeyEvent function in the cropper.js file in the pages directory whenever you press one of the keys that are in either in Command+Z and Command+Z+Shift, it really depends on which of those keys you pressed first. Let's say you pressed Shift first in the Command+Z+Shift combination then in the cropper.js file it will go into the Shift case regardless on if you pressed more keys. You cannot really tell the order in which the user will press keys in so I was thinking maybe for multiple key presses the single key cases in handleKeyEvent is not very suitable. So I will try to find a fix to that, one solution is mentioned [here](https://stackoverflow.com/questions/5203407/how-to-detect-if-multiple-keys-are-pressed-at-once-using-javascript). 

Any other suggestions would be greatly appreciated.

renderer/pages/cropper.js
```diff
@@ -20,6 +20,8 @@ actionBarContainer.bindCursor(cursorContainer);
 actionBarContainer.bindCropper(cropperContainer);
 
 let lastRatioLockState = null;
+let metaKeyDown = false;
```
sindresorhus, line 23:
```suggestion
let iaMetaKeyDown = false;
```

renderer/pages/cropper.js
```diff
@@ -20,6 +20,8 @@ actionBarContainer.bindCursor(cursorContainer);
...
+let ShiftDown = false;
```
sindresorhus, line 24:
```suggestion
let iaShiftKeyDown = false;
```

renderer/pages/cropper.js
```diff
@@ -96,6 +110,14 @@ export default class CropperPage extends React.Component {
       case 'i':
         this.remote.getCurrentWindow().setIgnoreMouseEvents(true);
         this.dev = !this.dev;
+        break;
+      case 'z':
+        if (metaKeyDown && ShiftDown) {
```
sindresorhus, line 115: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/shiftKey

renderer/containers/cropper.js
```diff
@@ -330,7 +356,56 @@ export default class CropperContainer extends Container {
...
+        y,
+        width,
+        height,
+        ratio,
+        displayId
+      });
+    }
+  };
+
+  undo = () => {
```
sindresorhus, line 388: There's a lot of duplicated logic between the `undo` and `redo` methods. I would try to unify the logic into an internal method which both of these call.

renderer/containers/cropper.js
```diff
@@ -122,8 +126,24 @@ export default class CropperContainer extends Container {
   };
 
   updateSettings = updates => {
-    const {x, y, width, height, ratio, displayId} = this.state;
+    // If already undone, if moved again then erase everything after current pointer
+    if (this.state.isUndone) {
+      let newCropperHistory = this.state.cropperHistory;
+      const CropperHistoryLength = this.state.cropperHistory.length;
```
sindresorhus, line 132:
```suggestion
      const cropperHistoryLength = this.state.cropperHistory.length;
```

sindresorhus (changes requested):

sindresorhus:
Have you looked into using something like https://github.com/frontendphil/react-undo-redo or https://github.com/homerchen19/use-undo ?

sindresorhus:
Closing for lack of response.

PR #1016, opened 2021-05-23
TypeScript Rewrite: Exports Window

karaggeorge:
Rewrites main/renderer code for exports window. Also revisited some logic in the Conversion/Export classes created in the Editor rewrite. Fixed some small issues and added a few things:

- Retry button on Editor window for failed/canceled exports
- Retry option in the export menu in the Exports window
- Copy option in the export menu in the Exports window
- Better error/cancelation handling in export process


renderer/components/editor/conversion/conversion-details.tsx
```diff
@@ -1,10 +1,10 @@
 import {UseConversionState} from 'hooks/editor/use-conversion';
 
 const ConversionDetails = ({conversion}: {conversion: UseConversionState}) => {
-  const message = conversion?.message ?? 'Loading…';
+  const message = conversion?.message || 'Loading…';
```
karaggeorge, line 4: This is because `message` is initialized with `''` in the main process.

main/conversion.ts
```diff
@@ -1,83 +1,43 @@
...
 
   private conversionProcess?: PCancelable<string>;
 
-  constructor(video: Video, format: Format, options: ConversionOptions) {
+  constructor(
+    public readonly video: Video,
+    public readonly format: Format,
+    public readonly options: ConversionOptions
+  ) {
+    // eslint-disable-next-line constructor-super
```
karaggeorge, line 39: This is to silence a warning because of the TypedEmitter annotation. Couldn't find a way around it.

Warning said this doesn't need a `super()` call, but obviously without one we get a compilation error

renderer/components/exports/export.tsx
```diff
@@ -0,0 +1,182 @@
...
+    },
+    state?.canCopy && {
+      label: 'Copy',
+      click: () => copy()
+    },
+    canRetry && {
+      label: 'Retry',
+      click: () => retry()
+    }].filter(Boolean));
+  }, [canRetry, retry, openInEditor]);
```
karaggeorge, line 56: Not sure about text/order/dividers for this. It's the context menu in the Exports window. At the moment it can have one of three states:

```
Open Original
```

```
Open Original
Copy
```

```
Open Original
Retry
```

renderer/components/editor/conversion/conversion-details.tsx
sindresorhus: I would add a code comment so we don't try to revert to `??` in the future.

sindresorhus: But wouldn't it be better to not initialize it with `''`?

main/windows/kap-window.ts
```diff
@@ -162,9 +165,9 @@ export default class KapWindow<State = any> {
...
-  private readonly executeIfNotDestroyed = (callback: () => void) => {
-    if (!this.browserWindow.isDestroyed()) {
-      callback();
-    }
-  };
+  // private readonly executeIfNotDestroyed = (callback: () => void) => {
+  //   if (!this.browserWindow.isDestroyed()) {
+  //     callback();
+  //   }
+  // };
```
sindresorhus, line 172: Intention to leave it like this?

sindresorhus:
Probably not related to this PR, but I noticed there's no "Copy" button when converting to MP4, while there is when converting to GIF.

sindresorhus:
I got this when choosing "Open original" in the export history:

Screen Shot 2021-05-23 at 14 46 34

sindresorhus:
> ../node_modules/@babel/runtime/helpers/esm/asyncToGenerator.js (21:0)

Why is Babel transpiling async functions? Electron supports them natively.

sindresorhus:
The "export completed" screen needs a way to show the file in Finder. In the export history, you can click the list item to do so. I also think the "export completed" screen should have a way to "Open original".

The "export completed" screen should indicate what file format I just exported. I didn't remember whether it was GIF or MP4.

The "export completed" screen should show the actual GIF instead of a static thumbnail.

renderer/components/exports/export.tsx
sindresorhus: I think it should be:

```
Open Original
```

```
Copy
------
Open Original
```

```
Retry
------
Open Original
```

renderer/components/editor/conversion/conversion-details.tsx
karaggeorge: Yeah that's fair. Mostly I just didn't want to make it nullable. But I could also initialize it with 'Loading' since that's the value it's set to in the UI when it's empty

main/windows/kap-window.ts
karaggeorge: This was throwing an error that it was unused but I wasn't sure if it'd get used in the future

karaggeorge:
Can we copy MP4? I thought you can only have gif and apng in your clipboard

I'll look into that error and Babel

The conversion preview is actually just the video element that plays in the editor state, but we just resize and pause it. Should I replace it with the actual result of the conversion (whatever format that may be) and have it play?

sindresorhus:
> Can we copy MP4? I thought you can only have gif and apng in your clipboard

Yes. We don't copy the animated GIF directly since the pasteboard doesn't support that. We copy a file URL to it. We can do the same with MP4. https://github.com/wulkano/Kap/blob/eaf8193b76ac6cd908f5da62584f1346c537cc24/main/plugins/built-in/copy-to-clipboard-plugin.ts#L6-L8

sindresorhus:
> Should I replace it with the actual result of the conversion (whatever format that may be) and have it play?

Only if it's GIF, I would say. As we cannot play AV1, for example. It's GIF that is interesting to see a preview of, as you can then determine whether you need to re-export using different settings.

karaggeorge:
@sindresorhus I think I addressed almost everything here. Only question, regarding:

> I also think the "export completed" screen should have a way to "Open original".

In Exports, Open Original is opening the file back in the editor. What do you mean by this in the editor screen? We added showing the file in Editor, when you click the preview (dragging should also work now). Does that cover that request? The location is the temp file for most plugins, and the actual file for `Save to Disk` plugin

sindresorhus:
> In Exports, Open Original is opening the file back in the editor. What do you mean by this in the editor screen? We added showing the file in Editor, when you click the preview (dragging should also work now). Does that cover that request? The location is the temp file for most plugins, and the actual file for Save to Disk plugin. But in either case it's the converted file, not the original

Yeah. You're right. It doesn't make sense there since you can just click the "Back" button to get the original in the editor.

sindresorhus:
The new changes look good.

sindresorhus:
I don't think it's clear enough to the user that they can drag and drop the thumbnail or that they can click it to show in Finder.

Maybe we should have a popover the first time the export completed view is shown with a tip about drag and drop and click the thumbnail?

Maybe we should also make the filename clickable (with underline) and make it "open in finder"?

sindresorhus:
In the export history, I noticed that new entries are added to the end. Shouldn't the most recent export be first?

sindresorhus:
`.apng` doesn't fit in the export completed screen:

Screen Shot 2021-05-24 at 16 51 47

sindresorhus:
There's something weird with the aspect ratio in the preview: (Notice how everything is wider)

Screen Shot 2021-05-24 at 16 52 45

sindresorhus:
While converting, the circular progress is not that visible. I the overlay we show when hovering over should be there permanently	while converting.

Screen Shot 2021-05-24 at 16 58 23

karaggeorge:
> Maybe we should have a popover the first time the export completed view is shown with a tip about drag and drop and click the thumbnail?

Native dialog with some text? Or do we have a better idea for design?

> Maybe we should also make the filename clickable (with underline) and make it "open in finder"?

Is the underline always there, or on hover?

> .apng doesn't fit in the export completed screen:

We discussed this with @skllcrn , it was either ellipsis from the start or adding the full name on hover. Ended up going with the latter. So if you hover over you should see the extension.

One more thing we could do is add it to the description:
`888 x 722 at 15 FPS` -> `888 x 722 GIF at 15 FPS` (using the format pretty name)

sindresorhus:
> Native dialog with some text? Or do we have a better idea for design?

I don't think it should be a blocking dialog. It's just a tip, not important information. A tooltip or toast would be best. It should only be shown once. This is how I do it in Gifski:

Screen Shot 2021-05-25 at 01 01 47

Maybe @skllcrn has opinions about this.

> Is the underline always there, or on hover?

I would have it always there, but ask @skllcrn.

> We discussed this with @skllcrn , it was either ellipsis from the start or adding the full name on hover. Ended up going with the latter. So if you hover over you should see the extension.

I would still make the default name template fit for all extensions. Just make the window a tiny bit wider.

> One more thing we could do is add it to the description:
888 x 722 at 15 FPS -> 888 x 722 GIF at 15 FPS (using the format pretty name)

I think the extension is better.

skllcrn:
Love the drag and drop tooltip! I don't this underlines are a common pattern on macOS, maybe we could add a way to open the directory from the File menu too? 

Regarding seeing the full file name including the extension, making the window slightly larger so that default filenames always fit seems reasonable. That said, a quick search tells me that macOS filenames can be up to 255 characters, so truncation is needed in any case, and hovering to see the full value is a nice affordance.

sindresorhus:
> I don't think underlines are a common pattern on macOS, maybe we could add a way to open the directory from the File menu too?

You're right, it's not a common pattern, but I have seen it done sometimes. I cannot find an example right now.

How about this instead; A hamburger menu next to the "Copy" button which would have a "Show in Finder" menu item? It would let us add other things there in the future too.

> That said, a quick search tells me that macOS filenames can be up to 255 characters, so truncation is needed in any case, and hovering to see the full value is a nice affordance.

Sure, but most people just use the default template.

skllcrn:
> A hamburger menu next to the "Copy" button which would have a "Show in Finder" menu item? It would let us add other things there in the future too.

Sounds good to me!

karaggeorge:
I'll add the menu for those. And in the future, we can add the native share menu on that, and actions from the export plugins. What icon should I use for that?

Also, does that mean we are removing the show in finder when you click on the preview? Or are we doing both

sindresorhus:
> Also, does that mean we are removing the show in finder when you click on the preview? Or are we doing both

Both

karaggeorge:
@sindresorhus I implemented the things we mentioned. Not sure about the tooltip styling, just used react-tooltip for it

sindresorhus:
I think the tooltip needs some work. It was not clear to me at first it was even a tooltip as it blended into the background.

Screen Shot 2021-05-27 at 15 54 17

Here's what I would do:

- Add more shadow to it.
- Make its background darker.
- More vertical padding and slightly less horizontal padding inside it.
- A tiny bit more border radius on it.

And `application` => `app`.

sindresorhus:
I  also thought of another thing. I think we should follow the Big Sur style for toolbar buttons:




Meaning, just icons.

sindresorhus:
@karaggeorge Maybe we should just get this merged and then open an issue about the remaining stuff? The remaining stuff is not essential.

PR #1017, opened 2021-05-23
Electron v10

karaggeorge:
Upgraded to v9 originally, fixed the issues.

Then I saw with the end of support message here: https://www.electronjs.org/blog/electron-12-0

So I decided to try to go to v10 as well, only breaking change from 9 -> 10 was the explicit `enableRemoteModule: true`

This is a PR to the exports PR branch that is also open. Made a decent amount of changes so I didn't want to miss any issue in those in the version upgrades

### DIFF
--- a/main/index.ts
+++ b/main/index.ts
@@ -1,4 +1,4 @@
-import {app} from 'electron';
+import {app, protocol} from 'electron';
 import {is, enforceMacOSAppLocation} from 'electron-util';
 import log from 'electron-log';
 import {autoUpdater} from 'electron-updater';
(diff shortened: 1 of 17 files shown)


main/index.ts
```diff
@@ -102,6 +102,11 @@ const checkForUpdates = () => {
     app.setAsDefaultProtocolClient('kap');
   }
 
+  protocol.registerFileProtocol('file', (request, callback) => {
```
sindresorhus, line 105: Can you add a code comment about why this exists. Might be worth moving it into a util file.

sindresorhus (changes requested):

main/index.ts
karaggeorge: Moved to a util file and added a comment 👍

sindresorhus:
Seems to be working well 👍

Issue #629, opened 2019-02-15
Support mp4 export with HEVC encoding


karaggeorge:
Currently (since #626) we support opening external `hevc` files or recording with `hevc` encoding if your hardware supports it. But in either case, our mp4 exports are always `h264`. Ideally we should be able to export with `hevc` encoding if our source is `hevc`. 

The issue was that when converting `ffmpeg`, if you don't specify a video codec it will use the default it has for that type. For mp4s etc it is `h264`. If we want to tell it to use `hevc` for encoding, we can either do:
- `-vcodec copy` which will keep the encoding at `hevc`, but since the video stream never gets decoded, you can't change dimensions, fps or anything else
- `-vcodec hevc` which in theory should work but it fails on my machine. I can't find examples for the options I need to use for that or test any of it since I can't get it to work. It might be because my hardware can't handle `hevc` to begin with, even though `ffmpeg` says I have that encoder.

If the same criteria for recording `hevc` exist for encoding to `hevc`, we might be able to use that same check to use the encoder, but until someone can test it and find the right `ffmpeg` arguments, we can't export in `hevc` format.

IssuehuntBot:
@issuehunt has funded $100.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/65411043/issues/629) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/65411043/issues/629).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

protium-dev:
Is this still relevant? I could help with testing or implementing it

codetheweb:
@sindresorhus I think I forgot to open the PR through IssueHunt, could you reward it when you get a chance? 😃

Issue #970, opened 2021-01-07
Native Apple Silicon support


sungkhum:
On my new Mac mini with the M1 processor, encoding videos is super slow with Kap. I was wondering if there were any plans to support the new M1 processor. If there are, I would be happy to beta test and give feedback. Thanks!

sindresorhus:
Yes, but it's probably not happening soon. Some pieces need to fall into place first:

- [ ] https://github.com/kribblo/node-ffmpeg-installer/issues/50
- [ ] New version of https://github.com/wulkano/Aperture that includes an Apple Silicon slice in the binary.
- [ ] Upgrade to Electron 11, which is blocked by https://github.com/electron/electron/issues/23757.
- [ ] Configure electron-builder to also do an Apple Silicon build.

---

*All the other binaries we use don't need an Apple Silicon to build yet as they wouldn't benefit from it.*

sungkhum:
Thanks for the update @sindresorhus  - totally makes sense. But glad that it could be possible! Kap is a great tool!

raulrpearson:
It does seem like it could become a long wait. `node-ffmpeg-installer` seems to be [relying on evermeet.cx](https://github.com/kribblo/node-ffmpeg-installer#the-binaries) for macOS binaries, but they in turn seem to have [no intention of supporting those binaries](https://evermeet.cx/ffmpeg/apple-silicon-arm) in the foreseeable future. There is [that cypress-io fork](https://github.com/kribblo/node-ffmpeg-installer/issues/50#issuecomment-774237083), though.

@sindresorhus, in the meantime, is there a workaround that we could use? Building our own ffmpeg and inserting into the Kap.app folder? Saving the un-encoded video capture and converting it ourselves? A 2-minute capture took 40 minutes of encoding and a reasonable amount of heat on my M1 which is a bit disappointing 😅.

I'm happy to help fix this, but I'm not sure how to.

tortuetorche:
Hi folks,

>@sindresorhus, in the meantime, is there a workaround that we could use? Building our own ffmpeg and inserting into the Kap.app folder?

Great idea 💡! Here what I did to work around the encoding issue:

Firstly you need to [install Homebrew](https://brew.sh/) or your Mac, then run these commands in a terminal:
```bash
brew install ffmpeg
[...]

PR #1021, opened 2021-06-21
Add support for HEVC, use ffmpeg-static

codetheweb:
Closes https://github.com/wulkano/kap/issues/629.

Changes:
- Adds support for libx265 / HEVC
- Switches out `@ffmpeg-installer/ffmpeg` for `ffmpeg-static`
	- The ffmpeg versions included by the current package are 3 years out of date. There's talk of [merging it with `ffmpeg-static`](https://github.com/kribblo/node-ffmpeg-installer/issues/55), but currently the latter is more up-to-date. It also [supports M1](https://github.com/eugeneware/ffmpeg-static/releases/tag/b4.4), moving us closer to complete [Apple Silicon support](https://github.com/wulkano/Kap/issues/970).

### DIFF
--- a/main/remote-states/editor-options.ts
+++ b/main/remote-states/editor-options.ts
@@ -10,9 +10,9 @@ const exportUsageHistory = new Store<{[key in Format]: {lastUsed: number; plugin
   name: 'export-usage-history',
   defaults: {
     gif: {lastUsed: 6, plugins: {default: 1}},
-    hevc: {lastUsed: 5, plugins: {default: 1}},
-    mp4: {lastUsed: 4, plugins: {default: 1}},
-    webm: {lastUsed: 3, plugins: {default: 1}},
+    mp4: {lastUsed: 5, plugins: {default: 1}},
+    webm: {lastUsed: 4, plugins: {default: 1}},
+    hevc: {lastUsed: 3, plugins: {default: 1}},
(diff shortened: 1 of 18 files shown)


main/remote-states/editor-options.ts
```diff
@@ -9,7 +9,8 @@ import {prettifyFormat} from '../utils/formats';
 const exportUsageHistory = new Store<{[key in Format]: {lastUsed: number; plugins: Record<string, number>}}>({
   name: 'export-usage-history',
   defaults: {
-    gif: {lastUsed: 5, plugins: {default: 1}},
+    gif: {lastUsed: 6, plugins: {default: 1}},
+    hevc: {lastUsed: 5, plugins: {default: 1}},
```
karaggeorge, line 13: I think we should add this after webm, to avoid similar issues we had with AV1

Thoughts on the order? @sindresorhus

sindresorhus: Agreed, at least as long as Chrome doesn't support HEVC playback.

sindresorhus:
Seems to work fine for me.

main/remote-states/editor-options.ts
codetheweb: Fixed in 086b988.

sindresorhus:
@karaggeorge LGTY?

PR #1051, opened 2021-11-19
Update to Electron v12, allow packaging as arm64 for macOS

codetheweb:
Closes https://github.com/wulkano/Kap/issues/970.

To my knowledge, the only binary now that does not support arm64 is gifsicle. I [opened a PR](https://github.com/imagemin/gifsicle-bin/pull/132), up to you if you want to wait for that to merge this.

### DIFF
--- a/main/windows/config.ts
+++ b/main/windows/config.ts
@@ -23,7 +23,8 @@ const openConfigWindow = async (pluginName: string) => {
     modal: true,
     webPreferences: {
       nodeIntegration: true,
-      enableRemoteModule: true
+      enableRemoteModule: true,
+      contextIsolation: false
     }
   });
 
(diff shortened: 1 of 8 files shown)


sindresorhus:
Thanks for working on this. Make sure you have thoroughly tested that nothing broke in the Electron update.

codetheweb:
Everything seems to be working fine so far in my testing (changing preferences, encoding to different formats, using plugins, etc.).

 Looks like the PR on gifsicle was merged but hasn't been released yet.

sindresorhus:
We cannot upgrade to the latest version of `gifsicle-bin` as it requires ESM, which Electron do not support. However, `gifsicle-bin` just fetches the binary from the repo, which is now updated, so in theory, we should not need to do anything.

codetheweb:
Makes sense. Thank you!

I would update `https://kap-latest-release.now.sh/api` but it doesn't look like the source for that route is public.

ajwhite:
Heads up -- the community is experiencing a big impact on performance somewhere around the time of this patch, described here https://github.com/wulkano/Kap/issues/1125

PR #1076, opened 2022-01-17
Adjust CI logic for two artifacts

karaggeorge:
We had: 
```
mv dist/*.dmg dist/Kap.dmg
```

but with 2 .dmg files generated for Intel/M1 this was failing in CI
Adjusted the logic to upload both artifacts. Also updating the logic for the release endpoint and webhook


AlecRust:
FYI `brew bump-cask-pr --version=3.5.2 kap` seems to fail after this change.

PR #1096, opened 2022-02-18
Replace remote usage with @electron/remote

timothyis:
This pull request replaces all usage of the remote module – built in to Electron before v14 – with the new @electron/remote module.

I have tested this, but I'm not sure if I've missed any edge-cases.

Note: We should work on replacing all usage of this module with other methods, such as using `ipc` instead.

Closes #1089.

PR #1102, opened 2022-03-13
Update/replace remote

cgarrovillo:
Before I move any further on this, thought I'd open this up for a review. This PR only removes `is` & `isFirstAppLaunch` so far from `electron-util`,  tests showing no regression.

One thing that stands out to me at first glance in the codebase is utils/ needing to be shared between main/ & renderer/

### DIFF
--- a/main/utils/environment.ts
+++ b/main/utils/environment.ts
@@ -1,0 +1,13 @@
+import electron from 'electron';
+
+export const isDevelopment = () => {
+  if (typeof electron === 'string') {
+    throw new TypeError('Not running in an Electron environment!');
+  }
+
+  const isEnvSet = 'ELECTRON_IS_DEV' in process.env;
+  const getFromEnv =
+    Number.parseInt(process.env.ELECTRON_IS_DEV!, 10) === 1;
+
+  return isEnvSet ? getFromEnv : !electron.app.isPackaged;
+};
--- a/main/utils/launch.ts
+++ b/main/utils/launch.ts
@@ -1,0 +1,25 @@
+import Electron from 'electron';
+import fs from 'fs';
+import path from 'path';
+
+export const isFirstAppLaunch = (): boolean => {
+  const checkFile = path.join(
+    Electron.app.getPath('userData'),
+    '.app-launched'
+  );
+
+  if (fs.existsSync(checkFile)) {
+    return false;
+  }
+
+  try {
+    fs.writeFileSync(checkFile, '');
+  } catch (error: any) {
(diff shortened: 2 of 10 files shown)


main/utils/environment.ts
```diff
@@ -0,0 +1,13 @@
+import electron from 'electron';
+
+export const isDevelopment = () => {
+  if (typeof electron === 'string') {
+    throw new TypeError('Not running in an Electron environment!');
+  }
+
+  const isEnvSet = 'ELECTRON_IS_DEV' in process.env;
+  const getFromEnv =
+    Number.parseInt(process.env.ELECTRON_IS_DEV!, 10) === 1;
```
timothyis, line 10: I'm not sure we use `ELECTRON_IS_DEV`, it might save a lot of time and space to just use the `electron.app.isPackaged` method everywhere we need this. Less complexity too.

main/utils/launch.ts
```diff
@@ -0,0 +1,25 @@
+import Electron from 'electron';
+import fs from 'fs';
+import path from 'path';
+
+export const isFirstAppLaunch = (): boolean => {
+  const checkFile = path.join(
+    Electron.app.getPath('userData'),
+    '.app-launched'
```
timothyis, line 8: Has this been renamed from the electron-utils string? If it's been renamed, it will think everyone has opened the app for the first time regardless of if they have or not. We should maintain the old string.

renderer/utils/environment.ts
```diff
@@ -0,0 +1,13 @@
+import electron from 'electron';
+
+export const isDevelopment = () => {
```
timothyis, line 3: Since this is duplicating code, why not use ipc to send this information from the main to renderer?

timothyis:
Nice work @cgarrovillo! I've left some comments based on the work I've done as well. I feel as though we may be working on exactly the same thing however, so I'm going to push my code tomorrow and link it here so we can sync up.

cgarrovillo:
Sounds good, feel free to overwrite my changes then,  and we can talk about what's left.  If there's tedious but easy tasks I can do I'll be happy to take those off your plate.

sindresorhus:
Since https://github.com/wulkano/Kap/pull/1096 is not moving forward, I think this pull request is a good step in the right direction.

@cgarrovillo Would you be willing to address the pull request feedback?

timothyis:
I've partially resolved this in #1148. Will continue after. Thanks for opening up a PR though ❤️

PR #1112, opened 2022-04-12
Adding unit tests and mkv formatting for myself

makhambettorezhan:
I know that adding mkv format is not asked in the issues box or this feature may never needed because of having mp4 or webm for video formats but for me I like mkv format, and use it most of the times when recording my screen. So am i going in the right direction? Thank you in advance


sindresorhus:
Thanks for the pull request, but I don't think we're interested in adding MKV support.

Issue #1125, opened 2022-06-15
Kap is extremely slow to show capture selection (10 seconds)


nicmar:
**macOS version:** 10.12.1
**Kap version:** Version 3.5.4 (3.5.4.1810)

#### Steps to reproduce
Pressing Kap button in the menubar

#### Current behaviour
Takes 8-10 seconds before i can select an area to record. Also right click "About Kap" doesn't give any information or feedback.

#### Expected behaviour
In the earlier versions, this was extrememly fast, like less than 1 second, but it has been slower in later versions, like 3-4 seconds. It was still acceptable.

#### Workaround
 No workaround, just have to wait.

It might have been worse since upgrading to Monterey but I'm not sure.

malandles:
Same behavior on macOS Monterey 12.4 with Kap version 3.5.4 (3.5.4.1810).

sveggiani:
Noticing this for a time now. It have made it almost unusable for me :/

carlosperate:
Having the same issue in an Intel Macbook Pro with macOS Monterey 12.1.

I've tried going back to each version from the releases pages until one of them worked well. In this case I had to go back to v3.4.1.

v3.4.2 didn't work (likely the signing issues indicated in the release notes) and v3.5.0 already shows this problem. Looking at the diff, is fairly small, so maybe we can try manually reverting these changes to see if it makes a difference: https://github.com/wulkano/Kap/compare/v3.4.1...v3.5.0

wkoutre:
Same issues here, macOS 12.4 w/ multiple monitors.

Downgrading to 3.4.1 as @carlosperate suggested has worked for me.

ajwhite:
Can confirm `3.4.1` resolved the issue for me as well.

ajwhite:
@carlosperate this contribution stands out to me in that diff, https://github.com/wulkano/Kap/pull/1051 -- upgrading electron & supporting different build target architectures

@codetheweb @sindresorhus can you see that impacting this behavior?

ajwhite:
Cross referencing similar thread https://github.com/wulkano/Kap/issues/1088

codetheweb:
Hmm, something must have changed in the new Electron version.

Might be related: https://github.com/desktop/desktop/issues/12978

literalpie:
Any idea if going to a newer version of electron would fix this? The latest is v19.0, and Kap is on 12.x. Going to new electron would also move this from Node 14.x to 16.x.

[...]

PR #1148, opened 2022-10-27
Replace internal remote module usage

### LINKED ISSUE #1089: Migrate to `@electron/remote`
It blocks us from upgrading Electron and we're left on an old Electron version...

https://github.com/electron/remote/blob/main/README.md#migrating-from-remote

timothyis:
The intended purpose of this pull request is to replace all [built-in remote module](https://nornagon.medium.com/electrons-remote-module-considered-harmful-70d69500f31) usage in our own code.  

There are multiple instances of remote being used in external modules, but we should address those in a further PR.

This is a work in progress.

Fixes #1089


sindresorhus:
Very nice work! It works great for me in dev mode.

sindresorhus:
I did run into an issue when in the built version though (`npm run list`). I cannot get past the screen recording permission prompt no matter how many times I give permission and restart. I cannot reproduce this issue in Kap 3.6.0.

.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules
 /app/dist
 /dist
 /dist-js
+*.DS_Store
```
sindresorhus: Nitpic, but this should be in your own global gitignore, not here.

https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

timothyis: Ah, thank you! Great point.

timothyis:
Ah, perhaps I'm not up-to-date with main. Let me see what I can do here 🙏🏼

timothyis:
Looks like this still happens after rebasing with main. For me the following occurs after rebasing:

1. Double click Kap in the `dist` directory and a blank space appears in the tray where Kap is meant to be. Nothing else happens. (Kap shows up in the activity monitor however). I think this is when the "Move to applications folder required" message should pop up.
2. Force quit Kap from the activity monitor (won't just quit normally)
3. Open Kap from the applications folder (it moved itself there automatically), the Kap icon appears in the dock and tray and I get a dialog asking for permissions. No combination or disabling or re-enabling works to solve this issue. Kap still won't quit normally and needs force-quitting. 

I think perhaps these parts still use Electron remote upstream somewhere and I need to check that.

sindresorhus:
> I think perhaps these parts still use Electron remote upstream somewhere and I need to check that.

Most likely, yes.

timothyis:
Just to note on the first issue; the dialog for moving Kap to the apps directory isn't a remote issue. I completely ripped remote out and it still didn't work. In my minimal test app, I upgraded Electron and electron-builder to the latest versions and it worked as expected. I'll try updating Electron until that first part works and then move on to the permissions issue.

wojtiku:
Sorry for intruding here. I love Kap a lot but is has been unusable for the past few months because of a ~30s delay before being able to record. This drove me to try different things including building locally and trying out this branch. To the point though:

After upgrading to MacOS Ventura (Intel), I also had a permission issue with Kap where I was unable to grant the permission no matter how many times I would agree - the dialog just kept appearing. This includes this branch.

What I found out to always work though was selecting kap on the list and clicking the minus (1) button and then clicking the plus button (2) and selecting the binary I just installed. This fixes the permission issue every time for me.
image

Hope this makes it easier for you to debug.

P.S.

Regarding issue #1125, even though this branch doesn't fix it for me as I hoped it would, it changes things - on 3.6.0 I need to wait about 20-30s after clicking the Kap icon for anything to show up and to record. On this branch, the UI (grey box with controls and the red record button) is shown immediately after clicking Kap icon but it is frozen for 20-30s before I can do anything.

I love Kap and I can't find any alternatives since it stopped working for me so I really cheer to your efforts.

timothyis:
Hi @wojtiku, thank you for commenting. I too found that fix worked for me, but this sadly also stopped me from debugging it! I wish I could reproduce it.

I'm very busy with work right now, but I'm hoping I can find some time to finish this PR in the next few weeks.

I spoke with Sindre privately about the best method forward and hoping that this can be closed soon and we can move onto other issues.

Just to note that this branch does not yet include the Electron updates to latest, or switching to IPC instead of remote, which should bring a lot of performance improvements. 

Really appreciate your kind words and patience 🙏🏼

wojtiku:
Regarding reproduction, I was able to consistently get to a permission-related problem with the following steps on this branch:
1. use minus button to remove Kap from the permission list
2. **do not** add it manually
3. open Kap built on this branch (I did not verify with current prod)
4. you will get the permission dialog (go to system prefs or deny)
5. click go to system prefs
6. allow kap to record the screen
7. click `quit and reopen` in the dialog
8. after reopening click kap in the top toolbar
9. everything is frozen indefinitely, seems like there is invisible overlay on the screen preventing any interaction<br>fortunately I have terminal under a keyboard shortcut and `killall Kap` solves the issue

Hope this helps. I don't think I could contribute with such a complex issue not knowing the project but if you need a tester feel free to mention me and I will test on my side.

Thanks!

timothyis:
Replaced by #1234

PR #1208, opened 2024-02-13
Sped up extremely slow loading of Cropper and context menus

michaeljelly:
I found Kap unusably slow! The main culprits:

1. Context menus were being built loaded on every click rather than once and cached
2. The Cropper window was being recreated every time Kap was triggered, and for every display or something.

Unfortunately it seems my default settings have messed with tabs/spaces stuff. But anyway if you want fast-kap, here it is! Feels like a totally different app


CleverCoder:
I just pulled this fork and built locally. I discovered three issues:
- Shortcut wouldn't work (although I think it's because I didn't create a signed build) on MacOS
- It only worked on one of my monitors (I have three total - including my Macbook)
- It started immediately with the rectangle selection engaged

The multi-monitor issue is a big one. Maybe I have something goofed with my Electron build? I basically "yarn" and "yarn run pack"

Otherwise. Fantastic effort! If I get time later, I might try and pinpoint these issues locally.

michaeljelly:
I'll have a new build today with fixes for the 2nd one :)

karaggeorge:
@michaeljelly can you rebase with `main`? Pushed a commit to fix the CircleCI config so we can get a build to test

skllcrn:
Hyped!

CleverCoder:
Tested latest code. Couple of issues:
Getting a build error on my mac:

```sh
sean@SAITKEN-MAC Kap % yarn run pack   

> kap@3.6.0 build
> yarn build-main && yarn build-renderer

command not found: export
```

Think it's caused by this:
image

Also, there's a bunch of console.log() that was added.. Might want to remove, or switch to a logger (if there is one already in the codebase)

UPDATE: I fixed the build issue by removing "export" from the npm script. However, two things:
- Upon launch, it immediately goes into capture mode
- I still can't capture on all monitors. Just one of them.

michaeljelly:
Yep sorry I was getting errors without adding the NODE_OPTIONS, building seemed to work when I added the export on Mac but not when I just had it as above.

@CleverCoder I thought that was expected behaviour, is it not?

I believe I have now fixed the shortcuts issue and may have fixed the multiple monitors. Will rebase now and then we can see what's left to do.

michaeljelly:
rebased @karaggeorge

michaeljelly:
CircleCI build failed with this error:

onHosts.watchOS of plug-in com.apple.dt.IDEWatchSupportCore
2024-02-19 13:44:36.988 xcodebuild[993:5278] Requested but did not find extension point with identifier Xcode.IDEKit.ExtensionPointIdentifierToBundleIdentifier for extension Xcode.DebuggerFoundation.AppExtensionToBundleIdentifierMap.watchOS of plug-in com.apple.dt.IDEWatchSupportCore
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module '/Users/distiller/project/.yarn/releases/yarn-1.22.21.cjs'
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at node:internal/main/run_main_module:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Exited with code exit status 1

skllcrn:
> Upon launch, it immediately goes into capture mode

This is the current behaviour, correct @michaeljelly!

CleverCoder:
yeah... going into capture mode after launch is weird. I want to have it
always running, but not make assumptions. If I launch it manually, I
expect it to sit silently in my task bar / header nav menu (whatever it's
called in MacOS) and not assume I want to make a capture until I hit the
keyboard shortcuts or clicky on a thing.
I don't recall this as existing behavior.
I'll pull the latest and run some tests tomorrow. I want this to be useful,
as the commercial options are just over the top... in terms of trying to
leech my precious $$$.... :) this is simple, and to the point, which I
respect.

Cheers!
- Sean

CleverCoder:
Also, tested with the latest code and it still only works on one of my monitors. I have two external monitors connected to my 2019 16" MacBook Pro, and only one of the external monitors works with this code.
In addition, the shortcut is working for me now, and when I activate Kap, it starts off right in the middle of creating a rectangular region selection, even though I never clicked the mouse. It's quite janky. 🤷 

And I think I found another issue. I successfully performed one capture, but when I attempted to use it a second time, the selection area doesn't work, and it seems stuck in a weird errored state. After quitting and restarting Kap, it's working again, still with the weird initial selection area just being strange. Always have to click off then re-do the selection.

skllcrn:
Thank you @CleverCoder, to keep things moving I suggest opening a separate issue with your suggestion, and continue focusing on tackling the main culprits of the performance issues here.

CleverCoder:
Glad to keep testing! The biggest issue (that is a regression, in this case) is the lack of supporting multiple displays. It's a deal-breaker with this change. Also, only working just one time is kind of a big deal. :)
🥂

arisbond:
will anyone fix this issue for intel  ~~

0x00000000000000000000000:
> I just pulled this fork and built locally. I discovered three issues:
> 
> * Shortcut wouldn't work (although I think it's because I didn't create a signed build) on MacOS
> * It only worked on one of my monitors (I have three total - including my Macbook)
> * It started immediately with the rectangle selection engaged
> 
> The multi-monitor issue is a big one. Maybe I have something goofed with my Electron build? I basically "yarn" and "yarn run pack"
> 
> Otherwise. Fantastic effort! If I get time later, I might try and pinpoint these issues locally.

Have all of the issues been resolved?

HugoHeneault:
Hey there! Would be awesome having it fixed up to main. Thanks for your times and fixes.

Maybe fixing lint so it matches current project requirements to make review simpler would help? 
Maintainer will have an easier review and so it goes forward?

yotam-frid:
Hey any plans to merge this? Looks amazing.

HugoHeneault:
I tried locally this morning and I confirm that it's not working multiple monitors anymore. Any plans to make it working @michaeljelly ? :)

PR #1234, opened 2024-10-20
Upgrade to Electron 33

karaggeorge:
This upgrades to Electron 33:

- Refactored all usages of `remote` to use `ipc` instead
- Refactored around some deprecated APIs

Known broken things that were not migrated:
- App icons in the cropper are not working
- ~Editor video is not centered~
- ~Editor window traffic lights are broken~
- Editor title bar has to stay down at all times (not show on hover) because of new `-webkit-app-region: drag;` limitations

I tried to upgrade as few packages as possible. I have another branch that upgrade all our utils, but we need to upgrade the entire project to use ESM (`type: "module"`). I did do that, but then the renderer Next.js app is not working. I tried upgrading Next.js and getting that to work, but `electron-next` was not working with it 🙈 

Overall, we still have `.js` code, class components, in-line `styled jsx` and a lot of other really outdated code
I _think_ we need to move away from `electron-next` and Next.js, and just have a regular webpack build (or vite) for the renderers

Also, a lot of code needs to be re-written to improve performance, especially the cropper


albinekb (approved):
Amazing job! 👏

x0d7x:
thx mate 👏

Issue #172, opened 2017-05-12
Show estimated GIF size in the editor window


sindresorhus:
As a user it's often hard to estimate just how big your GIF will end up being. GIFs are huge and wasteful. Would be useful to give the user an approximate size. I'm not 100% sure how, but I've seen it done in some other apps, so should be possible.

matheuss:
I'd love to see this 🙏 I have no clue how to do it tho 🤔

sindresorhus:
One idea would be to export part of the recording in the background and multiply it to get an average. Another idea would be to resize the recording to a small version and then export that in the background and multiply it by how much it was resized and check the size. Or combine those techniques.

Some more resources:
- https://stackoverflow.com/questions/23920098/how-to-estimate-gif-file-size
- http://jan.ucc.nau.edu/lrm22/pixels2bytes/calculator.htm

kevva:
Haven't looked through the recording code, but you could pipe to `ffmpeg` (see https://ffmpeg.org/ffmpeg-protocols.html#pipe) and then convert and write to another stream and check it's `data` event and count the total bytes written to it. Maybe not that performant though, or easy to implement :).

cryogenx:
+1 this would be very helpful

IssuehuntBot:
@issuehunt has funded $100.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/65411043/issues/172) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/65411043/issues/172).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

PR #1277, opened 2026-05-10
Show estimated GIF size in the editor

apples-kksk:
This adds a small `GIF ~x MB` estimate beside the export controls when GIF is selected.

The estimate samples up to the first 2 seconds of the selected range through the existing GIF conversion/compression path, then scales that size to the selected duration. If the editor options change, the previous in-flight sample is canceled so stale estimates do not pile up.

Checked:
- `yarn build-main`
- `./node_modules/.bin/tsc --project renderer/tsconfig.json --noEmit`
- `npx --yes -p node@16 -c 'yarn lint'` (existing warnings only)\n- `yarn test:main`\n- `npx --yes -p node@16 -c 'yarn build-renderer'`\n\nCloses #172

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,9 +2,33 @@ version: 2
 jobs:
   build:
     macos:
-      xcode: '13.4.1'
+      xcode: '14.3.1'
+    resource_class: m4pro.medium
     steps:
       - checkout
+      - run:
+          name: Use Node.js 16
+          command: |
+            nvm install 16
+            nvm alias default 16
+            {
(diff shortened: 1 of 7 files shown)


apples-kksk:
The CircleCI build did not reach the project steps. CircleCI rejected the current macOS executor before checkout with:\n\n```\nJob was rejected because resource class m4pro.medium, image xcode:13.4.1 is not a valid resource class\n```\n\nThat looks like the repository CircleCI image/resource-class configuration rather than a failure in this change. The local checks listed in the PR body completed successfully.

PR #1283, opened 2026-07-12
Add cancellable GIF size estimates

### LINKED ISSUE #172: Show estimated GIF size in the editor window

landeqiming666:
## Summary

Adds an approximate GIF size next to the editor export controls. The estimate converts a short sample with the same dimensions, frame rate, and selected range, then scales it to the full duration.

The estimate is canceled when export options change, the format changes, or the editor component is unmounted, so an obsolete background conversion does not continue after the result can no longer be shown. The probe deliberately does not invoke edit services because those can be interactive or have side effects.

## Validation

- `TS_NODE_TRANSPILE_ONLY=1 node ./node_modules/ava/cli.js`
- `node ./node_modules/typescript/bin/tsc --noEmit -p tsconfig.json`
- `node ./node_modules/typescript/bin/tsc --noEmit -p renderer/tsconfig.json`
- `node ./node_modules/next/dist/bin/next build renderer`
- `node ./node_modules/next/dist/bin/next export renderer`
- targeted XO lint
- `git diff --check`

Closes #172


landeqiming666:
CircleCI did not reach checkout or run this change. The job was rejected because the repository macOS executor combines `m4pro.medium` with `xcode:13.4.1`, which CircleCI now reports as invalid.

Local validation completed: full AVA suite, both TypeScript projects, focused lint, production renderer build/export, and `git diff --check`.

albinekb:
7

PR #1295, opened 2026-09-17
Fix Apple Silicon builds on Macs without Rosetta

### LINKED ISSUE #1294: Unhandled Promise Rejection in macOS 27
**macOS version:**    27.0 (26A428)
**Kap version:**      3.6.0

#### Workaround

Unhandled Promise Rejection
Error: Command failed: /Applications/Kap.app/Contents/Resources/app.asar.unpacked/node_modules/mac-screen-capture-permissions/screen-capture-permissions
spawnSync /Applications/Kap.app/Contents/Resources/app.asar.unpacked/node_modules/mac-screen-capture-permissions/screen-capture-permissions Unknown system error -86
    at Function.module.exports.sync (/Applications/Kap.app/Contents/Resources/app.asar/node_modules/mac-screen-capture-permissions/node_modules/execa/index.js:168:25)
    at exports.hasScreenCapturePermission (/Applications/Kap.app/Contents/Resources/app.asar/node_modules/mac-screen-capture-permissions/index.js:27:25)
    at ensureScreenCapturePermissions (/Applications/Kap.app/Contents/Resources/app.asar/dist-js/common/system-permissions.js:64:23)
    at Object.openCropperWindow [as open] (/Applications/Kap.app/Contents/Resources/app.asar/dist-js/windows/cropper.js:81:66)
    at Object.openCropperWindow (/Applications/Kap.app/Contents/Resources/app.asar/dist-js/tray.js:26:129)

robertkludwig:
Fixes #1294.

## What happens

On an Apple Silicon Mac without Rosetta, Kap throws as soon as you click the tray icon or open the cropper:

```
Unhandled Promise Rejection
Error: Command failed: .../mac-screen-capture-permissions/screen-capture-permissions
spawnSync .../screen-capture-permissions Unknown system error -86
```

`-86` is `EBADARCH`. Node has no name for it, so it prints the raw errno. In a shell the same failure reads `bad CPU type in executable`.

## Why

Kap ships an arm64 app, but several bundled helper executables are Intel-only:

```
$ lipo -archs Kap.app/Contents/Resources/app.asar.unpacked/node_modules/*/[a-z]*
mac-screen-capture-permissions/screen-capture-permissions  x86_64
macos-audio-devices/audio-devices                          x86_64
mac-windows/scripts/MacWindows                             x86_64
mac-open-with/open-with                                    x86_64
node-mac-app-icon/run                                      x86_64
gifsicle/vendor/gifsicle                                   x86_64
aperture/aperture                                          x86_64 arm64
```

This was invisible for years because Rosetta was usually installed. It no longer is: per Apple's macOS 27 release notes, *"If Rosetta was previously installed, it is not automatically restored after upgrading to macOS 27.0."* So every one of those helpers now fails on a machine that used to run them fine, and users experience it as Kap breaking on upgrade.

The screen capture permission check is the first helper Kap spawns, which is why the app appears dead rather than partially broken.

`aperture` is the one that is already correct, because it alone builds with `swift build --arch arm64 --arch x86_64`.

## What this changes

**1. Drop `mac-screen-capture-permissions`.** Electron reports this permission natively via `systemPreferences.getMediaAccessStatus('screen')`, documented since well before the Electron version Kap pins. Removing the dependency removes the executable rather than rebuilding it, and `system-permissions.ts` already used the same API for the microphone.

The old helper did two things: it read the permission, and when the permission was missing it called `CGRequestScreenCaptureAccess` to ask for it. `getMediaAccessStatus` only covers the first, so on the first denial this asks as well, through `desktopCapturer.getSources`, and falls back to the existing "Open System Preferences" dialog on later attempts. That keeps the original branch structure intact.

I verified that end to end rather than assuming it. On a Mac with the permission revoked via `tccutil reset ScreenCapture com.wulkano.kap` and the marker file cleared, launching this build produced the standard system prompt, *"Kap" would like to record this computer's screen and audio*, and Kap then appeared under Screen & System Audio Recording in System Settings. So first-run behaviour matches what the native helper did.

**2. `gifsicle` to `^5.3.0`.** The lockfile pinned 5.2.0, whose vendored binary is Intel-only. 5.3.0 is the first universal one. Note the range in `package.json` was already `^5.2.0`, so this is a lockfile problem, not a version-range problem.

**3. Fetch `ffmpeg-static` for the architecture being packaged.** It resolves its binary from `npm_config_arch || os.arch()` at install time, which is the *build machine's* architecture. Building both targets on one machine therefore puts the same ffmpeg in both, so one of the two published DMGs always ships an ffmpeg that cannot run. A `beforePack` hook re-runs its installer per target.

**4. Fail the build when a bundled executable is missing the architecture it is packaged for.** This is the check that would have caught all of the above at release time instead of in users' hands. `afterPack` walks the packaged app, finds Mach-O files and compares `lipo -archs` against the target.

## Still outstanding

Four dependencies still publish Intel-only helpers upstream and cannot be fixed from this repo:

| package | last published | affects |
|---|---|---|
| `macos-audio-devices` | 2020 | audio device list (falls back to `aperture`, so it degrades rather than breaks) |
| `mac-windows` | 2020 | recording a specific window |
| `mac-open-with` | 2021 | the "Open with" menu |
| `node-mac-app-icon` | 2018 | app icons in the window picker |

All four are a one-line fix — they build with plain `swift build`, so they inherit whatever Mac the maintainer used, where `aperture` passes `--arch arm64 --arch x86_64`. I verified all four build universal and run natively on arm64 with that flag added. `node-mac-app-icon` additionally needs `-Xswiftc -static-stdlib` dropped, which is now a hard error in current Swift, and its output path updated.

Until those publish, the architecture check lists them and does not fail the build. Screen recording itself works without them, because `aperture` and `ffmpeg` are fine.

## Verification

Built and packaged for arm64 on macOS 27 with Rosetta absent, confirmed by `arch -x86_64 /usr/bin/true` failing with `Bad CPU type in executable`.

- `yarn lint` across the repo: no errors.
- `yarn test:main`: 30 tests pass, including the conversion suite that exercises ffmpeg and gifsicle natively.
- `electron-builder --dir --arm64`: packages successfully with the new architecture check passing.
- Added tests for the permission paths, which had none. The first of them fails on `master` with the exact `errno -86` from the issue.
- The architecture check was confirmed to fail the build on the offending binaries before they were replaced, and to pass afterwards.

The resulting build runs and records on macOS 27 with no Rosetta installed.

Note on CI: the `build` check on this PR is red, but the job is rejected by CircleCI before any step runs, with `invalid-resource-class`. Every PR build in this repo has failed the same way since early September, and the config here is untouched. Details in a comment below.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


robertkludwig:
The red `build` check on this PR is not coming from this diff. The job is rejected by CircleCI before a single step runs, and it has been happening to every PR in this repo, not just this one.

From CircleCI's own API for the build on this branch:

```json
"executor": { "resource_class": null, "type": null },
"messages": [
  {
    "type": "error",
    "message": "Job was rejected because resource class macos.m1.medium.gen1, image xcode:13.4.1 is not a valid resource class",
    "reason": "invalid-resource-class"
  }
],
"started_at": null
```

`started_at` is null, no executor is ever assigned, and the job reports zero steps. Queue and stop timestamps are about 30ms apart.

Build history for the project shows when this started:

| build | date | result |
|---|---|---|
| 1917 | 2026-07-20 | success, ran 211s on `m4pro-medium` |
| 1920, 1921, 1922 | 2026-07-31 to 2026-09-03 | rejected, no executor |
| 1923 | this PR | rejected, no executor |

The cause is the free-plan macOS migration. CircleCI moved free-plan organisations to `m4pro.medium` and stated that macOS jobs run there "regardless of the resource class specified in your configuration", and the M1 and M2 classes reached end of life. `.circleci/config.yml` here pins `xcode: '13.4.1'`, which is not offered on M4, so the job cannot be placed.

I tried naming the old class explicitly to get it running again. That produced the error quoted above, so that route is closed and I reverted it. **This PR leaves `.circleci/config.yml` untouched.**

Fixing CI is not a one-line image bump either, and I would rather say so than guess in a PR. The renderer builds with Next 10 on webpack 4, which fails on current Node with `ERR_OSSL_EVP_UNSUPPORTED`. I reproduced that locally on Node 25. A newer Xcode image therefore also needs a pinned Node, and I verified that Node 16.20.2 arm64 still builds and tests this project cleanly on macOS 27. I have not verified which Xcode images are currently valid on `m4pro.medium`, since CircleCI's docs lag their deprecations, so I am not going to guess at one here. Happy to open a separate PR for it if you want, and to iterate there rather than on this one.

In the meantime, here is what the three CI steps do locally on macOS 27, arm64, with this branch:

```
yarn lint     0 errors (18 pre-existing warnings, all in untouched files)
yarn test     30 tests passed
electron-builder --dir --arm64   packaged, architecture check passed
```

